PR #4, opened 2018-08-07
Unify GeosCore module initializations to accept the same arguments

jimmielin:
This commit unifies all modules in `GeosCore` to accept the same set of arguments - `(am_I_Root, Input_Opt, State_Chm, State_Diag, RC)`

This is done in preparation for moving USE-associated arrays that are dependent on grid sizing (i.e. `(IIPAR, JJPAR, LLPAR)`) in these modules
to `State_Chm`, permitting multi-domain GC-WRF coupling and, in a broader sense, running simultaneous instances of GEOS-Chem modules in the same CPU.

**Important Note: GCHP Code Update Necessary**
This commit changes the entry point for `Init_TOMS` which is being called in GCHP by `gigc_chunk_mod.F90`, breaking compilation for GCHP. 

**Solution:** That entry point should be *removed* - please refer to the following code context taken from `gigc_chunk_mod`, around lines 305:

```fortran
    ! Initialize other GEOS-Chem modules
    CALL GC_Init_Extra( am_I_Root, HistoryConfig%DiagList, Input_Opt,    &
                        State_Chm, State_Diag, RC ) 
    ASSERT_(RC==GC_SUCCESS)

    ! ...

    ! Allocate array of overhead O3 columns for TOMS if chemistry is on
    IF ( Input_Opt%LCHEM ) THEN
       CALL Init_TOMS( am_I_Root, Input_Opt, RC )
       ASSERT_(RC==GC_SUCCESS)
    ENDIF
```

Within `GC_Init_Extra` (from `GC_Environment_Mod`), the following already exists:

```fortran
    !-----------------------------------------------------------------
    ! Initialize "toms_mod.F"
    !-----------------------------------------------------------------
    CALL Init_Toms( am_I_Root, Input_Opt, State_Chm, State_Diag, RC )
    IF ( RC /= GC_SUCCESS ) THEN
       ErrMsg = 'Error encountered in "Init_Toms"!'
       CALL GC_Error( ErrMsg, RC, ThisLoc )
       RETURN
    ENDIF
```

Which runs independent of `Input_Opt%LCHEM` selection. Therefore I believe that in the GCHP source, the `Init_TOMS` code is redundant and should be removed. It does not throw a duplicate allocation error, because there are `ALLOCATED` sanity checks within `Init_TOMS` code.

### DIFF
--- a/GeosCore/aerosol_mod.F
+++ b/GeosCore/aerosol_mod.F
@@ -2501,7 +2501,8 @@
 !\\
 ! !INTERFACE:
 !
-      SUBROUTINE Init_Aerosol( am_I_Root, Input_Opt, State_Diag, RC )
+      SUBROUTINE Init_Aerosol( am_I_Root, Input_Opt, 
+     &                         State_Chm, State_Diag, RC )
 !
 ! !USES:
 !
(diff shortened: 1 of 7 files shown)


msulprizio:
These changes passed diff tests and unit tests

PR #5, opened 2018-08-09
Add the MODEL_, MODEL_WRF pre-processor flags

jimmielin:
The added pre-processor flags `MODEL_`, `MODEL_WRF` are now used where "GCHP"-related changes are not ESMF specific.

This update adds two new pre-processor flags: `MODEL_` and `MODEL_WRF`.

The former is used in many places along with the `ESMF_` flag, with the intent of
replacing it (but not done for now, to keep compatibility with the existing GCHP codebase) where code is not ESMF-specific (does not use MAPL/ESMF) but instead
merely intended for GEOS-Chem to interface with an external model (e.g. accept
external date/time, pressure edges). This allows WRF-GC to compile without the `ESMF_` flag, which was essentially lying to the compiler as WRF-to-GEOS-Chem coupling does not use ESMF.

The latter `MODEL_WRF` flag is intended for future uses in the WRF-GC interface.

### DIFF
--- a/GeosCore/cleanup.F
+++ b/GeosCore/cleanup.F
@@ -71,17 +71,15 @@
       USE TOMS_MOD,                ONLY : CLEANUP_TOMS
       USE TPCORE_FVDAS_MOD,        ONLY : EXIT_TPCORE
       USE TPCORE_WINDOW_MOD,       ONLY : EXIT_TPCORE_WINDOW
-#if ! defined( ESMF_ )
+#if !defined( ESMF_ ) && !defined( MODEL_ )
       USE TRANSPORT_MOD,           ONLY : CLEANUP_TRANSPORT
 #endif
       USE UCX_MOD,                 ONLY : CLEANUP_UCX
       USE VDIFF_PRE_Mod,           ONLY : Cleanup_VDIFF_PRE
       USE WETSCAV_MOD,             ONLY : CLEANUP_WETSCAV
 
-      ! HEMCO
-!#if !defined(ESMF_) 
+      ! HEMCO 
       USE EMISSIONS_MOD,           ONLY : EMISSIONS_FINAL
-!#endif
 
 #if   defined( RRTMG )
       USE RRTMG_RAD_TRANSFER_MOD,  ONLY : CLEANUP_SURFACE_RAD
@@ -369,7 +367,7 @@
       CALL CLEANUP_SULFATE()
       CALL CLEANUP_STRAT_CHEM()
       CALL CLEANUP_TAGGED_CO()
-#if ! defined( ESMF_ )
+#if !defined( ESMF_ ) && !defined( MODEL_ )
       CALL CLEANUP_TRANSPORT()
 #endif
       CALL CLEANUP_TOMS()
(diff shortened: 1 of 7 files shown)


msulprizio:
Thanks for these updates, Jimmie! We are going to announce the release of 12.0.0 tomorrow (Aug 10) so we'd like to hold off on merging in any more updates at the time. We can merge these changes into Dev next week into what will be either 12.0.1 or 12.1.0 (depending on if a fix or benchmark update is ready first).

jimmielin:
Thanks for the heads up! 

I'm also preparing a fairly large update which migrates in-module arrays to `State_Chm`, which due to the significant architectural change will certainly need some more testing before it goes in, so it's great to know things will be merged into a future release instead. I'll submit them as PRs when ready so you can merge at a convenient time. Thanks!

lizziel:
Hi Haipeng and GCST,

I am looking closer at this update as I am updating GEOS-Chem for use with GEOS-5 and encountered a conflict between use of MODEL_GEOS and new flag MODEL_.  I see here that the intention of MODEL_ is to consolidate flags for all external models to use in conjunction with if ESMF. This led me to wonder if we are supposed to be compiling GEOS-5 with MODEL_. It looks like that is not the case yet but I find its presence confusing.

For clarity at this time I think we should stick with using MODEL_{name} rather than MODEL_. Each instance could be followed by comments in the code on the reasoning behind its use and the potential for generalization in the future. For example, rather than MODEL_ we would have MODEL_WRF with a brief note that it is there because of some feature (or lack of feature) in WRF that may also be applicable to other models interfacing with GEOS-Chem. Over time we may accumulate instances of MODEL_WRF || MODEL_CESM, etc. Eventually we can survey all of these instances and come up with a strategy for generalizing across models. 

Does this sound okay? Do you have any other ideas?

Lizzie

jimmielin:
Hi Lizzie and GCST,

Thanks for raising the issue. The original intent was to consolidate common external-model related code (e.g. accept external pressure edges, inhibiting the compilation of main.F, etc...) into a new `MODEL_` constant, which GEOS-5-GEOS-Chem, WRF-GC, CESM-GC, ... would be compiled with, plus their specific model flag (e.g. `MODEL_GEOS`, so `-DMODEL_ -DMODEL_GEOS` would be what you would see in the compile command).

When adding these constants I took care of checking whether the implementation was ESMF-specific or not, and if not, it went into the `MODEL_` constant (in addition to `ESMF_`: see note below), otherwise we kept it as `ESMF_`.

There is another interesting question which might be worth discussing. While I suggest GEOS-5-GEOS-Chem be compiled with `MODEL_` and `MODEL_GEOS`, GCHP itself isn't an "external model" and is sort of a special case and another version of GEOS-Chem. This is why I kept the `ESMF_` flag, because GCHP isn't meant to be compiled with the `MODEL_` flag (as it's just another version of GEOS-Chem.) This raises the question of (1) whether GEOS-5 should be compiled with `ESMF_` flag in addition to the prior two and (2) whether we should abstract the GCHP implementation as "another external model" and treat it similarly to GEOS-5 instead, maybe having `MODEL_GCHP`. (As I'm not sure whether GCHP & GEOS-5 implementations diverge significantly to the point where you cannot compile both with `ESMF_` on)

This is all a little confusing and depends on the specific needs of the models, and I'd love to discuss this further. Please let me know your thoughts, thanks!

Haipeng

@lizziel

PR #10, opened 2018-08-17
Adds Accept_External_ApBp for MODEL_WRF

jimmielin:
This update allows `PRESSURE_MOD` to accept external `AP`, `BP` parameters
for hybrid grid from an external model (WRF), as WRF uses different
hybrid grid parameters that are runtime-determined.

### DIFF
--- a/GeosUtil/pressure_mod.F
+++ b/GeosUtil/pressure_mod.F
@@ -38,6 +38,10 @@
 #if defined( ESMF_ ) || defined( MODEL_ )
       PUBLIC  :: Accept_External_Pedge
 #endif
+
+#if defined( MODEL_WRF )
+      PUBLIC  :: Accept_External_ApBp
+#endif
 !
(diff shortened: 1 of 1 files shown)


jimmielin:
Hi,

I'd also like to add a note for this. Right now `PRESSURE_MOD`'s `ACCEPT_EXTERNAL_PEDGE` accepts the external pressure edges from `State_Met%PEDGE`, as that is a defined 3-D field in `State_Met`. 

The reason we want `PRESSURE_MOD` to accept external `Ap`, `Bp` values for hybrid-grid is because WRF allows run-time customization of the vertical levels, so the `Ap`, `Bp` values are run-time determined and passed to GEOS-Chem at initialization time. This might only be relevant for WRF, so we did not add a `State_Met%` field only for `MODEL_WRF`, instead just passing in arrays `ApIn`, `BpIn`.

PR #17, opened 2018-08-23
Add NO_EXE=y option to Makefile to inhibit geos executable build

jimmielin:
This update to the Makefiles adds a new option `NO_EXE=y` to the Makefile,
for other models coupled to and driving GEOS-Chem from their routines.

(Hence the `geos` executable would not be used and creating it in the G-C build process
may cause confusion to the user)

When `NO_EXE=y` is enabled, the final `geos` executable is not built,
and `GeosCore` will build all prerequisite libraries leaving `libGeosCore.a`
in the `lib` folder instead for other models to link to in their compile
process.

When `NO_EXE=y` is enabled, the pre-processor directive `MODEL_` is
automatically set as GEOS-Chem must be operating under another model
when this option is enabled.

### DIFF
--- a/GeosCore/Makefile
+++ b/GeosCore/Makefile
@@ -161,6 +161,8 @@
 #  09 Aug 2016 - M. Sulprizio- Remove diag_pl_mod and add diag20_mod
 #  13 Jul 2017 - E. Lundgren - Remove passive_species_mod.F90
 #  24 Aug 2017 - M. Sulprizio- Remove support for GCAP, GEOS-4, GEOS-5 and MERRA
+#  23 Aug 2018 - H.P. Lin    - If NO_EXE is set (for coupled model driving G-C)
+#                              do not build "geos" but instead "libGeosCore.a"
 #EOP
 #------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


msulprizio:
This pull request required an additional fix.

commit d45ff1248d7aa3c92e162279fdb919a40fca8354 (HEAD -> dev/12.1.0, origin/dev/12.1.0, or
Author: Melissa Sulprizio <mpayer@seas.harvard.edu>
Date:   Tue Aug 28 08:19:09 2018 -0400

    Add fixes for NO_EXE option
    
    The GEOS-Chem executable was not getting created and modules in
    GeosCore were not compiled, even when not passing NO_EXE=y. To fix
    this, we now define a variable EXE_NEEDED in Makefile_header.mk to
    pass  to GeosCore/Makefile. The syntax in GeosCore/Makefile has
    been fixed to use ifeq($(EXE_NEEDED),1) instead of ifdef NO_EXE
    which will not work properly.
    
    Signed-off-by: Melissa Sulprizio <mpayer@seas.harvard.edu>

PR #18, opened 2018-10-31
Extend CFC Surface Mixing Ratio fix in 12.0.3 for GCHP to WRF-GC

jimmielin:
This update extends the fix to the bug in application of CFC surface mixing ratios in GCHP implemented in GEOS-Chem 12.0.3 (http://wiki.seas.harvard.edu/geos-chem/index.php/GEOS-Chem_12#Fix_bug_in_application_of_CFC_surface_mixing_ratios_in_GCHP) to WRF-GC by adding the `MODEL_WRF` pre-processor flag to the relevant fixes in addition to `MODEL_GEOS`.

WRF-GC which uses a similar interface into G-C as GCHP, the `Get_Year()` function does NOT return the correct simulation date (initialized as a dummy) until the GEOS-Chem initialization routine is finished. Thus this fix for GCHP will apply for WRF-GC and thus should be extended.

Thanks for the original authors for discovering and developing this fix.

### DIFF
--- a/GeosCore/ucx_mod.F
+++ b/GeosCore/ucx_mod.F
@@ -1180,7 +1180,7 @@
 
       ! Local variables for quantities from Input_Opt
       LOGICAL              :: LPRT
-#if defined( MODEL_GEOS )
+#if defined( MODEL_GEOS ) || defined( MODEL_WRF )
       INTEGER              :: TARGYEAR
 #endif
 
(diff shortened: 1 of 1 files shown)


lizziel:
Hi Haipeng,

The 12.0.3 fix for GCHP is actually outside of the MODEL_GEOS blocks. We define MODEL_GEOS only for use in GEOS-5, not for GCHP. I preserved Christoph's implementation pending testing of our 12.0.3 fix in GEOS-5 but it will be removed in the future. However, both implementations now call GET_YEAR in Emiss_Basic rather than in init_ucx so the time information should already be initialized by the time it gets called.

Have you tried running with the CFC fix without adding in your MODEL_WRF ifdefs? I would like to know if it does not work to correct the issue in WRF prior to accepting this pull request. Seb is going to be doing some additional work on the CFC handling in ucx_mod so this information would also be helpful for that.

Thanks,
Lizzie

jimmielin:
Hi Lizzie,

It seems that I got confused then! I recently switched WRF-GC to track the `dev/12.1.0` branch instead of the old `Dev` branch, so we didn't have any fix for this before. Looking closer into the fix it seems like it should cover WRF-GC without the ifdef statements (sorry for not looking closer into the code changes!), but I'll perform some testing without the ifdefs to make sure. I'll update here when done.

Thanks!
Haipeng

PR #21, opened 2018-11-24
Update planeflight_mod.F

ngc2244:
Skip flight observations outside the nested domains. Now the code works for all 0.5 and 0.25 nested domains. LAT and LON ranges are defined as in http://wiki.seas.harvard.edu/geos-chem/index.php/GEOS-Chem_horizontal_grids.


msulprizio:
Thanks for this update, Lei. We can include it in 12.2.0.

msulprizio:
Now in `dev/12.2.0` branch

PR #26, opened 2019-01-21
Grid-independent lightning NOx emission code updates.

ltmurray:
Lightning flash densities (flashes km-2 s-1) are now calculated offline at the native GEOS-FP or MERRA-2 model resolution, as well as the depth of convection. The offline flash density calculation and its climatological constraint to the LIS/OTD satellite climatology are otherwise as described in Murray et al. (2012).

Flash densities and convective depths are now specified as external meteorological fields in HEMCO_Config.rc and re-gridded to the simulation resolution, creating consistent lightning NOx emissions across all model configurations.


msulprizio:
Now in the **dev/12.4.0** branch

PR #27, opened 2019-01-24
Move domain dependent variables from Linoz and MODIS LAI

jimmielin:
This update moves linoz_mod `TLSTT`, modis_lai mod `XLAI2`, `XCHLR2`
module variables to the `State_Chm` and `State_Met` derived type objects.

It disables the dust archive archival for WRF-GC in dust_mod
by using the `ESMF_` pre-processor switch instead of the
generic `EXTERNAL_GRID` switch.

This series of updates is provided as a part of restructuring GEOS-Chem
for integration into the WRF-GC coupled model with multi-domain support.

This update has been verified through GEOS-Chem Difference Tests against
the current `dev/12.2.0` latest commit.


msulprizio:
Now merged into `dev/12.2.1`

PR #30, opened 2019-02-20
Prevent repeat allocation of dry deposition module variables.

jimmielin:
This is a minor code update for INIT_DRYDEP to skip variable allocation
if it has already been allocated once. This is correct because all the
grid-dependent variables in DRYDEP_MOD have been moved to State_Chm,
thus the dry deposition module only contains data generic to the species
list, which is consistent in a given run and does not need to be read
more than once.

This scenario is only possible in the multiple domain configuration of
WRF-GC model, and does not affect other GEOS-Chem configurations at this
time.

### DIFF
--- a/GeosCore/drydep_mod.F
+++ b/GeosCore/drydep_mod.F
@@ -4137,6 +4137,15 @@
       ErrMsg    = ''
       ThisLoc   = ' -> at Init_Drydep (in module GeosCore/drydep_mod.F)'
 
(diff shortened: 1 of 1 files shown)


msulprizio:
Thanks, Haipeng! I'm currently doing a diff test on this update to ensure if doesn't introduce any changes to model output. If it passes, we can bring this into 12.2.1.

PR #23, opened 2018-12-11
Add flexgrid_read_mod.o to list of dependencies for hcoi_gc_main_mod.o

jimmielin:
This is a minor fix for an error in compiling GEOS-Chem Classic that I noticed in my test build of GCC (in the run directory, `make mpbuild`)

The compile process fails occasionally due to the `Get_Met_Fields` subroutine in `HCOI_GC_Main_Mod` requiring `FlexGrid_Read_Mod` in a `use` statement, but this module was not included in the `Makefile` dependencies. This is fixed by this commit.

The error would look something like this: (Added so future Google searches can find this fix)
```
ifort -cpp -w -auto -noalign -convert big_endian -O2 -vec-report0 -fp-model source -openmp -mcmodel=medium -shared-intel -traceback -DLINUX_IFORT -DBPCH_DIAG -DNC_DIAG -DBPCH_TIMESER -DGEOS_FP -DGRID2x25 -DN
C_HAS_COMPRESSION -DUSE_REAL8 -module ../mod -I/home/hplin/gc/include -I/home/hplin/gc/include -c -free hcoi_gc_main_mod.F90
hcoi_gc_main_mod.F90(3492): error #7002: Error in opening the compiled module file.  Check INCLUDE paths.   [FLEXGRID_READ_MOD]
   USE FlexGrid_Read_Mod
-------^
compilation aborted for hcoi_gc_main_mod.F90 (code 1)
../Makefile_header.mk:1520: recipe for target 'hcoi_gc_main_mod.o' failed
make[4]: *** [hcoi_gc_main_mod.o] Error 1
```

Thanks!

Signed-off-by: Jimmie Lin <jimmie.lin@gmail.com>

PR #35, opened 2019-04-29
Add support for CMake

LiamBindle:
Hi everyone,

This is a pull request (PR) for the changes that implement CMake support into GEOS-Chem. This PR adds about 20 new files but doesn't modify any preexisting files or the Makefile structure. These files implement a new modular build system that can compile GEOS-Chem Classic, or be controlled to build `libGeosCore` for dependents like GCHP.

These changes have little effect on GEOS-Chem Classic users (GEOS-Chem Classic is already easy to build) but are the basis for ~~PR [#23](https://github.com/geoschem/gchp/pull/23)~~ PR [#27](https://github.com/geoschem/gchp/pull/27) (on GCHP) which aims to improve the ease and transparency of building GCHP. Because these changes have a wider bearing than GCHP, I thought here (this PR) was the most appropriate place to discuss these changes rather than in my GCHP PR.

#### Motivation

A priority of the GCHP Working Group has been making the model easier to use. Particularly for new users, compiling GCHP can be challenging because your operating environment needs to be configured properly and the build process is rather hard to follow (the log files are very verbose and not necessarily clear). 

I think that these user-level challenges stem from the Makefile structure being difficult to maintain for large codebases like GCHP (GCHP contains more than 3000 Fortran, C, and C++ files compared to GEOS-Chem Classic which has 365 Fortran files). Makefile structures rely on dependents calling the Makefiles of their dependencies. This forms a monolithic structure with internal relationships that are difficult to maintain for dependencies with complex build configurations like GCHP's MAPL dependency. With a CMake-based build system, these internal relationships are maintained automatically by the CMake engine, rather than by the manual effort of developers.

#### Task list

Below is a list of tasks I need to be completed before this PR is ready to merge.

- [x] User-level documentation (my in-progress draft is [here](http://fizz.phys.dal.ca/~atmos/wiki/index.php/Building_GCHP_with_CMake_(draft)))
- [x] Developer documentation (inline)
- [x] Verify that compiler definitions are correct for each run directory (see [here](<http://fizz.phys.dal.ca/~atmos/wiki/index.php/Verification_of_the_CMake-based_build_system#Preprocessor_definitions>))
- [x] Verify that compiler options are the same for each run directory (see [here](http://fizz.phys.dal.ca/~atmos/wiki/index.php/Verification_of_the_CMake-based_build_system))
- [x] Verify that the executable for each run directory compile (see [here](http://fizz.phys.dal.ca/~atmos/wiki/index.php/Verification_of_the_CMake-based_build_system))
- [x] Run important diff tests (see [here](http://fizz.phys.dal.ca/~atmos/wiki/index.php/Verification_of_the_CMake-based_build_system))
- [x] Merge HEAD into my branch

#### Updates to this PR

I'm going to be updating this PR regularly so this is the best place to track my progress. Please let me know if you have any questions or concerns.

Liam


JiaweiZhuang:
Thanks for this important work! MAPL's current Makefile is such a mess, and I'd really love to see it being replaced.

Quick comments: 
- Should the documentation be maintained in a more version-controllable environment, like a GitHub repo? I am not sure how the Dalhousie wiki organizes the files internally. If they are just static MarkDown files then it is easy to migrate; WordPress-like dynamic sites are a bit more annoying.
- Is it possible to do the merging incrementally instead of pulling everything all at once? Although you are not touching existing files, I am still concerned about diverging & lagging model versions. Mixing Make and CMake seems a horrible idea, and I am not sure what would be the "minimum working unit" of this PR...

LiamBindle:
No problem! I'm excited it's coming together!

###### RE: Where to put documentation
Yeah, I'm happy to write the docs in whatever format is most convenient. The Dal wiki and SEAS wiki both use Wikitext so I think that migrating between the two should be easy, but I don't mind writing the docs in something else. Would you suggest something different? 

###### RE: Merging incrementally 
If I understand correctly, you're asking if we can merge these CMake changes but continue using the existing Make-based build system until a later date. That is correct. These changes don't have any effect on the Make-based build system (building is done out of source with CMake so the Makefiles generated by CMake don't conflict with the preexisting Makefiles in GEOS-Chem). 

Did I understand correctly?

JiaweiZhuang:
> The Dal wiki and SEAS wiki both use Wikitext so I think that migrating between the two should be easy

I guess the [GCHP doc](http://wiki.seas.harvard.edu/geos-chem/index.php/GEOS-Chem_HP) is the proper place for this. It is also Wikitext-based so should be fine!

> If I understand correctly, you're asking if we can merge these CMake changes but continue using the existing Make-based build system until a later date.

I actually mean migrating each subdirectory (e.g. `GeosCore`, `GCHP`, `MAPL_Base`) to CMake and **mix** it with the existing Make-based build system, so the migration/refactoring can be done incrementally. Each stage will get its own CI and diff tests, which is probably easier to debug (the sub-problem size is smaller). 

Mixing two build systems looks like a horrible idea but we are sort of doing so right now? I don't think you will ever touch ESMF's GNU-Make-based build system?

LiamBindle:
> I guess the GCHP doc is the proper place for this. It is also Wikitext-based so should be fine!

Sounds good!

> I actually mean migrating each subdirectory (e.g. GeosCore, GCHP, MAPL_Base) to CMake and mix it with the existing Make-based build system, so the migration/refactoring can be done incrementally. Each stage will get its own CI and diff tests, which is probably easier to debug (the sub-problem size is smaller).

Oh, sorry, now I think I understand what you mean. By "mixing" the two build system, do you mean something like for example: implement CMake files in `GeosCore/`, integrate the Makefiles generated those by CMake in in `GeosCore/` into the existing Make-based build system, and then run CI tests on those changes? If so, **unfortunately, I don't think there is a good way to do this** (please correct me if you know otherwise). This is because any subset of these 20 files can't actually do anything on their own (CMake generates a complete Makefile structure for the entire project and this Makefile structure is intended to be treated like a black box&mdash;working with those generated Makefiles would be challenging and dangerous). Because of this, I think the "minimum working unit" is this the 20 files included in this PR (open to suggestions though!).

If I've misunderstood you, could you clarify what you mean by "mix" the build systems?

If I do understand correctly, are you getting at the question of how we go about verifying the new CMake-based build system in manageable increments? This is a really good question, and something we should talk about. Lately I've been putting a lot of thought into this question, and at some point I needed to write it out, so I've done that below (see [here]("My current thoughts on incrementally verifying the CMake-based build system")).

Before that though:

> I don't think you will ever touch ESMF's GNU-Make-based build system?

That's correct. **Moving forward, I think ESMF should be considered an external dependency** (similar to NetCDF-Fortran) for the following reasons:

1. ESMF doesn't plan to add CMake support&mdash;so mixing their build system with our new CMake-based build system is difficult if it's possible at all (we could wrap it as an external executable process, but IMO that's an antipattern). 
2. ESMF is pretty easy to compile as it is. 
3. Even GCHP developers are unlikely to modify the ESMF source code, so there's little need to recompile/reinstall it (please correct me on this if I'm wrong though).

Any thoughts on this?

------

Okay, on to my current thoughts on verifying the CMake-based build system.

#### My current thoughts on incrementally verifying the CMake-based build system

Here are some steps that I'll put out as initial ideas for verifying the CMake-based build system. Comments and suggestions would be greatly appreciated. I think it's likely that a fresh set of eyes could pick out places where my methodology could be improved.

CI testing is something I think/hope will be easier with the CMake-based build system, but this initial verification is tricky and my ideas require a significant amount of manual effort. The CMake files are pretty short and simple&mdash;it's the testing that's tricky IMO. Anyways, here are my ideas:

1. **Check all necessary files are being compiled**.  This one is easy. Missing files will lead to unresolved symbols which will result in link errors.

2. **Check that preprocessor definitions for each build directory are the same between the two build systems**. This one is involved and critical. Preprocessor definitions are used extensively to activate and disable blocks of code. I have written a little tool to help with verifying this step: this tool compiles the `geos.mp` executable for each run directory independently with the Makefile-based build system and the CMake-based build system, scrapes their build logs for preprocessor definitions, and then compares those definitions.

3. **Check that compiler options/flags are the same between the two build systems**. This is also somewhat involved but less critical than 2 because flags don't affect *the source code*. I'm not really sure what the best way to check these are&mdash;I could scrape build logs for compiler options/flags but I'm not sure if this is really necessary. Additionally, most compiler options are specified in [lines 1063-1454](https://github.com/geoschem/geos-chem/blob/b2fb3ecf0956ef8cfd1180064841e3cb2583ad47/Makefile_header.mk#L1063-L1454) of `Makefile_header.mk` (besides the odd file/directory specific need to make `REAL` 8 bytes wide).

4. **Diff tests**. Again, pretty involved and absolutely critical. Additionally, these are complicated by the fact the Make-based build system and CMake-based build system are not compatible. My current approach is to run 1-day simulations and `ncdiff` (from NCO) the resulting restart files from each build system's executable. Because (in theory) the exact same code is being compiled by the exact same compiler with the same definitions, flags, and options very minor (if any) differences should be expected.

   Note: A few months ago I ran successful diff tests (not first try, but after resolving some runtime errors) for `geosfp_4x5_standard` and `gchp_standard`. Not that that means much, but I usually try to follow a test-driven workflow and it just feels relevant to mention.

Those are my current thoughts on verifying the CMake-based build system in relatively bite-sized chunks. 

Thanks in advance!

JiaweiZhuang:
> By "mixing" the two build system, do you mean something like for example: implement CMake files in `GeosCore/`, integrate the Makefiles generated by those CMake files (in `GeosCore/`) into the existing Make-based build system, and then run CI tests on those changes?

Exactly.

> **unfortunately, I don't think there is a good way to do this** (please correct me if you know otherwise). This is because any subset of these 20 files can't actually do anything on their own (CMake generates a complete Makefile structure for the _entire project_ and this Makefile structure is intended to be treated like a black box).

This generally true. But it should also be possible to treat a sub-component as a CMake External Project? That External Project will be built with the original GNU Make. Then CMake will use that pre-compile component from a higher-level. I guess this is what you need to do for the ESMF directory? Maybe other sufficiently-independent components (say GIGC, FV3) can be treated in a similar way?

Ref:
- https://cmake.org/cmake/help/latest/module/ExternalProject.html
- https://stackoverflow.com/questions/15175318/cmake-how-to-build-external-projects-and-include-their-targets
- https://mirkokiefer.com/cmake-by-example-f95eb47d45b1

> If I do understand correctly, are you getting at the question of how we go about verifying the new CMake-based build system in manageable increments?

Right. Instead of being incremental in terms of source code directories, it is probably more sane to be incremental in terms of verification steps, as you've suggested later. Just pick any way that is most convenient for you. 

> **Moving forward, I think ESMF should be considered an external dependency**

Yes, this is a must (geoschem/gchp#7). Also, Spack can already install ESMF (spack/spack#2831, spack/spack#2809), which further simplifies the build a lot. But before that happens you would need to treat the ESMF subdirectory as an external project? Or should we swap out ESMF first so you can make progress?  


> My current thoughts on incrementally verifying the CMake-based build system

Thanks for the detailed plan!  The Diff tests are crucial but are more like benchmarks than CI. It is expensive to run and cannot be automated by online CI services. 

Also, should we expect the executable (i.e. `geos`) to be bit-wise identical in CMake vs Make builds? This seems a quick alternative to actually running the diff tests. A more fine-grained way is to compare each object file `*.o` (which will reflect any changes in compiler options/flags)

Ref:
- https://stackoverflow.com/questions/12118403/how-to-compare-binary-files-to-check-if-they-are-the-same

LiamBindle:
> But it should also be possible to treat a sub-component as a CMake External Project? That External Project will be built with the original GNU Make.

That's a possibility! As you've mentioned later on though, I think it probably makes more sense to verify the new build system in steps as opposed to by directory because most CMakeLists.txt are just boilerplate commands (e.g. see [GeosUtil/CMakeLists.txt](<https://github.com/LiamBindle/geos-chem/blob/feature/CMake/GeosUtil/CMakeLists.txt>)).

>  I guess this is what you need to do for the ESMF directory?

> But before that happens you would need to treat the ESMF subdirectory as an external project?

It's possible to build ESMF as an External Project, and it's something I've explored, but IMO it's a bit cleaner have it as it's own step in [GCHP Hardware and Software Requirements](<http://wiki.seas.harvard.edu/geos-chem/index.php/GCHP_Hardware_and_Software_Requirements>). When I explored this in the past I came to the following conclusions:

1. Building ESMF as an External Project is possible, but not fully automatable.
   - CMake 3.5's [`FindMPI.cmake`](<https://cmake.org/cmake/help/v3.5/module/FindMPI.html>) doesn't have a way to get the MPI library name, so the `ESMF_COMM` environment variable would have to be set by the user, or passed to CMake as an argument.
2. The GCHP build logs are clean and readable. It feels like a shame to pollute them with the output from ESMF's build.
3. External Projects are run at build time, so it is difficult to preemptively detect build configuration errors. 

Because of these, I opted to not make ESMF an External Project in my initial draft of the CMake-based build system. I don't feel strongly about this though, so I'm open to other opinions.

>  Maybe other sufficiently-independent components (say GIGC, FV3) can be treated in a similar way?

They could be, but I opted against this in my current draft. My reasoning was again that External Projects are run during the build step (i.e. `make`) rather than the configure step (i.e. `cmake`). This means errors (e.g. from missing dependencies or bad paths to dependencies) show up part way through your `make` command rather than when you're configuring CMake (i.e. during `cmake` commands).



> Also, should we expect the executable (i.e. `geos`) to be bit-wise identical in CMake vs Make builds? This seems a quick alternative to actually running the diff tests. A more fine-grained way is to compare each object file `*.o` (which will reflect any changes in compiler options/flags)

Ooo, I like this idea! But honestly, I don't know if we should expect them to be bitwise identical (I'll look into this). To build on this, I could probably extend that tool I wrote to use `readelf` and `nm` to inspect differences in the assembly of all the symbols in the built executables. Thanks for the idea!

JiaweiZhuang:
> I opted to not make ESMF an External Project in my initial draft of the CMake-based build system.

So we probably need to solve geoschem/gchp#7 before incorporating CMake. This is good to know.

> My reasoning was again that External Projects are run during the build step (i.e. `make`) rather than the configure step (i.e. `cmake`).

Yeah this seems a good reason. I agree with your current approach.

> To build on this, I could probably extend that tool I wrote to use `readelf` and `nm` to inspect differences in the assembly of all the symbols in the built executables.

Looking forward to your findings!

LiamBindle:
Hi everyone,

I've got `feature/CMake` to a place where I'm happy with it, and I think it's just about ready for review! Regarding the review, **what version of GEOS-Chem should GC-Classic's CMake support staged for**? I'm not sure how close 12.4 is to release, but if that is an option, I think I could rebase on `dev/12.4.0` and rerun my verification tests within a day or two (at which point it would be ready for review). If 12.5 would be better, I wait until that's closer.

**Note**: As of writing, `feature/CMake` is based at 12.3.2.

### What is "done"
CMake support for the following is included in this PR:
- Building GC-Classic for all 73 run directories
- GTMM standalone and GTMM coupled to GC-Classic
- HEMCO standalone
- Intel and GNU compiler support
- A basic set of instructions (which are [here](http://fizz.phys.dal.ca/~atmos/wiki/index.php/Building_GEOS-Chem_Classic_with_CMake_(draft)))
  - If anyone has a chance to look at the instructions, I'd really appreciate any feedback that you have!
  - Also, I wrote a tutorial that introduces the `cmake` command and goes over how to use it. You can find that [here](https://github.com/LiamBindle/An-Introduction-to-Using-CMake/wiki).

### What is not included in this PR

Below are the things that I haven't done as a part of this PR. Kind of continuing what @JiaweiZhuang and I were talking about earlier though, I'm not sure if these should be included in this PR.

- Difference tests depend on the Makefile-based build system ([Performing Difference Tests with GEOS-Chem](http://wiki.seas.harvard.edu/geos-chem/index.php/Performing_Difference_Tests_with_GEOS-Chem))
- Generating ProTex documentation (a day or two of work)
- I haven't done anything with the `Custom/` or `int/` subdirectories of KPP
- I haven't done anything with `PKUCPL/`
- Am I missing anything?

### What I have tested

As of 9d27642 my tests passed. Below is a description of what I test.

1. Checking that `geos.sp` build with the Makefile-based and CMake-based build systems output bitwise-identical restart files for 1-day `geosfp_4x5_benchmark` run with 1800s/3600s timesteps.

   - Intel and GNU compilers

   > ```
   > $  md5sum **/GEOSChem.Restart.20160702_0000z.nc4
   > 5041454f99cee13c01a58c6e95f4e383  gnu/cmake/geosfp_4x5_benchmark/GEOSChem.Restart.20160702_0000z.nc4 
   > 5041454f99cee13c01a58c6e95f4e383  gnu/ref/geosfp_4x5_benchmark/GEOSChem.Restart.20160702_0000z.nc4 
   > 5ecc65c3ddfe932d755fef9071f34a0a  intel/cmake/geosfp_4x5_benchmark/GEOSChem.Restart.20160702_0000z.nc4 
   > 5ecc65c3ddfe932d755fef9071f34a0a  intel/ref/geosfp_4x5_benchmark/GEOSChem.Restart.20160702_0000z.nc4
   > ```

   - For more details see [here](http://fizz.phys.dal.ca/~atmos/wiki/index.php/Verification_of_the_CMake-based_build_system#Quick_1-day_benchmark_comparison)

2. I compile `geos.mp` for for all 73 run directories with the Makefile-based and CMake-based build systems. Then I scrape the build logs and check that

   1. The same KPP directory was built
   2. The same compiler definitions were used
   3. The same compiler flags were used

   I do this with Intel and GNU compilers.

   - For more details see [here](http://fizz.phys.dal.ca/~atmos/wiki/index.php/Verification_of_the_CMake-based_build_system#Scraped_build_log_diffs)

### Status of my GCHP PR

- https://github.com/geoschem/gchp/pull/23
  - Close to being ready for review. 
  - BUT,  waiting for `feature/MAPL_v1.0.0`.
  - I need to review my changes to `GCHP/Shared`'s CMake files, and run my tests. A ballpark guess is that if `feature/MAPL_v1.0.0` was ready to be merged, it would take me a week or two to have GCHP's `feature/CMake` ready for review.

------

I'm looking for some guidance on how to proceed. Should I rebase on `dev/12.4.0` and rerun my tests so that GC-Classic's support can be added in 12.4? Or should I wait until 12.5?

Thanks,

Liam

lizziel:
Hi Liam, 12.5 would be a good version for this update to go into. 12.4 is currently being wrapped up. Once it is released I am going to merge in the MAPL v1.0.0 work into new branch dev/12.5.0. That is the branch that we would want to pull the cmake updates into. 

Thanks for your work on this! Will you be able to help out with documenting how to use cmake with GEOS-Chem on the wiki? This work will need to happen before the 12.5 release if we start supporting cmake in that version.

LiamBindle:
Hi Lizzie, sounds good. I'll try to get things ready to merge into `dev/12.5.0` then. 

>  Will you be able to help out with documenting how to use cmake with GEOS-Chem on the wiki?

For sure! 

I'll update this PR to go onto `dev/12.5.0` when it appears and mark it as ready for review.

Thanks!

Liam

LiamBindle:
Hi everyone,

I think that CMake support is ready for people to try out! :smile:

If anyone is interested in giving it a go, please do! I'd really appreciate any feedback or ideas that you have&mdash;anything and everything! It's no rush though. CMake support is scheduled for 12.5, so we have time.

### A few notes to get you started

- Instructions for building GC-Classic with CMake are [here](http://fizz.phys.dal.ca/~atmos/wiki/index.php/Building_GEOS-Chem_Classic_with_CMake_(draft))
- Instructions for building GCHP with CMake are [here](http://fizz.phys.dal.ca/~atmos/wiki/index.php/Building_GCHP_with_CMake_(draft))
- I wrote a tutorial that introduces people to the `cmake` command and goes over how to use it. I've been told it takes about 30-45 minutes to complete, and by the end of it, you should feel comfortable using `cmake`. That tutorial is [here](https://github.com/LiamBindle/An-Introduction-to-Using-CMake/wiki).
- CMake should work out of the box with your current GEOS-Chem environment. If you do this you can skip the "Setting things up" part of the instructions (i.e. start [here](http://fizz.phys.dal.ca/~atmos/wiki/index.php/Building_GEOS-Chem_Classic_with_CMake_(draft)#Getting_GEOS-Chem_Classic.27s_source_code)).
- The procedure for building GC-Classic and GCHP is the same (besides GCHP having additional external dependencies)
- You shouldn't be able to get *compiler* errors, even if you try. Any problems should be identified by `cmake` during the build configuration stage, and you should be able to resolve them with another `cmake` command.
- Please stress test it!

If you'd like a bit more help getting started, let me know when you plan to try it out and I'll make sure I'm on Slack.

**Note**: I'll be away June 7-11.

--Liam

lizziel:
Thanks Liam. I am going to try this out with GCHP today.

--
Lizzie Lundgren
Senior Scientific Programmer
GEOS-Chem Support Team
geos-chem-support@g.harvard.edu<mailto:geos-chem-support@g.harvard.edu>
http://wiki.geos-chem.org/GEOS-Chem_Support_Team

Please direct all GEOS-Chem support issues to the entire GEOS-Chem Support Team
at geos-chem-support@g.harvard.edu<mailto:geos-chem-support@g.harvard.edu>. This will allow us to serve you better.

lizziel:
Hi Liam, given CMake in GCHP will be delayed until after 12.5 pending the GCHP restructuring, we think it best to hold off on CMake in GC Classic to release at the same time. We can still test this now, however, but I want to give you the heads up that the version of release will be a bit later now. This is actually an opportunity since perhaps we could do some simple restructuring in the GEOS-Chem repo at the same time as GCHP.

LiamBindle:
Hi Lizzie,

Sure, that makes sense to me! And yeah, if we could do test-driving now though, that would be really helpful from my end. It's probably a good idea to keep CMake in mind for upcoming structural changes to GC-Classic and GCHP.

msulprizio:
Hi Liam -- I am starting work on this now, with the goal of including CMake in GEOS-Chem Classic in 12.6.0.

LiamBindle:
Hi Melissa -- sounds good! Let me know if you need help with anything.

msulprizio:
This is now live in https://github.com/geoschem/geos-chem/tree/dev/12.6.0. I still need to add APM option to CMake.

PR #37, opened 2019-05-24
Cloud pH diagnostic fix. Now to get accurate cloud pH divide the diag…

jmoch1214:
diagnostic fix for cloud pH. Divide Chem_pHCloud by the diagnostic Chem_isCloud to get the real cloud pH value.

### DIFF
--- a/GeosCore/sulfate_mod.F
+++ b/GeosCore/sulfate_mod.F
@@ -2713,6 +2713,7 @@
       ! Pointers
       REAL(fp), POINTER     :: Spc(:,:,:,:)
       REAL(fp), POINTER     :: pHCloud(:,:,:)
+      REAL(fp), POINTER     :: isCloud(:,:,:) ! jmm 3/1/19
       REAL(fp), POINTER     :: SSAlk(:,:,:,:)
       REAL(fp), POINTER     :: H2O2s(:,:,:)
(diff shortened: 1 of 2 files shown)


lizziel:
Thanks Jonathan. Your PR includes changing default State_Chm%CloudpH from 4.5 to 0.0. Will this update change any species results in a standard full chemistry run?

jmoch1214:
It shouldn't because the entire sulfur code won't be called if the
State_Chm%CloudpH stays at 0. So there won't be any calculated sulfite or
bisulfite or any other cloud chemistry because there is no cloud. What was
happening before is that when there is no cloud the cloud pH diagnostic is
set to 4.5, which heavily weights the time averaged value to 4.5 because
there is not always cloud in the box. By setting it at 0 and then dividing
be the State_Chm%isCloud value we can get the cloud pH value averaged for
when there actually is cloud.  The fact that the code doesn't crash is also
an example that this change doesn't affect and species and just the
diagnostic. For a pH of 0 you would have -Infinity H+, which would crash
the code. But since the code can run that is also evidence that it doesn't
affect anything outside of how the diagnostic is calculated.

msulprizio:
This is now in the `dev/12.5.0` branch and will be pushed to mater when 12.5.0 has been benchmarked and approved.

PR #41, opened 2019-06-25
simple parameterization for CO2 dependence of stomatal resistance

ayhwong:


msulprizio:
Thanks, Anthony! This update is now in the dev/12.6.0 branch and will be pushed to master when that version is benchmarked and approved. See:

* https://github.com/geoschem/geos-chem/tree/dev/12.6.0
* https://github.com/geoschem/geos-chem-unittest/tree/dev/12.6.0

PR #78, opened 2019-10-02
[PULL REQUEST] V12.4.0 sfc fix

tsherwen:
Set fixed PBL concs. using HEMCO instead of hardcoded values

This is effectively a structural and not science change, but hopefully will be useful to many people in the community and several people are already using earlier forms of this so it would be great to see it in the main code.

@sdeastham and @christophkeller  helped with the development. @ltmurray's  recent methane update is related but unaffected by this.

**---- What is it?**

(1) Any field starting with `SfcVMR_<species hame>` is evaluated by HEMCO, then the provide species is set in sfcvmr_mod.F90. Any species can now be added by this method with a single line in HEMCO_Config.rc.

https://github.com/tsherwen/geos-chem/tree/v12.4.0_SfcFix

(2) long-lived organohalogen species (CHCl3, CH3Cl, CHBr3, CH2Cl2) are fixed for years 1750-2014 using offline NetCDF made from the CMIP6 files [Meinshausen et al 2017]. The fixed of global methane has also be added using the CMIP6 files, where no data is included for the NOAA files compiled by Lee Murray (before 1979). 

Meinshausen, M., Vogel, E., Nauels, A., Lorbacher, K., Meinshausen, N., Etheridge, D. M., Fraser, P. J., Montzka, S. A., Rayner, P. J., Trudinger, C. M., Krummel, P. B., Beyerle, U., Canadell, J. G., Daniel, J. S., Enting, I. G., Law, R. M., Lunder, C. R., O'Doherty, S., Prinn, R. G., Reimann, S., Rubino, M., Velders, G. J. M., Vollmer, M. K., Wang, R. H. J., and Weiss, R.: Historical greenhouse gas concentrations for climate modelling (CMIP6), Geosci. Model Dev., 10, 2057–2116, https://doi.org/10.5194/gmd-10-2057-2017, 2017.

**---- Changes in the code**

This functionality is hardcoded in several places in the code, such as for long-lived species for organohalogens in UCX/the halogen scheme and for methane. Where this is now redundant, it has been removed. 

**---- Changes in the output**

Non-UCX Organohalogen species will now be set to appropriate and latitudinally resolved values for a given year. Previously only a handful of values were hardcoded. Methane will only change in historical simulations before 1979 (outside the window of the NOAA dataset from Lee).

The values for UCX species are now set from NetCDF files prepared by Seb from the existing files. As these values are already being used in the model elsewhere, this should not cause notable changes.


msulprizio:
This is now merged into the dev/12.7.0 branch. An internal 1-month benchmark evaluating the impact of this update results in small changes in many bromine and chlorine species species (e.g. CH3Br, CH3Cl, CH3Cl2, HFCFs), consistent with the changes in output section provided above. These differences can be seen in the difference plots included below.

- [Bromine_Surface.pdf](https://github.com/geoschem/geos-chem/files/3956171/Bromine_Surface.pdf)
- [Chlorine_Surface.pdf](https://github.com/geoschem/geos-chem/files/3956182/Chlorine_Surface.pdf)
- [GC_12.7.0_SfcFix_OH_metrics.txt](https://github.com/geoschem/geos-chem/files/3956185/GC_12.7.0_SfcFix_OH_metrics.txt)

yantosca:
Hi Tomas,

Just wanted to let you know of a few issues with the data files for GCHP:

1. The SfcFix/GMI/surface_VMRs.nc and SfcFix/UCX/surface_VMRs.nc files had a time field with units of int64.  This is OK for GEOS-Chem "Classic" but the MAPL library cannot read integer data.  So this caused an error reading the data into GCHP.

2. We would prefer data files with one year per file, instead of one file with all years from 1750-2014.  This will facilitate the new GEOS-Chem "dry-run" capability that will ship with GC 12.7.0.  The "dry-run" will allow users to only download the data files that will actually be used for a given simulation.

3. I think the reference date of 1750-01-01 may be problematic for GCHP.  The calendar system changed from Julian to Gregorian in October 1752, so I am not sure how robust the MAPL library is in deciphering this transition.  Best to use a later date (or ideally, if splitting files into one file per year, use the 00 UTC at Jan 01 of the given year as the reference date.

4. Also, we were wondering why the data is in hours since 1750 instead of days since.

lizziel:
HI Tomas,

Adding on to what Bob noted above, it would be helpful if the data was broken up into annual files that each contain all species for that year. So rather than all years in a species file, all species in a year file. This has two main benefits:
1. Users can download only for the years they need
2. The model will only need to open and close one file for a given year rather than a different file for each species. Opening and closing files comes with performance overhead.

tsherwen:
Hello Bob and Lizzie, 

Sorry for the delayed response, we only have fewer than two workdays left this year and I am tied up with a dry run of planning for a research campaign. 

So to confirm, everything is working OK in GC classic but there are file read issues in GCHP/MAPL?

I have no objections to changing the NetCDF files to yearly/other requirements or putting all the species in one file. This won’t affect the science side of the update. 

To answer the specific points
* The UCX files are converted versions of the existing UCX files being used in GEOS-Chem from @sdeastham. If the existing files have been remade to meet this new MAPL requirement, maybe they could be shared with Seb so he can run his script to remake UCX/GMI files?
* The calendar was set to standard. The values are just one per month, so this choice won't really affect the data and this can follow whatever works best with GC/HEMCO/MAPL.  
* Hours since X appeared to be and still appears to be the recommend time stamp [on the wiki](http://wiki.seas.harvard.edu/geos-chem/index.php/The_COARDS_netCDF_conventions_for_earth_science_data#COARDS_coordinate_vectors). 

I just tried to follow the information that was on the wiki for these files (e.g. COARDS checker has been done etc), but that looks like it there is more information now. If the requirements/preferences you've mentioned here aren't there, please could we add a page with them to the wiki?

yantosca:
Hi Tomas,

Thanks for replying.  FYI, I have updated the files so that there is now one file per year with all species.  See:

1. http://ftp.as.harvard.edu/gcgrid/data/ExtData/HEMCO/CMIP6/v2019-12/README
2. http://ftp.as.harvard.edu/gcgrid/data/ExtData/HEMCO/SfcFix/v2019-12/README

Also, thanks for letting us know that "hours since" is on the wiki.  When splitting up data by year, we often do is to set the reference date to the first date in the file.  Also a reference date of 1750 might be problematic because of the Julian/Gregorian calendar switch in 1752.  At least, GCHP is a lot more stricter when it comes to reading input data than GCC.

We will update the wiki with more information.  Thanks for your feedback!

Thanks again and Happy Holidays!
Bob Y.

PR #88, opened 2019-10-11
Implement python-based alternative to bc for runConfig.sh

sdeastham:
Previously, setup would fail silently if the node did not have
bc ("bash calculator") installed when estimating NX and NY. Now,
if bc is not available, python will be used instead; and if
python is not found then an explicit error is thrown.

Signed-off-by: Sebastian D. Eastham <seastham@mit.edu>


lizziel:
Thanks for this update!

PR #100, opened 2019-10-19
Definitions for PM25 metadata occur two times

FeiYao-Edinburgh:
Definitions for PM25 metadata occur two times. I believe we only need one?


msulprizio:
Now merged into dev/12.7.0 branch

PR #101, opened 2019-10-21
PM25 calculation formula consistent with wiki

FeiYao-Edinburgh:
Would it be great to let PM25 calculation formula be consistent with that in [wiki](http://wiki.seas.harvard.edu/geos-chem/index.php/Particulate_matter_in_GEOS-Chem)?

### DIFF
--- a/GeosCore/aerosol_mod.F
+++ b/GeosCore/aerosol_mod.F
@@ -998,7 +998,7 @@
          !
          ! PM25 = 1.33 (NH4 + NIT  + SO4) + BCPI + BCPO +
          !        2.10 (OCPO + 1.16 OCPI) + 1.16 SOA    +
-         !        DST1 + 0.38 DST2 + 1.86 SALA + SOAGX + SOAMG
+         !        DST1 + 0.38 DST2 + 1.86 SALA
          !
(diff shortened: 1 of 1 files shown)


yantosca:
Thanks for writing.  I believe the PM2.5 definition was given to us by the Aerosols Working Group.  You might want to doublecheck with them before we change it in the code.  It might be easier to update the wiki than the code.

msulprizio:
The comment in aerosol_mod.F has now been changed to:

```
         !==============================================================
         ! P A R T I C U L A T E   M A T T E R
         !
         ! Compute PM2.5 concentration [kg/m3]
         !
         ! PM25 = 1.33 (NH4 + NIT  + SO4) + BCPI + BCPO +
         !        2.10 (OCPO + 1.16 OCPI) + 1.16 SOA*    +
         !        DST1 + 0.38 DST2 + 1.86 SALA
         !
         !   * If using simple  SOA, SOA = SOAS;
         !     If using complex SOA, SOA = TSOA + ASOA + ISOAAQ
```

This will go into 12.7.0 (see commit 9ef5b20c0a3f411bb51feb49f4321f9fb75637e0)

PR #122, opened 2019-11-03
[PULL REQUEST] Implement a dry-run option for GEOS-Chem classic.

jimmielin:
Hi all,

This commit implements the bulk of the GEOS-Chem "dry run" option as we've discussed and proposed it.
As this is a preliminary implementation I am opening this PR for review and discussion. Please note the list of current caveats:

* Works only for GEOS-Chem Classic. In GCHP, GC does not have access to IO code
* The files not being read via `hcoio` are critical and even dry run will crash: see https://github.com/geoschem/HEMCO/issues/3:
```
READ_LUT_NCFILE: Reading /home/ubuntu/ExtData/HEMCO/PARANOX/v2015-02/ship_plume_lut_02ms.nc
READ_LUT_NCFILE: Reading /home/ubuntu/ExtData/HEMCO/PARANOX/v2015-02/ship_plume_lut_06ms.nc
READ_LUT_NCFILE: Reading /home/ubuntu/ExtData/HEMCO/PARANOX/v2015-02/ship_plume_lut_10ms.nc
READ_LUT_NCFILE: Reading /home/ubuntu/ExtData/HEMCO/PARANOX/v2015-02/ship_plume_lut_14ms.nc
READ_LUT_NCFILE: Reading /home/ubuntu/ExtData/HEMCO/PARANOX/v2015-02/ship_plume_lut_18ms.nc
     - INIT_LIGHTNOX: Reading /home/ubuntu/ExtData/HEMCO/LIGHTNOX/v2014-07/light_dist.ott2010.dat
AeroCom: reading /home/ubuntu/ExtData/HEMCO/VOLCANO/v2019-08/2016/07/so2_volcanic_emissions_Carns.20160701.rc
```
(via @yantosca )
* `CHEM_INPUTS` is critical and missing these will crash dry run

To run GEOS-Chem with the dry run option, simply use:
```
./geos --dry-run
```

Implementation details:
* `--dry-run` is detected and flags `Input_Opt%DryRun` and `HcoState%isDryRun`
* `main.F` detects `%DryRun` and skips most of calculations in time-stepping
* `hcoio_read_std_mod.F90` detects `%isDryRun` and **does not read files**. It only checks if it exists and creates **empty data containers** with appropriate names and missing values. This is so other modules do not crash if they cannot find the data container.

I will attach sample outputs of the dry run and non dry-run option here or in a separate issue (apparently it cannot be done in a PR post) and see if I can adapt the automated scripts to read these instead (help welcome!)

Possible outputs from dry run:
* Regular geos-chem crash messages (we are running GEOS-Chem, so we can validate all GEOS-Chem input configs automatically 😉)
* `HEMCO: Opening <file name>` (if file exists)
* `HEMCO: REQUIRED file NOT FOUND <file name>` (not found)
* `HEMCO: OPTIONAL file NOT FOUND <file name>` (not found)
Note due to the nature of fortran output, there may be linebreaks inside the message and the file name, scripts will have to strip those.

Original commit message:
```
This update implements a dry-run option for GCC which can be
launched using the --dry-run command line argument. When GCC
is ran in a dry run option, GEOS-Chem will skip all computations
and only run initialization, check files (populating HEMCO with
empty data containers) and time-stepping to verify existence of
all required config, emissions and misc. files read through HCO.

The output of ./geos --dry-run can be fed into an automated
script to retrieve all necessary information.

This update is realized through patching of the main.F GCC program,
HCOIO_READ_STD_MOD, and multiple parts of HEMCO. It introduces
the Input_Opt%DryRun and HcoState%isDryRun option.

```

### DIFF
--- a/GeosCore/hcoi_gc_main_mod.F90
+++ b/GeosCore/hcoi_gc_main_mod.F90
@@ -53,7 +53,7 @@ MODULE HCOI_GC_Main_Mod
   PRIVATE :: CheckSettings
   PRIVATE :: SetHcoGrid
   PRIVATE :: SetHcoSpecies 
-#if !defined(ESMF_) && !defined( MODEL_WRF )
+#if !defined( ESMF_ ) && !defined( MODEL_WRF )
   PRIVATE :: Get_GC_Restart
   PRIVATE :: Get_Met_Fields
   PRIVATE :: Get_Boundary_Conditions
@@ -88,6 +88,7 @@ MODULE HCOI_GC_Main_Mod
 !  29 Nov 2016 - R. Yantosca - grid_mod.F90 is now gc_grid_mod.F90
 !  24 Aug 2017 - M. Sulprizio- Remove support for GCAP, GEOS-4, GEOS-5 and MERRA
 !  16 Jan 2019 - L. Murray   - Add offline lightning flash rates for LNOx emissions
+!  02 Nov 2019 - H.P. Lin    - Passthrough Input_Opt%DryRun for HEMCO DryRun
 !EOP
 !------------------------------------------------------------------------------
 !BOC
@@ -432,11 +433,19 @@ CONTAINS
     ! compared to a stand-alone version: in ESMF, the source file name
     ! is set to the container name since this is the identifying name
     ! used by ExtData.
-#if defined(ESMF_)
+#if defined( ESMF_ )
     HcoState%isESMF = .TRUE.
 #else 
     HcoState%isESMF = .FALSE.
 #endif
+
+    ! Are we running HEMCO in a dry-run mode?
+    ! This is dictated by the GEOS-Chem environment. If GEOS-Chem is in a dry-run
+    ! mode, no compute is performed and files are only "checked". Simulations will
+    ! NOT stop on missing files. This is intended to be a quick sanity check to make
+    ! sure that GEOS-Chem IO are all correctly set up, which is why most of the runs
+    ! fail to complete successfully. (hplin, 11/2/19)
+    HcoState%isDryRun = Input_Opt%DryRun
 
     ! Set deposition length scale. This determines if dry deposition
     ! frequencies are calculated over the entire PBL or the first
--- a/GeosCore/main.F
+++ b/GeosCore/main.F
@@ -245,6 +245,7 @@
       CHARACTER(LEN=255)       :: historyConfigFile
       CHARACTER(LEN=512)       :: ErrMsg
       CHARACTER(LEN=512)       :: Instr
+      CHARACTER(LEN=255)       :: Argv
 
 #if defined( RRTMG )
       !-----------------------------
@@ -437,6 +438,46 @@
       
       ! Debug output
       IF ( prtDebug ) CALL Debug_Msg( '### MAIN: a READ_INPUT_FILE' )
+
+      !-----------------------------------------------------------------
+      ! Prepare the GEOS-Chem "dry run" option
+      ! If in a "dry-run" mode, GEOS-Chem will simply check whether files
+      ! are present (and possibly in the correct format) and go through
+      ! time-steps to check met fields and other IO issues.
+      ! No actual "compute" is performed.
+      !
+      ! The "dry-run" option is initialized using the command line extra
+      ! argument ./geos --dry-run
+      ! 
+      ! Logs will be directly printed to standard output.
+      ! This option is currently only supported in GEOS-Chem Classic.
+      !
+      ! Additionally, this flag must be set after reading input file, or
+      ! its value will be overwritten by READ_INPUT_FILE.
+      ! (hplin, 11/1/19)
+      !-----------------------------------------------------------------
+      CALL GETARG( 1, Argv )   ! Need to allow for multiple argvs later
+      IF ( Argv == '--dry-run' ) THEN
+         Input_Opt%DryRun = .TRUE.
+      ENDIF
+
+      ! Are we running in "Dry-run" mode?
+      ! If Input_Opt%DryRun is set to TRUE, then GEOS-Chem is in a dry run mode.
+      ! No actual processing and calculations are performed and only time-stepping
+      ! through to verify config, input and output correctness is done.
+      ! (hplin, 11/1/19)
+      IF ( Input_Opt%DryRun ) THEN
+         WRITE( 6, '(a)' )
+         WRITE( 6, '(a)' ) REPEAT( '!', 44 )
+         WRITE( 6, '(a)' ) 'GEOS-CHEM IS IN A DRY RUN MODE!'
+         WRITE( 6, '(a)' ) 'You will NOT get output for this run!'
+         WRITE( 6, '(a)' ) '--dry-run is used for checking existence' //
+     &                     ' input files and config correctness.'
+         WRITE( 6, '(a)' ) 'REMOVE --dry-run FROM COMMAND ONCE CHECK' //
+     &                     ' IS PASSED TO RUN GEOS-CHEM PROPER!'
+         WRITE( 6, '(a)' ) REPEAT( '!', 44 )
+         WRITE( 6, '(a)' ) 
+      ENDIF
 
       !-----------------------------------------------------------------
       ! %%%% REPLICATING GCHP FUNCTIONALITY IN EXISTING GEOS-CHEM %%%%
@@ -723,17 +764,19 @@
 
       ! Compute the Olson landmap fields of State_Met
       ! (e.g. State_Met%IREG, State_Met%ILAND, etc.)
-      CALL Compute_Olson_Landmap( am_I_Root, Input_Opt, State_Grid,
-     &                            State_Met, RC )
+      CALL Compute_Olson_Landmap( am_I_Root,  Input_Opt,
+     &                            State_Grid, State_Met, RC )
       IF ( RC /= GC_SUCCESS ) THEN
          ErrMsg = 'Error encountered in "Compute_Olson_Landmap"!'
          CALL Error_Stop( ErrMsg, ThisLoc )
       ENDIF
 
-      ! Initialize PBL quantities but do not do mixing
-      ! Add option for non-local PBL (Lin, 03/31/09)
-      CALL Init_Mixing( am_I_Root,  Input_Opt,  State_Chm,
-     &                  State_Diag, State_Grid, State_Met, RC     )
+      IF ( .not. Input_Opt%DryRun ) THEN
+         ! Initialize PBL quantities but do not do mixing
+         ! Add option for non-local PBL (Lin, 03/31/09)
+         CALL Init_Mixing( am_I_Root,  Input_Opt,  State_Chm,
+     &                     State_Diag, State_Grid, State_Met, RC )
+      ENDIF
 
       ! Trap potential errors
       IF ( RC /= GC_SUCCESS ) THEN
@@ -795,7 +838,7 @@
       ! for Rn-Pb-Be and full-chem benchmark simulations
       ! NOTE: Species concentrations enter the subroutine in [kg/kg dry air]
       ! and are converted locally to [kg] (mps, 11/29/16)
-      IF ( LSTDRUN ) THEN 
+      IF ( LSTDRUN .and. (.not. Input_Opt%DryRun) ) THEN
          CALL StdRun( am_I_Root,  Input_Opt, State_Chm, State_Diag,
      &                State_Grid, State_Met, RC,        LBEGIN=.TRUE. )
 
@@ -986,9 +1029,11 @@
          CALL GEOS_Timer_Start( "=> HEMCO diagnostics",  RC )
          CALL GEOS_Timer_Start( "Output",                RC )
 #endif
-
-         ! Write HEMCO diagnostics (ckeller, 4/1/15)
-         CALL HCOI_GC_WriteDiagn( am_I_Root, Input_Opt, .FALSE., RC )
+         ! Do not do actual output for dry-run
+         IF ( .not. Input_Opt%DryRun) THEN
+            ! Write HEMCO diagnostics (ckeller, 4/1/15)
+            CALL HCOI_GC_WriteDiagn( am_I_Root, Input_Opt, .FALSE., RC )
+         ENDIF
 
          ! Trap potential errors
          IF ( RC /= GC_SUCCESS ) THEN
@@ -1007,7 +1052,8 @@
          !==============================================================
          !      ***** O B S P A C K   D I A G N O S T I C S *****
          !==============================================================
-         IF ( Input_Opt%Do_ObsPack .and. ( ELAPSED_TODAY == 0 ) ) THEN
+         IF ( Input_Opt%Do_ObsPack .and. ( ELAPSED_TODAY == 0 )
+     &        .and. (.not. Input_Opt%DryRun) ) THEN
 
 #if defined( USE_TIMERS )
             CALL GEOS_Timer_Start( "All diagnostics",        RC )
@@ -1030,7 +1076,8 @@
          ENDIF
 
          ! Original diagnostics
-         IF ( ITS_TIME_FOR_BPCH( Input_Opt ) ) THEN
+         IF ( ITS_TIME_FOR_BPCH( Input_Opt ) 
+     &        .and. (.not. Input_Opt%DryRun) ) THEN
 
 #if defined( BPCH_DIAG )
             !===========================================================
@@ -1072,13 +1119,17 @@
             CALL GEOS_Timer_Start( "HEMCO",  RC )
 #endif
 
-            ! Force the output of a HEMCO restart file (ckeller, 4/1/15)
-            CALL HCOI_GC_WriteDiagn( am_I_Root, Input_Opt, .TRUE., RC )
+            IF ( .not. Input_Opt%DryRun ) THEN
+               ! Force the output of a HEMCO restart file (ckeller, 4/1/15)
+               CALL HCOI_GC_WriteDiagn( am_I_Root, Input_Opt, .TRUE., 
+     &                                  RC )
+            
 
-            ! Trap potential errors
-            IF ( RC /= GC_SUCCESS ) THEN
-               ErrMsg = 'Error encountered in "HCOI_GC_WriteDiagn"!'
-               CALL Error_Stop( ErrMsg, ThisLoc )
+               ! Trap potential errors
+               IF ( RC /= GC_SUCCESS ) THEN
+                  ErrMsg = 'Error encountered in "HCOI_GC_WriteDiagn"!'
+                  CALL Error_Stop( ErrMsg, ThisLoc )
+               ENDIF
             ENDIF
 
 #if defined( USE_TIMERS )
@@ -1138,7 +1189,7 @@
          !==============================================================
          !        ***** L E A F   A R E A   I N D I C E S *****
          !==============================================================
-         IF ( ITS_A_NEW_DAY() ) THEN
+         IF ( ITS_A_NEW_DAY() .and. (.not. Input_Opt%DryRun) ) THEN
 
             ! Initialize the State_Met%XLAI_NATIVE field from HEMCO
             CALL Get_XlaiNative_from_HEMCO( am_I_Root, Input_Opt,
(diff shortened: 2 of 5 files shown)


jimmielin:
[dry_run_correct.log](https://github.com/geoschem/geos-chem/files/3801247/dry_run_correct.log)
[dry_run_missing.log](https://github.com/geoschem/geos-chem/files/3801248/dry_run_missing.log)

jimmielin:
Note that since `--dry-run` technically runs the whole of the GC "framework" (all initializations are still performed and some time step processes remain to support `EMISSIONS_RUN`) the memory footprint and performance of this is still non-negligible.

As a result of running init routines all the species info etc. are still printed. I think it would be useful to make `--dry-run` not only a check for files but also add printing information about the inventories etc. (@christophkeller) so one could parse a `--dry-run` output again **after** running a simulation, to get all information that is needed to replicate the configuration of this GC run, for paper publishing, documentation, etc.

kilicomu:
Hi @jimmielin - happy to talk elsewhere about this - do you mind briefly describing the contexts in which you see the `--dry-run` feature being used?

jimmielin:
> Hi @jimmielin - happy to talk elsewhere about this - do you mind briefly describing the contexts in which you see the `--dry-run` feature being used?

Hi Killian, apologies for not including more context in the PR. The `--dry-run` feature was originally envisioned to solve a common problem and source of frustration with GEOS-Chem users: downloading input data required for a run. This is especially necessary as users set up their cloud environments to run GEOS-Chem (see https://github.com/geoschem/geos-chem-cloud/issues/25) and need to fetch GEOS-Chem inputs to complete their run, and there are currently no good scripts to look at your configuration, get the list of files and automatically download them.

While previously you could download all HEMCO folders, for most uses one does not need to download all the available HEMCO inventories. This is especially true now that HEMCO is used to read in met fields, as (1) one only wants the met archives used in the run and (2) GEOS-Chem stops mid-run if a met field is not found in the middle of time stepping. This results in waste of simulation time.

The dry run option is exactly a tool to solve this. It runs through GEOS-Chem using the GEOS-Chem itself, the ultimate authority in determining which inputs are needed, and it will step through and output to standard output (for now) all files read (and if they are missing), incl. emission inventories and met fields. A script can then be used to parse this output and feed it to download data required for the run.

We've also previously proposed that this could be used for successful runs, too. One might want to attach a list of GEOS-Chem configurations and HEMCO emissions used in the run in a paper, for example, and `--dry-run` can output that exactly. Sharing raw `HEMCO_Config.rc` files might not be straightforward, as some inventories may replace each other and only HEMCO can determine this at run-time.

I hope this clarifies the context in which this feature was developed, please feel free to provide suggestions or ask further questions 😃

jimmielin:
@JiaweiZhuang has been testing the `--dry-run` feature in a docker container and he's seeing GEOS-Chem still rewinding to 1715 for the met fields. It seems like that `HEMCO_Config.rc` does not specify met fields using the `E` (exact) flag:

```
* ALBEDO    $METDIR/$YYYY/$MM/$MET.$YYYY$MM$DD.A1.$RES.$NC     ALBEDO   1980-2019/1-12/1-31/0-23/+30minute RFY xy  1  * -  1 1
```

Only using `RFY` here.
This also happens for some fields depending on met field year (e.g. `OFFLINE_LIGHTNING`) and `--dry-run` will insist on rewinding to 1715. I'm looking at how to fix this as we cannot know "ahead-of-time" what year is available. Perhaps the post-processing script will need to recognize this and simply "accept the newest year available" when it sees 1715.

msulprizio:
> @JiaweiZhuang has been testing the `--dry-run` feature in a docker container and he's seeing GEOS-Chem still rewinding to 1715 for the met fields. It seems like that `HEMCO_Config.rc` does not specify met fields using the `E` (exact) flag:
> 
> ```
> * ALBEDO    $METDIR/$YYYY/$MM/$MET.$YYYY$MM$DD.A1.$RES.$NC     ALBEDO   1980-2019/1-12/1-31/0-23/+30minute RFY xy  1  * -  1 1
> ```
> 
> Only using `RFY` here.
> This also happens for some fields depending on met field year (e.g. `OFFLINE_LIGHTNING`) and `--dry-run` will insist on rewinding to 1715. I'm looking at how to fix this as we cannot know "ahead-of-time" what year is available. Perhaps the post-processing script will need to recognize this and simply "accept the newest year available" when it sees 1715.

This issue is fixed in commit 3c26785fd1225ecf8634acbf225b0e513a0fcde2 in the dev/12.7.0 branch and will be included in 12.7.0.

JiaweiZhuang:
@jimmielin This is fabulous! 

I've packaged the dryrun-enabled executable as a Docker image, along with necessary input data, so that it can be used as a black box: https://github.com/JiaweiZhuang/gc-dryrun-docker

It generates this log file: [dryrun_12h.log](https://github.com/geoschem/geos-chem/files/3805269/dryrun_12h.log)

The dry-run finished in 2 minutes for this 12-hour 4x5 configuration. This can be further sped-up by increasing the time step from 10/20min to 30/60 min: [dryrun_12h_longstep.log](https://github.com/geoschem/geos-chem/files/3805271/dryrun_12h_longstep.log)

One issue with this image is that the required `CHEM_INPUTS` and `HEMCO` files add up to 3 GB, making the total image size 4 GB. 

In `CHEM_INPUTS`:
```
du -sh ./*
1.8M    ./FAST_JX
184K    ./FastJ_201204
448K    ./Linoz_200910
1.3G    ./MODIS_LAI_201204
1.2G    ./MODIS_LAI_201707
294M    ./Olson_Land_Map_201203
141M    ./TOMAS_201402
156K    ./UCX_201904
```

In `HEMCO`:
```
du -sh ./*
168K    ./LIGHTNOX
167M    ./PARANOX
4.0K    ./VOLCANO
```

Hopefully those data can be reduced, especially `MODIS_LAI`.

JiaweiZhuang:
For the parser script https://github.com/geoschem/geos-chem-cloud/issues/25#issuecomment-541446027, just change `line.startswith('HEMCO: Opening')` to `HEMCO: REQUIRED file NOT FOUND` and it will work exactly the same way.

yantosca:
GC should be reading the MODIS data from the MODIS_XLAI folder.

HEMCO: REQUIRED file NOT FOUND /ExtData/HEMCO/MODIS_XLAI/v2017-07/1710/MODIS_XLAI.025x025.171007.nc

(NOTE: In 12.7.0, this will be updated to use HEMCO/Yuan_XLAI instead).

So I am not sure if we still need the CHEM_INPUTS/MODIS_LAI_201204 or CHEM_INPUTS/MODIS_LAI_201707 data anymore.

JiaweiZhuang:
> So I am not sure if we still need the CHEM_INPUTS/MODIS_LAI_201204 or CHEM_INPUTS/MODIS_LAI_201707 data anymore.

I removed these two directories and the code still runs (https://github.com/JiaweiZhuang/gc-dryrun-docker/commit/df8b6b6927d8fc14cdafadd34aa72afdced0452c)

Now `ExtData/` is only 600 MB and the image size is 1.5 GB. Much better.

JiaweiZhuang:
@yantosca Would you/GCST be able to formalize this dry-run feature in future versions? Three major steps:

- [ ] Incorporate this PR into master branch, along with fixes like geoschem/HEMCO#3 and https://github.com/geoschem/geos-chem/commit/3c26785fd1225ecf8634acbf225b0e513a0fcde2
- [ ] Connect it to the parser and data download script geoschem/geos-chem-cloud#25
- [ ] Ideally, also package the utility as a container, ref https://github.com/JiaweiZhuang/gc-dryrun-docker

yantosca:
@JiaweiZhuang We can maybe try to get this into 12.7.0.

jimmielin:
Thanks @JiaweiZhuang @yantosca @msulprizio for the feedback. I noticed that I PRd to merge this onto `master` instead of `dev/12.7.0`, which is probably the correct choice. Hope that it can be fixed easily.

Hopefully we can also fix the non-HEMCO offenders (some of `CHEM_INPUTS`, and `PARANOX`, `LIGHTNOX`, `VOLCANO`) to either recognize the dry-run option or read them in a more sensible way. Some of these are custom-format ASCII files, though. We will probably just have to wrap them under dry-run as special cases, or people missing `CHEM_INPUTS` shouldn't be able to run GEOS-Chem in any way (it's like running GEOS-Chem without `input.geos` 😂)

One thing we also noticed is that GEOS-Chem without all the compute still moves through the time-stepping very slowly. I've taken care to comment out most of what's in the heartbeats except time-stepping computes (i.e. `IS_IT_NEW_DAY` etc) but it is still visibly slow. Maybe there is a possible performance optimization we can squeeze out of this. I can look into it as a side project later on.

yantosca:
Thanks Haipeng.  I can pull this into a branch off of dev/12.7.0 with the command line github instructions.  Should not be a problem.

Also that's a very interesting observation about the time.  GC uses an Astronomical Julian Date algorithm to compute time going forward (as that is the only algorithm I know that is really robust across end of years, end of centuries, and end of millenia).  Also we have to do function calls for e.g. GET_YEAR, GET_MONTH, GET_DAY, ITS_A_NEW_DAY so maybe this is where the bottleneck is.  We can look into this later.  That might be a good case for optimization.

yantosca:
Here is a profile with TAU from a 1-day run.   I am going to do a deeper dig into this when I have time.  
[dryrun.profile.log](https://github.com/geoschem/geos-chem/files/3810780/dryrun.profile.log)

The top routines by inclusive time are:
1 22.4% in hco_calcemis (hco_calc_mod)
2 14.3% in diagn_updatedriver (in hco_diagn_mod.F90), spent on constant propagation
3. 13.7% in get_current_emissions (hco_calc_mod)

Am wondering if disabling the HEMCO diagnostics with dryrun would speed things up significantly.  It also ooks like it may still be trying to do some computations deep down inside.

yantosca:
FYI, a quick look at the HEMCO log using WARNINGS: 3 VERBOSE:3 shows the following output:

```
  HEMCO vertical sigma-hybrid coordinates:
 Ap [Pa] (first and last):    0.0000000000000000        1.0000000000000000
 Bp [unitless] (first and last):    1.0000000000000000        0.0000000000000000
HEMCO: Entering HCO_INIT (hco_driver_mod.F90) ( 1)
HEMCO: Entering tIDx_Init (hco_tidx_mod.F90) ( 2)
HEMCO: Leaving tIDx_Init (hco_tidx_mod.F90) ( 2)
HEMCO: Entering HcoClock_Init (HCO_CLOCK_MOD.F90) ( 2)
HEMCO: Leaving HcoClock_Init (HCO_CLOCK_MOD.F90) ( 2)
Successfully added diagnostic EmisACET_Total to collection    1
Successfully added diagnostic EmisACET_Anthro to collection    1
```
etc, so it seems to be also initializing the diagnostic containers, which we may not need for a dry-run.  Or I'm not sure if this output is just generated but no diagnostics are updated.

Also in the HEMCO.log, it looks like HEMCO may still be trying to do some computations like applying scale factors.  This might be taking up some excess CPU cycles.  If we can shunt those then that might result in speedup.

```
HEMCO: Entering GET_CURRENT_EMISSIONS ( 3)
                                                              
 Evaluate field NEI11_EGU_HONO
Applying scale factor EDGAR_TODNOX
 Scale field EDGAR_TODNOX
 Time slice:            1
 IX, IY:           25          25
 Scale factor (IX,IY,L1):    0.00000000
 Mathematical operation :            1
Applying scale factor NEI99_DOW_NOX
 Scale field NEI99_DOW_NOX
 Time slice:            1
 IX, IY:           25          25
 Scale factor (IX,IY,L1):    0.00000000
 Mathematical operation :            1
Applying scale factor NEI11_NO_YRSCALE
 Scale field NEI11_NO_YRSCALE
 Time slice:            1
 IX, IY:           25          25
 Scale factor (IX,IY,L1):   0.887000024
 Mathematical operation :            1
Applying scale factor CONUS_MASK
 Mask field CONUS_MASK found and added to temporary mask.
HEMCO: Leaving GET_CURRENT_EMISSIONS ( 3)
```

yantosca:
Here is an output from TAU paraprof browser.
2019_11_07_16_08_29_Greenshot

Also note: 11% of the dry-run time is spent in routine getvertindex, which we probably shouldn't need to execute.

jimmielin:
> Here is an output from TAU paraprof browser.
> 2019_11_07_16_08_29_Greenshot
> 
> Also note: 11% of the dry-run time is spent in routine getvertindex, which we probably shouldn't need to execute.

Thanks Bob! This really helps. It is indeed safe to patch out the `hco_calcemis` calls. I will do that in the coming days and add to this PR.

yantosca:
Hi Haipeng,

I can block out the call myself and do another timing test in the feature/dryrun branch.  No worries.

yantosca:
I have pushed commit https://github.com/geoschem/geos-chem/commit/7c752e372f523a64543ade8c19d2f10641f80b19, which now skips extra HEMCO operations when running in dry-run mode.

A 1-month dry-run (20160701 to 20160801) now only takes 2 min 24 sec!  Much better than before.  I think we are good to merge this commit into 12.7.0.

yantosca:
In my previous comment I cited that a 1-month dry-run took 2 min 24 sec.  This was with VERBOSE: 3 and WARNINGS: 3 in the HEMCO_Config.rc file.  Turning those settings back to VERBOSE: 0 and WARNINGS: 1 generates less output to HEMCO.log, and makes the 1-month dry-run complete in 134 seconds (2 min 14 sec).  A little faster.  We could probably experiment with suppressing extra printout to the log, as that will also save some time.

yantosca:
I have added further modfiications in https://github.com/geoschem/geos-chem/commit/6db81f4a1d3bca073381f70efb0363b29022a9cd.  We skip operations such as excess printouts and array allocation when running GEOS-Chem in dry-run mode. 

Here is the output of a 1-month GEOS-Chem dry-run.  It only took 13 seconds to produce.  Much better than before!

[GC_dryrun_1month.log](https://github.com/geoschem/geos-chem/files/3826219/GC_dryrun_1month.log)

I will do some difference tests and unit tests to make sure that the code that I added does not change the existing GEOS-Chem simulation when you do not use the dry-run.

yantosca:
I ran a difference test to validate the dryrun modifications:

dryrun

Ref = branch RefTest
Dev = dev/12.7.0

Because I rebased the dryrun updates, all of the dryrun updates were made between RefTest and dev/12.7.0.  This gave identical results.

```
###############################################################################
### VALIDATION OF GEOS-CHEM OUTPUT FILES
### Run ID: benchmark
##@
### IDENTICAL : HEMCO_restart.201607010100.nc.{Ref,Dev}
### IDENTICAL : GEOSChem.AdvFluxVert.20160701_0000z.nc4.{Ref,Dev}
### IDENTICAL : GEOSChem.AerosolMass.20160701_0000z.nc4.{Ref,Dev}
### IDENTICAL : GEOSChem.Aerosols.20160701_0000z.nc4.{Ref,Dev}
### IDENTICAL : GEOSChem.Budget.20160701_0000z.nc4.{Ref,Dev}
### IDENTICAL : GEOSChem.CloudConvFlux.20160701_0000z.nc4.{Ref,Dev}
### IDENTICAL : GEOSChem.ConcAfterChem.20160701_0000z.nc4.{Ref,Dev}
### IDENTICAL : GEOSChem.DryDep.20160701_0000z.nc4.{Ref,Dev}
### IDENTICAL : GEOSChem.JValues.20160701_0000z.nc4.{Ref,Dev}
### IDENTICAL : GEOSChem.LevelEdgeDiags.20160701_0000z.nc4.{Ref,Dev}
### IDENTICAL : GEOSChem.ProdLoss.20160701_0000z.nc4.{Ref,Dev}
### IDENTICAL : GEOSChem.SpeciesConc.20160701_0000z.nc4.{Ref,Dev}
### IDENTICAL : GEOSChem.StateChm.20160701_0000z.nc4.{Ref,Dev}
### IDENTICAL : GEOSChem.StateMet.20160701_0000z.nc4.{Ref,Dev}
### IDENTICAL : GEOSChem.Transport.20160701_0100z.nc4.{Ref,Dev}
### IDENTICAL : GEOSChem.WetLossConv.20160701_0000z.nc4.{Ref,Dev}
### IDENTICAL : GEOSChem.WetLossLS.20160701_0000z.nc4.{Ref,Dev}
### IDENTICAL : HEMCO_diagnostics.201607010000.nc.{Ref,Dev}
##%
###############################################################################
```

yantosca:
Timing update: A 1-year dryrun (20160101 to 20170101) now takes 152.3 seconds (3 minutes 32 sec), which is acceptable.  

We can close out this PR, it will go live in 12.7.0.

lizziel:
@jimmielin Quick question on the implementation of isDryRun logical in HEMCO. HcoState has a field for type Options which holds run-time options for HEMCO. The dryrun option would be a good fit to store there, so accessed via HcoState%Options%isDryRun rather than HcoState%isDryRun. Is there a reason you did not put it there? If not, I'd like to update the implementation to use Options instead. Incidentally, isESMF is also not in Options and i'd like to move it there as well.

jimmielin:
> @jimmielin Quick question on the implementation of isDryRun logical in HEMCO. HcoState has a field for type Options which holds run-time options for HEMCO. The dryrun option would be a good fit to store there, so accessed via HcoState%Options%isDryRun rather than HcoState%isDryRun. Is there a reason you did not put it there? If not, I'd like to update the implementation to use Options instead. Incidentally, isESMF is also not in Options and i'd like to move it there as well.

Hi @lizziel, thanks for pointing that out! It was added exactly because `%isESMF` was there so it seemed logical to put that option there. I was not aware of the `HcoState%Options` structure, which indeed is a better place to put things. It'd be great if you could move both of the logicals to `Options`, which make more sense. Thanks!

lizziel:
Will do! Thanks for the quick reply.

lizziel:
I tested out the dryrun option (non-cloud) with dev/12.7.0. I found that I really wanted to see a dryrun output log that listed the files read, and was surprised that one did not get generated. I strongly recommend adding this feature rather than requiring users to retrieve an external tool to parse the names from the existing logs. The alternative is including a dryrun tool in the run directory to generate the list, but if fortran can write it out to a special dryrun file that would be ideal.

jimmielin:
> I tested out the dryrun option (non-cloud) with dev/12.7.0. I found that I really wanted to see a dryrun output log that listed the files read, and was surprised that one did not get generated. I strongly recommend adding this feature rather than requiring users to retrieve an external tool to parse the names from the existing logs. The alternative is including a dryrun tool in the run directory to generate the list, but if fortran can write it out to a special dryrun file that would be ideal.

I think that to steer away from writing specialized routines in Fortran, I suggest including a python script that generates the list based on the `--dry-run` output. Dealing with strings in Fortran is a nightmare and including this tool both in the run directory for sanity checks and with GCPy would be more useful and easier to code.

yantosca:
I have made several updates to the original dry-run code:

- The dry-run now prints out all (non-standard) files that are read in by HEMCO extensions
- The dry-run now prints out all file paths to its own log file (and not the GC and HEMCO logs)
- You can now invoke the dry run in multiple ways:
```
   ./geos --dryrun 
   ./geos --dryrun
   ./geos -d dryrun.log
```
- As shown above, if you give a filename as a second argument, then it will send dry-run output to that file.  Otherwise it will use a default file name GEOSChem.DryRun.log

Still to do: Write some code to enable the HEMCO standalone main program to parse these arguments to invoke a dry-run.  That should be pretty simple.

This capability is going into 12.7.0.

yantosca:
Also, here is a sample dry-run of a 1-day simulation done on Cannon (so most files are present).

- [dryrun.log](https://github.com/geoschem/geos-chem/files/3852874/dryrun.log)

lizziel:
Thanks @yantosca. I think ideally the logfile would include a list of unique files read and show paths only. This would involve removing duplicates and removing any text surrounding the file path. This way users could see a complete concise list of files used and files not found regardless of run length.

yantosca:
I can try to tweak the format that gets written to the dryrun log file.   I think it'd be hard to remove duplicates by the code although that could be done in post-processing.

I think we need a label of maybe "FOUND" and "NOT FOUND" in front of the file path, so that the external parser (python or bash) will know which files to download and which are already present.
Thoughts?

lizziel:
@jimmielin suggested having a python script in the directory to generate the log rather than mess with the fortran. This would work well and would have the advantage of being easily customizable. For FOUND vs NOT FOUND, I think the lack of NOT FOUND could be used to qualify if found.

yantosca:
OK I'll add NOT FOUND to only the missing files.  Thanks!

yantosca:
Here is updated output of the dryrun log.  I think this has a much better format.
* [dryrun_updated.log](https://github.com/geoschem/geos-chem/files/3853046/dryrun_updated.log)

yantosca:
Now verified that this update does not break GCHP.  Pushing to 12.7.0 now.

lizziel:
Thanks @yantosca, the new format looks good. What are your thoughts on creating this file, without duplicates, using a python script in the run directory?

yantosca:
@lizziel So the GEOS-Chem code can now produce the dryrun.log (as shown in prior comments).  This dryrun.log now contains both found and missing files, although there can be duplicates.

@JiaweiZhuang had written a Python function to extract path names from a log file (see function extract_data_path in https://github.com/geoschem/geos-chem-cloud/issues/25,  That routine uses the sorted function to sort the files (and I think that also eliminates duplicates at the same time).

I modified Jiawei's function so that it ignores the comments.  (I pushed another commit https://github.com/geoschem/geos-chem/commit/b2530773b099c8508b14b4a40d08425a969c860f to make sure all comments begin with "!" so the parser can ignore them).  This should be able to read the output of the dryrun.log and return a list of paths:

```python
def get_data_paths(filename, prefix_filter=""):
    """
    Returns a list of pathnames from a GEOS-Chem log file.
    This can be used to get a list of files that should be
    downloaded from gcgrid or from Amazon S3.

    Args:
    -----
        filename : str
            GEOS-Chem standard log file

        prefix_filter : str
            Restricts the output to file paths starting with
            this prefix (e.g. "/home/ubuntu/ExtData/HEMCO/")
            Default value: ''

    Returns:
    --------
        data list : list of str
            List of full pathnames of data files found in
            the log file.

    Author:
    -------
        Jiawei Zhuang (jiaweizhuang@g.harvard.edu)
        Modified by Bob Yantosca (yantosca@seas.harvard.edu)
    """

    # Initialization
    prefix_len = len(prefix_filter)
    data_list = set()  # only keep unique files

    # Open file (or die with error)
    try:
        f = open(filename, "r")
    except FileNotFoundError:
        raise FileNotFoundError("Could not find file {}".format(filename))

    # Read data from the file line by line.
    # Add file paths to the data_list set.
    line = f.readline()
    while line:

        # Skip comment headers
        if line[0] is not '!':

            # Find the data path
            if len(line.split()) > 0:
                data_path = line.split()[-1]
            else:
                data_path = line

            # remove common prefix
            if data_path.startswith(prefix_filter):
                trimmed_path = data_path[prefix_len:]
            data_list.add(trimmed_path)

        # Read next line
        line = f.readline()

    # Close file and return
    f.close()
    data_list = sorted(list(data_list))
    return data_list
```

From there, it should be a simple matter to also create a bash script with the download commands from either S3 or Compute Canada.

yantosca:
Or if you only want to find the data paths that are missing, you can modify the code above to look for "NOT FOUND" and then just return those.

lizziel:
I tested this out and see that the information sent to the GEOS-Chem main log and HEMCO.log is now diverted to the new logfile and without the information of what is reading it (e.g. HEMCO versus somewhere else). I think there is value keeping the prefix information since it is informative if there is a problem, and I think it would be best to have that information in the main GEOS-Chem log so you know which timestep it is read as well as general context within the execution order.

The extra log I would like to see in GC Classic non-cloud runs is a summary log of unique file paths. I agree with @jimmielin that this would be a pain to do in fortran. The new implementation gets there half-way but I think the removal of redundancies is a critical feature that is missing. I'm therefore in favor of using a python script to generate the file post-dryrun from the main GEOS-Chem log. The file could be called input_list.dryrun, or something like that, so it is not confused with an actual run log.

I realize the python function to parse the file would be in gcpy but I think a non-gcpy-dependent script would be handy for general users. We could also keep it in a tools or test directory within GEOS-Chem source code since we are planning to move unit tests and other testing code to the repository anyway.

yantosca:
I have updated the GEOS-Chem standalone and have also added a HEMCO standalone.  The status of each file and its full path are printed out to both the GEOS-Chem log (for GC dry-runs) and the HEMCO standalone (for HEMCO dry-runs).

The GEOS-Chem dry-run can be invoked with one of the following commands:
```
./geos --dry-run > log
./geos --dryrun > log
./geos -d > log
```
and the GEOS-Chem log file that it creates looks like this: [gc_dryrun.log](https://github.com/geoschem/geos-chem/files/3864851/gc_dryrun.log)

The HEMCO standalone dry-run can be invoked with:
```
./hemco_standalone.x -c HEMCO_sa_Config.rc --dry-run
./hemco_standalone.x -c HEMCO_sa_Config.rc --dryrun
./hemco_standalone.x -c HEMCO_sa_Config.rc -d
```
and the log file that it creates looks like this: [hco_dryrun.log](https://github.com/geoschem/geos-chem/files/3864853/hco_dryrun.log)

We would need to parse these logs with a Python script into:
1. A log file of unique file names read by the simulation.
2. A download script from AWS S3
3. A download script from Compute Canada

I will go ahead to merge this up to 12.7.0 unless you have any objections (testing w/ GCHP first).

yantosca:
I have created a python script that will parse a GEOS-Chem or HEMCO log with dry-run output into a log of unique files, and a script to download data from AWS S3.  (Still need to work on the download from Compute Canada).  The script only uses "os" and "sys" packages, and can be distributed independently of GCPy (e.g. as in a run directory).  It could also be included as a GCPy example script.

The input file is the [gc_dryrun.log](https://github.com/geoschem/geos-chem/files/3864851/gc_dryrun.log) file from my previous comment.

* Python script:  [parse_dryrun_output.py.txt](https://github.com/geoschem/geos-chem/files/3865638/parse_dryrun_output.py.txt)
* List of unique files needed for the GEOS_Chem simulation: [log.unique.txt](https://github.com/geoschem/geos-chem/files/3865641/log.unique.txt)
* Script to download from AWS S3: [downlaod_from_aws_s3.sh.txt](https://github.com/geoschem/geos-chem/files/3865647/downlaod_from_aws_s3.sh.txt)

NOTE: I renamed the python script to *.py.txt to link it on Github.  You can just delete the .txt if you download it.  Ditto for download_from_aws_s3.sh.txt

I think we are now 2/3 of the way there!!!

yantosca:
I will close this out as this is now in dev/12.7.0.  Some minor work still needs to be done on the download scripts but that can be wrapped up very soon.

PR #124, opened 2019-11-05
[PULL REQUEST] Bug fix for RH in heterogeneous chemistry

cdholmes:
RH is in percent but assumed to be fraction 0-1 in some places.
Affects gamma values for NO3 and N2O5, since version 12.6.
Corrected gamma values are lower for sea salt.

Signed-off-by: Christopher Holmes <cdholmes@fsu.edu>

### DIFF
--- a/KPP/Standard/gckpp_HetRates.F90
+++ b/KPP/Standard/gckpp_HetRates.F90
@@ -244,6 +244,7 @@ MODULE GCKPP_HETRATES
 !  17 Oct 2018 - C.D. Holmes - Added cloud heterogeneous chemistry (CloudHet);
 !                              Gamma updates for NOx species
 !  13 Dec 2018 - E. McDuffie - Report gammaN2O5 as a State Chem parameter
+!  04 Nov 2019 - C.D. Holmes - Bug fixes for gammaN2O5 (RH dependence)
 !EOP
 !------------------------------------------------------------------------------
 !BOC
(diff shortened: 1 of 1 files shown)


yantosca:
This PR has been merged into dev/12.7.0.  It will become live in GC 12.7.0, and we will run an internal benchmark to assess the effect of this update.

cdholmes:
I suggest that this bug fix should also be backported to 12.6. Currently, 12.6.2 still has this unfixed bug and heterogeneous chemistry was one of the headline features of 12.6. I expect that the bugfix will raise O3 over the oceans by a couple ppb.

yc-chan:
I wonder if we also have to change the description for RH in function N2O5_InorgOrg. 

I think the RH variable in the "line "eps = 1.5e-1_fp * OCratio + 1.6e-3_fp * RH" has the right unit (%), but the input parameter section still states it is in unitless fraction.

yantosca:
It is not in our procedure to backport bug fixes to master once a version has been benchmarked and approved.  For this reason we issue patch releases (e.g. 12.6.1, 12.6,2, etc.).

cdholmes:
Bob,
That’s what I meant. I think this fix should be released as 12.6.3.
Chris

Christopher D. Holmes, PhD
Earth, Ocean and Atmospheric Science
Florida State University
http://acgc.eoas.fsu.edu
________________________________

yantosca:
I updated the comment in routine N2O5_InorgOrg in this commit: https://github.com/geoschem/geos-chem/commit/0cdf070cb9de0d5b5d46cd801b777917d00290fa, which has been pushed to dev/12.7.0.

We are including this update in 12.7.0 instead of 12.6.2 or 12.6.3, because this update will change the benchmark results, and therefore should go into a "Y" (feature) version instead of a "Z" (patch) version.  We will run an internal benchmark to assess the change that this fix brings in.

PR #150, opened 2019-11-22
Use ENTRYPOINT in release.dockerfile

LiamBindle:
Yesterday I learned that a better way to get `/opt/geos-chem/bin` into `$PATH` is to use `ENTRYPOINT` rather than a source line in `/etc/bash.bashrc`. This PR includes that change to `.ci-pipelines/release.dockerfile`.


yantosca:
Thanks for this PR.  I've merged it into dev/12.7.0 and it will ship with 12.7.0.

PR #157, opened 2019-11-26
HEMCO standalone gridfix

cdholmes:
We found that HEMCO standalone was not able to run at the native resolution of GMAO 0.5x0.625 grids (MERRA-2, GEOS-FP). Using these grids in HEMCO standalone requires specifying the grid edges and centers in the HEMCO_sa_Grid.rc file in order to properly treat the half-size polar boxes. However, specifying all of the grid points at this native resolution exceeded the maximum line length that HEMCO standalone could read.

This fix increases the maximum line length and raises an error if the line exceeds this max length.

### DIFF
--- a/HEMCO/Core/hco_chartools_mod.F90
+++ b/HEMCO/Core/hco_chartools_mod.F90
@@ -871,7 +871,7 @@ CONTAINS
 ! LOCAL VARIABLES:
 !
     INTEGER             :: IOS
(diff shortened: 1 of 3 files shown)


yantosca:
These updates were added in manually so as to avoid conflicting with newer updates to HEMCO.  See: https://github.com/geoschem/geos-chem/issues/161.

PR #184, opened 2019-12-29
[PULL REQUEST] Fix WRF-GC SetGridFromCtrEdges interface

jimmielin:
Hi GCST,

In the migration to FlexGrid, the SetGridFromCtrEdges function in
gc_grid_mod.F90 for `MODEL_WRF` was not completely updated.

This update fixes this interface for GEOS-Chem > 12.4.0 to operate with
WRF-GC. It does not affect other simulations since it is wrapped in
the `MODEL_WRF` pre-processor switch.

Happy Holidays!
Haipeng


msulprizio:
This is now merged into dev/12.7.1

PR #198, opened 2020-01-18
[PULL REQUEST] Remove requirement that data comes from "GIGCchem"

sdeastham:
DiagList only permits data to come from the "GIGCchem" component, but we can want data from other components (e.g. mass fluxes from the GIGCenv component). This requirement should be removed.

### DIFF
--- a/Headers/diaglist_mod.F90
+++ b/Headers/diaglist_mod.F90
@@ -525,19 +525,22 @@ CONTAINS
              RETURN
           ENDIF
 
-          ! Throw an error if we cannot find the gridcomp name
-          ! (e.g. "'GIGCchem',").  GCHP will choke if this isn't found.
-          G = INDEX( Line, "'GIGCchem'," )
-          IF ( G == 0 ) THEN
-             ErrMsg = 'The name of the GCHP gridded component '           // &
-                      "(e.g. 'GIGCchem') for attribute "  // '" '         // &
-                      TRIM( AttName ) // '" must be enclosed in '         // &
-                      'single quotes and be followed by a comma. '        // &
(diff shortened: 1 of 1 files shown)


lizziel:
Agreed. @yantosca, do you remember why this check is here?

yantosca:
I don't remember.  Maybe we can go ahead and take that out then.

lizziel:
Yeah, we already have this code that skips the variable if not in GIGCchem so that it is isn't added to diaglist. Exiting with an error would potentially be a problem since there can be other diagnostics we want to get out of history that are outside the chem child grid comp.

```
       ! Skip line if GIGCchem not present
       ! GEOS-Chem is names 'GEOSCHEMCHEM' on NCCS discover,
       ! scan accordingly (ckeller, 12/29/17)
#if defined( MODEL_GEOS )
       IF ( INDEX( Line, 'GEOSCHEMCHEM' ) .le. 0 ) CYCLE
#else
       IF ( INDEX( Line, 'GIGCchem' ) .le. 0 ) CYCLE
#endif
```

yantosca:
Hmm.  Now I recall.  I think we put that there because "GIGCchem" indicates that the line contains a diagnostic field.  Maybe we need a better search algorithm for that then.

lizziel:
Once we get start using CPP flags for Classic vs GCHP I think some of the HISTORY.rc parsing error handling can be switched to Classic only. Some of it may be more strict than MAPL rules for HISTORY.rc syntax.

lizziel:
Unless there are objections I am going to merge this PR into 12.7.1 and then delete the commented out code.

lizziel:
This is now merged in 12.7.1.

PR #204, opened 2020-01-23
ObsPack Bug Fix, Update for Instantaneous Sampling

chfite:
There are multiple proposed changes/fixes to obspack_mod.F90. 

1) There was a bug in the Seconds_Since_1970 function where the time conversion factor for fractional minutes of the day was 3600s instead of what should actually be 1440s (applied here). This means that ObsPack has been using the incorrect time steps for computing averages for users up until this point. 

2) In the ObsPack wiki http://wiki.seas.harvard.edu/geos-chem/index.php/ObsPack_diagnostic there was mention of defining a CT_sampling_strategy of "4", meaning instantaneous sampling. However, the current obspack_mod.F90 does not include an option for instantaneous sampling. Here I provide the changes that allow for ObsPack to do instantaneous sampling.
      - In the ObsPack_Read_Input function a CASE 4 is added for instantaneous sampling
      - In the ObsPack_Sample function cases are used for the sampling strategy, in which, CASE 4 finds the closest model time step to 
        the observational time step samples at that time. 

This has been tested and does now sample instantaneously at the closest model time step, REGARDLESS of what the dynamic time steps are for a given simulation.

### DIFF
--- a/ObsPack/obspack_mod.F90
+++ b/ObsPack/obspack_mod.F90
@@ -518,6 +518,7 @@ CONTAINS
                                central_time(5,N),                            & 
                                central_time(6,N)  )
 
+
        ! Pick the start and end time of the averaging interval
        ! depending on the averaging strategy listed in the file
        SELECT CASE ( State_Diag%ObsPack_Strategy(N) )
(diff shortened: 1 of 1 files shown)


yantosca:
Thanks Charley.  We will probably be able to get this into the code soon (after 12.7.0 benchmarking clears).

yantosca:
This update has now passed unit tests.  It has now been added to dev/12.7.1, and will ship with the 12.7.1 version (soon after 12.7.0 passes benchmarking).

PR #225, opened 2020-02-17
[PULL REQUEST] Remove unused do_regrid_a2a and read_input_grid routines in MAP_A2A.

jimmielin:
## Overview
`Regrid_A2A_Mod` seems to have to (unused?) routines, `DO_REGRID_A2A` and consequently `READ_INPUT_GRID`, which taps directly into `m_netcdf` utilities.

It does not make sense to have IO routines in a regridding module, and they are used.

## Commit message
There are two unused I/O routines in the MAP_A2A regridding module that should not be here as this is an regridding module.
This commit stages their removal from the module.

I've set the target for this to `dev/12.8.0` but please feel free to move. Thanks!

### DIFF
--- a/GeosUtil/regrid_a2a_mod.F90
+++ b/GeosUtil/regrid_a2a_mod.F90
@@ -22,7 +22,6 @@ MODULE Regrid_A2A_Mod
 !
 ! !PUBLIC MEMBER FUNCTIONS:
 !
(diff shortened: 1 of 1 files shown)


msulprizio:
This PR has now been merged into dev/12.7.2

PR #257, opened 2020-03-22
Fix Init_Pressure error for NZ /= 47, 72 in MODEL_WRF

jimmielin:
Do not error out in PRESSURE_MOD:INIT_PRESSURE if models are not 47 or 72 in `MODEL_WRF`.

WRF-GC runs on its own set of vertical coordinates which are intentionally not 47 or 72 to avoid a HEMCO hardcoded vertical regridding workaround.
This patch removes an error message thrown in init_pressure notifying that the vertical levels are not hard-coded.

Thanks!

### DIFF
--- a/GeosUtil/pressure_mod.F90
+++ b/GeosUtil/pressure_mod.F90
@@ -794,11 +794,13 @@ CONTAINS
 
     ELSE
 
+#if !defined( MODEL_WRF )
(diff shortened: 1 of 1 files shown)


msulprizio:
This is now merged into branch `dev/12.8.1`.

Issue #273, opened 2020-04-08
[BUG/ISSUE] Regrid_A2A_Mod is potentially broken for nested-grid simulations


jimmielin:
## Background
`Regrid_A2A_Mod` is used in GEOS-Chem Nested-Grid, WRF-GC and (temporarily) HEMCO_CESM to perform regridding of emission inventory data to the model grid.

In both WRF-GC and HEMCO_CESM I've been seeing horizontal regridding artifacts at CPU boundaries like so:
![image](https://user-images.githubusercontent.com/123603/78738197-419d5b00-791f-11ea-957d-950d55070a3b.png)

This led me to investigate `Regrid_A2A_Mod`.

## Describe the bug
The issue is that `Regrid_A2A_Mod` when used within HEMCO's `hco_interp_mod.F90` calls `MAP_A2A` with `ig = 0, iv = 0` like so:
```fortran
          ! Do the regridding
          CALL MAP_A2A( NX,      NY, LonEdgeI,    LatEdgeI, ORIG_2D,  &
                        HcoState%NX, HcoState%NY, LonEdgeO, LatEdgeO, &
                        REGR_2D, 0, 0, HCO_MISSVAL )
```

This triggers a particular treatment in `Regrid_A2A_Mod.F90`'s `YMAP_R*R*` routine:

```fortran
     !===================================================================
     ! Final processing for poles
     !===================================================================
     if ( ig .eq. 0 .and. iv .eq. 0 ) then

        ! South pole
        sum = 0.e+0_fp
        nlon= 0.0d0
        do i=1,im
           if(abs(q2(i,1)-miss)>tiny_r8 ) then
              sum = sum + q2(i,1)
              nlon= nlon + 1.0d0
           endif
        enddo

        if ( nlon > 0.0d0 ) sum = sum / nlon
        do i=1,im
           q2(i,1) = sum
        enddo

        ! North pole:
        sum = 0.e+0_fp
        nlon= 0.0d0
        do i=1,im
           if( abs(q2(i,jn)-miss)>tiny_r8 ) then
              sum = sum + q2(i,jn)
              nlon= nlon + 1.0d0
           endif
        enddo

        if ( nlon > 0.0d0 ) sum = sum / DBLE( im )
        do i=1,im
           q2(i,jn) = sum
        enddo

     endif
```

This is too wordy. Let me summarize the code:
```fortran
! Final processing for poles
if(ig .eq. 0 .and. iv .eq. 0) then
[...]

PR #274, opened 2020-04-08
Only perform pole-averaging of regridded data if actually regridding to the poles.

jimmielin:
Please refer to #273 for discussion. I've selected `dev/12.8.0` because that was the base of my commit. Please feel free to cherry-pick to whichever branch as appropriate. Thanks!

Haipeng

----------
Only perform pole-averaging of regridded data if actually regridding to the poles.

This commit fixes an issue affecting GEOS-Chem Nested, MODEL_WRF, MODEL_CESM (in HEMCO_CESM) in which Regrid_A2A performs northernmost and southernmost quantity averaging regardless of whether the boundaries are actually on the poles.

Now the check is actually performed, because REGRID_A2A is being increasingly used for regional purposes.

### DIFF
--- a/GeosUtil/regrid_a2a_mod.F90
+++ b/GeosUtil/regrid_a2a_mod.F90
@@ -783,34 +783,38 @@ CONTAINS
      if ( ig .eq. 0 .and. iv .eq. 0 ) then
 
         ! South pole
-        sum = 0.e+0_fp
-        nlon= 0.0d0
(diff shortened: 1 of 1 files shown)


msulprizio:
This is now merged into branch `dev/12.8.1` in commit https://github.com/geoschem/geos-chem/commit/45d28661fe20038da58cf7f9591d5e5db2359125.

PR #319, opened 2020-05-20
Enable KppStop disabling of fail-safe for MODEL_WRF

jimmielin:
Hi GCST,

This minor update extends `MODEL_GEOS`'s functionality for optionally ignoring integrate errors within the GEOS-Chem FlexChem integrator and logging it instead, to WRF-GC (`MODEL_WRF`)

This will be used as a handy debugging feature for WRF-GC's further developments. It does not affect other models.

Thank you!

Best,
Haipeng

~~**Edit:** This seems to cause some problems and registry crashes when printing the list, even if the DiagList is empty, for some reason. Please hold off this PR for now, sorry!~~
Fixed. What a peculiar bug! It was caused because I was using Windows line endings and directly uploaded to Cannon for compiling. It would initialize properly but throw a segmentation error in `diaglist_mod` when checking for `associated( current )`. I switched to Unix line endings to compile `state_diag_mod.F90` and everything was fine!

### DIFF
--- a/GeosCore/flexchem_mod.F90
+++ b/GeosCore/flexchem_mod.F90
@@ -487,7 +487,7 @@ CONTAINS
        CALL DEBUG_MSG( '### Do_FlexChem: after FAST_JX' )
     ENDIF
 
-#ifdef MODEL_GEOS
+#if defined( MODEL_GEOS ) || defined( MODEL_WRF )
     ! Init diagnostics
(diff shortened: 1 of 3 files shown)


yantosca:
Now added into 12.8.2.  Benchmark diff tests showed no changes w/r/t the 12.8.1 version.

PR #323, opened 2020-05-22
Do not call OMP external functions outside of MODEL_CLASSIC

jimmielin:
Hi GCST,

In GEOS-Chem 12.8.0 the new timers request OMP thread information, which is unavailable in non-OpenMP environments. This minor change encloses these calls within `MODEL_CLASSIC` pre-processor statements.

I've used `#ifdef MODEL_CLASSIC` here because I'm not aware if hybrid MPI/OMP is used in GCHP, but please otherwise feel free to set this as `#if !defined(MODEL_WRF)` instead. Thank you!

Best,
Haipeng

### DIFF
--- a/GeosCore/flexchem_mod.F90
+++ b/GeosCore/flexchem_mod.F90
@@ -180,7 +180,9 @@ CONTAINS
     INTEGER,  SAVE         :: CH4_YEAR  = -1
     REAL(fp), SAVE         :: C3090S,   C0030S,   C0030N,    C3090N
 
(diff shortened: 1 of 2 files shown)


yantosca:
This pull request has now been merged into dev/12.8.2, and will be included in 12.8.2.  NOTE: we have also added additional safeguards when MODEL_CLASSIC=y and OMP=n in commit: https://github.com/geoschem/geos-chem/commit/535cbfef6cbea786197e0d689e22317315aeed66.

PR #325, opened 2020-05-24
Move module variables into State_Met, State_Chm

jimmielin:
Hi GCST,

I've been meaning to finish this for a while but as we discussed two years ago, there are some global module variables that should be moved to `State_Chm` for the sake of avoiding module state and also compatibility with a "nested-grid" version of WRF-GC.

Please find attached three commits to move module variables to state modules from:
* `Set_Global_CH4_Mod`. `SFC_CH4` is now in `State_Chm`.
* `Wetscav_Mod`. Most are met quantities (`CLDICE`, etc.) which I moved to `State_Met` as I thought it would fit better there.
* `Toms_Mod`. Most are HEMCO pointers. Moved to `State_Chm`.

All the relevant interface updates, etc. are also included for GC-Classic and GCHP. The code was tested on GC-Classic using CMake.

Commits were based off `master` as of 12.8.1 release. Please let me know if I should base off the updates on a newer development branch in the future to make merging easier.

Thanks!

Best,
Haipeng


yantosca:
This update produces zero diffs in benchmark difference tests.  It has now been merged into dev/12.8.2 and will go into 12.8.2.

PR #326, opened 2020-05-25
Move MODEL_WRF return to top of INIT_DRYDEP.

jimmielin:
Hi GCST,

This change for `MODEL_WRF` fixes the issue that an allocation check and return was not early enough, resulting in a second function call zeroing out internal species indices and breaking dry deposition instead.

Nothing except 4 hours of weekend debugging was harmed during the production of this pull request.

Best,
Haipeng


yantosca:
This update was applied manually into 12.8.2 in order to avoid bringing in other modifications that will go into 12.9.0.

PR #334, opened 2020-05-28
[PULL REQUEST] Implement GEOS-Chem chemistry within CESM

fritzt:
Hi GCST,

I'm creating this PR to document one of the first stages of the CESM-GC implementation.

Here's an overview of the changes:
- Addition of a new preprocessor flag: `MODEL_CESM`.
- Hand off the I/O performed in `drydep_mod` and `fast_jx_mod` to CAM. 
- Addition of a new `State_Chm` variable to store dry deposition velocities. In CESM, dry deposition velocities can be passed from the land model. To fill in the missing data, GEOS-Chem needs to compute over ocean and ice.
- In `drydep_mod`, split up `DO_DRYDEP` into `CALL DEPVEL` and `CALL UPDATE_DRYDEPSAV`. The former computes dry deposition velocities, while the latter updates dry deposition frequencies. This last step is done externally in CESM in order to merge with the data from the land model.
- Rename a few GEOS-Chem variables to avoid namespace conflict with CAM (in `planeflight_mod`, `vdiff_mod`).
- Temporarily bypass a few UCX routines.
- Temporarily bypass LINOZ in CESM as the default vertical grid does not extend sufficiently high.
- Update the calls to HCO_GetPtr to `dev/GC_13.0.0`. The HEMCO subfolder is not compiled directly in CESM but is rather its own independent module, which is based off `dev/GC_13.0.0`. 

I initially applied my changes off of 12.6.3. This PR applies all commits since 12.6.0 to `CESM`.

Regards,
Thibaud Fritz


lizziel:
You can ignore the quick build failure. I don't expect this branch to build yet. The CESM branch needs to be added to the ignore list for continuous integration testing on Azure.

Issue #329, opened 2020-05-28
[BUG/ISSUE]UCX horizontal resolution issue with FlexGrid


eamarais:
## Describe the bug
I’m finding an issue with FlexGrid over a custom domain (West Africa) at the native horizontal resolution of the met fields when using standard chemistry mechanism with UCX in version 12.8.1. My run stops with the debug error message in UCX_INIT “'NetCDF NOx coeffs only preprocessed for 2x25 and 4x5 - try setting the module variable UCXNETCDF=.FALSE.”
 
My current fix is to just hardcode this to “FALSE” for 4x5 boundary condition and nested grid runs.

sdeastham:
Thanks for raising this! My apologies – I’ve principally been using GCHP recently so I knew in principle that there’s no reason for GC’s UCX code to be hardcoded for 4x5 or 2x2.5. But I hadn’t realized we still had this old legacy code. It’s an inconsistency which I agree we need to iron out. I think it would make most sense to:
 
1)	Remove the UCXNETCDF option, and trim ucx_mod.F90 so that only the UCXNETCDF=.FALSE. options are used throughout
2)	Unify the behavior in GEOS and GEOS-Chem. There seem to be a few places in ucx_mod.F90 which use more recent BCs for GEOS than for GEOS-Chem, which seems unnecessary.
 
The good news is that hardcoding UCXNETCDF=.FALSE. should have the exact intended effect.
 
@GCST: How soon could we:
 
a)	Change the default for UCXNETCDF to .FALSE.?
b)	Incorporate a fix (I’m happy to write this) which removes the old, redundant code and updates GC-Classic & GCHP to be consistent with GEOS?
 
My guess is that we could get a) in soon but b) will be a few versions later? Part b) will actually take some work, because I’m realizing that there’s a (science) issue with some of the assumptions about how the NOx data is indexed.

msulprizio:
This is now resolved in https://github.com/geoschem/geos-chem/commit/af4edc9406284abf51c931e9a544117c5320ae26 and will be included in GEOS-Chem 13.0.0

PR #344, opened 2020-06-04
Fix handling of NOx coefficients for meso-chem

sdeastham:
The GET_JJNOX routine in ucx_mod.F90 has a generic approach which can function for any
grid resolution, but is only used for nested grids. However, this approach is also needed for
any grid other than 2x2.5 and 4x5. As such, mesospheric chemistry currently uses incorrect
NOx coefficients in GCHP simulations, or any simulation other than 2x2.5 or 4x5. Using the
generic routine for all simulations resolves this problem.

### DIFF
--- a/GeosCore/ucx_mod.F90
+++ b/GeosCore/ucx_mod.F90
@@ -4231,8 +4231,7 @@ CONTAINS
 ! !IROUTINE: get_jjnox
 !
 ! !DESCRIPTION: Subroutine GET\_JJNOX maps grid box at location IISIM, JJSIM of
-! the simulation grid onto the latitude grid of the NOXCOEFF array. JJNOX can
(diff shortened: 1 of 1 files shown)


yantosca:
For reference: This PR should fix the issue brought up in https://github.com/geoschem/geos-chem/issues/329.

sdeastham:
@yantosca Correct! I think this is the only actual "science" problem related to #329. The other problems raised in that issue should be resolvable by stripping out the `UCXNETCDF=.TRUE.` options.

msulprizio:
This PR was closed when the dev/12.9.0 branch was deleted upon approval of GEOS-Chem 12.9.0. It has been merged into dev/13.0.0 in commit c157908c20ab52a353969e934eb11916bfcb22d4.

Issue #346, opened 2020-06-08
[BUG/ISSUE] RRTMG stratospheric aerosol shortwave forcing is non-physical


sdeastham:
## Describe the bug
When running GEOS-Chem with RRTMG enabled, the shortwave forcing attributed to stratospheric aerosol is positive (as calculated using NetCDF diagnostics, and quantifying stratospheric aerosol forcing as "baseline flux" - "strat aerosol flux" under all-sky conditions). This does not agree with observations, and is likely due to a bug in RRTMG's interface code.

## To Reproduce
Compile and run GEOS-Chem with RRTMG for 1 time step, storing out both `RadAllSkySWTOA_BA` and `RadAllSkySWTOA_ST` in `HISTORY.rc`. Take the `BA` value and subtract the `ST` value to get the forcing. The forcing is positive.

## Expected behavior
Shortwave forcing at TOA for stratospheric aerosol should be negative, as stratospheric aerosol tends to scatter incoming radiation.

## Error messages
N/A

## Required information
Please include the following:
 - GEOS-Chem version: 12.8
 - ifort 19
 - Harvard Odyssey/Cannon cluster
 - Unmodified code

## Additional context
A solution to this will be posted shortly as a pull request.

lizziel:
See also pull request https://github.com/geoschem/geos-chem/pull/347.

PR #348, opened 2020-06-08
Enable GCHP to be compiled with RRTMG

sdeastham:
Currently, GCHP cannot be compiled with RRTMG. This fixes that. With this set of commits, GCHP can be compiled by supplying `-DRRTMG` to the `cmake` build call, with no other compile modifications. The resulting executable can be run either WITH or WITHOUT RRTMG. This commit also adds the necessary files to ExtData.rc, an appropriate HISTORY.rc collection (disabled by default), and fixes a bug in RRTMG described in issue #346 (although this is also resolved independently by PR #347). Currently, this code will result in RRTMG being run on every chemistry timestep, which is very expensive (a timer has been added to check this behavior). Further work will be needed to have RRTMG called less frequently, if selected.


lizziel:
Thanks @sdeastham. I manually did a merge locally to resolve conflicts but am running into build problems. I don't think these are specific to your updates but pre-date them with the RRTMG simulation. Did you test within the [GEOS-Chem Classic wrapper](https://github.com/geoschem/geos-chem-classic) in addition to GCHPctm?

sdeastham:
I did not test with GC Classic - can you post the issues?

lizziel:
Yes, I will post them, but after I get the version prior to the merge building with RRTMG. I am certain there are issues unrelated to your updates and will deal with those first in case they end up resolving everything.

lizziel:
This PR is now merged. There were a couple build issues. The first issue was pre-existing before this PR. The -override-limits flag worked for GCC but not Intel compilers. This PR fixed it for Intel, but inadvertently caused it to no longer build with GCC. The second was introduced with this PR. The dependency on GEOSChemBuildProperties was removed for GCHP but is required for GC-Classic. I fixed both of these issues directly following the merge.

PR #386, opened 2020-07-28
Update KPP functions to have unique names in KPP namespace

tsherwen:
Updated a some of the new KPP functions to have unique names in *.eqn/*.kpp files.

Some of the new functions have the same name as species used elsewhere in KPP namespace (e.g. PAN, NIT). This causes issues for various parsers I have been using with GEOS-Chem's KPP files since version 10/11, and suspect it could affect other people using these parsers or other automated methods for working with KPP files. For KPP functions in the past, we have typical prefixed the functions (E.g. with 'GC_') and this removes the namespace issues.  This pull request implements this approach for these new functions too.


yantosca:
This has been pulled into a branch.  We will perform difference tests to make sure that we get null differences.

yantosca:
I merged this PR into the dev/GC_13.0.0 branch, which will be eventually merged up to main for the 13.0.0 release.  

These updates result in zero differences w/r/t the reference code, as long as parallelization is turned off.  However, I believe that this is caused by an unrelated parallelization error (e.g. some variable not being held PRIVATE), which I will address in a separate issue.  Therefore I think we are OK to close out this PR.

PR #494, opened 2020-10-29
Simplify some met calculations

cdholmes:
Remove unnecessary assumptions about saturation vapor pressure from calc_met_mod.F90

By using specific humidity, we can avoid inconsistencies in the saturation vapor pressure formula between GEOS-Chem and the upstream meteorological model. 

Also fix a bug in potential temperature calculation.

Confirmed that transport still properly conserves mass after these updates.


lizziel:
Hi @cdholmes, this PR includes heterogeneous chemistry updates. Could you do those as a separate PR?

cdholmes:
PR #495 supersedes this one and excludes heterogeneous updates.

Issue #458, opened 2020-09-25
[DISCUSSION] GEOS-Chem pressures and water vapor


yantosca:
@cdholmes wrote:

>There appear to be some subtle problems in the way that the dry air pressures, masses, and box sizes are calculated in calc_met_mod.F90. Difference from the current approach are likely to be small, but could matter in very humid regions. I noticed these things when trying to resolve an issue that @chfite  was having. 
>
>I can fix all of these issues, but I want to make sure there aren’t other unintended consequences. What transport and mass conservation tests can we run to check the results? 
>
>1. The dry air mass is defined in calc_met_mod.F90 in a way that is not consistent with the total air mass (or pressure) and specific humidity. The calculation can also be much simpler, as shown on these few lines.

```
! Total air mass
ADwet = Area * ( Pbottom – Ptop ) / g0
! Dry air mass
AD = ADwet * ( 1 – SPHU * AIRMW / H2OMW )
```

>  Most of the remaining issues are just consequences of item 1. 
>
>2. The partial pressures of dry air and water vapor should be defined from total pressure and specific humidity (not using RH) to guarantee consistency with the mass distribution in the model. [The current approach will give the same result if and only if the Esat defined in GC is the same as GEOS-5. Switching to SPHU will ensure that GC remains consistent regardless of what GEOS-5 does now or if it changes in the future.]

```
! Correct approach
Ev = P * ( SPHU * AIRMW / H2OMW )
PDRY = P - Ev
```

>3. The virtual temperature (TV) used to determine grid box height and volume should be based on conditions at the grid cell center, not at the edge as currently done.  
>
>4. The pressure edges at edge L is currently defined by extrapolating DOWN from the midpoint of level L. You would get a different edge pressure if you extrapolated UP from the midpoint of level L-1. That doesn’t seem right. Unless there is a reason for this that I can’t see, this seems like a bug. I think the correct method is to use the dry air mass as specified above to calculate dry air pressure. 
>
>5. Potential temperature is typically defined at a fixed reference temperature (often 1000 hPa) not at surface pressure. GC currently uses surface pressure. For some applications, that might be OK, but as far as I can see, the potential temperature is only used in planeflight_mod, so it should definitely be set at a fixed temperature.
>
[...]

PR #495, opened 2020-10-29
[PULL REQUEST] 12.9.2+simplify calc met

cdholmes:
Remove unnecessary assumptions about saturation vapor pressure from calc_met_mod.F90

By using specific humidity, we can avoid inconsistencies in the saturation vapor pressure formula between GEOS-Chem and the upstream meteorological model.

Also fix a bug in potential temperature calculation.

Confirmed that transport still properly conserves mass after these updates.

### DIFF
--- a/GeosCore/calc_met_mod.F90
+++ b/GeosCore/calc_met_mod.F90
@@ -181,8 +181,6 @@ CONTAINS
 !  (2)  PEDGE_DRY (REAL(fp)) : Dry air partial pressure at box bottom     [hPa]
 !  (3)  PMID      (REAL(fp)) : Moist air pressure at grid box centroid    [hPa]
 !  (4)  PMID_DRY  (REAL(fp)) : Dry air partial pressure at box centroid   [hPa]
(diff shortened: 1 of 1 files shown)


lizziel:
Hi @cdholmes, thanks for this PR. We will not be able to get it in 13.0.0 but it will go into one of the early versions of 13. However, we'll want to thoroughly test mass conservation with it first. Have you done tests already, and if yes, could you post a description of all tests and results here?

cdholmes:
To check mass conservation, I ran a TransportTracers simulation for SF6, with transport on and all other processes off (emissions, chemistry, deposition, mixing). I simulated 1 hour with 10 minute transport steps. The final restart file was binary identical before and after applying this update. [This means that the saturation vapor pressure equation used in GC must be the same as in GEOS. With this update, we no longer need to assume or know the saturation vapor pressure in calc_met_mod.] 

This PR does not address all of the concerns that I raised in Discussion #458. It just simplifies calc_met_mod and eliminates an unnecessary assumption about saturation vapor pressure.   

This PR also fixes a bug in potential temperature calculation. That doesn't affect transport, but does affect output diagnostics.

msulprizio (approved):

PR #522, opened 2020-11-16
[PULL REQUEST] Updates of wet process treatments for LUO_WETDEP

ganluoasrc:
Wet process treatments described in [Luo et al. (gmd-13-2879-2020)](https://gmd.copernicus.org/articles/13/2879/2020/) have been updated for LUO_WETDEP.

### DIFF
--- a/GeosCore/convection_mod.F90
+++ b/GeosCore/convection_mod.F90
@@ -446,6 +446,9 @@ CONTAINS
 ! !DEFINED PARAMETERS:
 !
     REAL(fp), PARAMETER    :: TINYNUM = 1e-14_fp
+#ifdef LUO_WETDEP
+    REAL(fp), PARAMETER    :: pHRain = 5.6_fp
+#endif
 !
 ! !LOCAL VARIABLES:
 !
@@ -1072,6 +1075,9 @@ CONTAINS
                               K,         IC,         BXHEIGHT(K),   &
                               T(K),      QDOWN,      SDT,           &
                               F_WASHOUT, H2O2s(K),   SO2s(K),       &
+#ifdef LUO_WETDEP
+                              pHRain,                               &
+#endif
                               WASHFRAC,  AER,        Input_Opt,     &
                               State_Chm, State_Grid, State_Met,  RC )
 
--- a/GeosCore/drydep_mod.F90
+++ b/GeosCore/drydep_mod.F90
@@ -152,6 +152,9 @@ MODULE DRYDEP_MOD
   INTEGER                        :: id_MENO3, id_ETNO3
   INTEGER                        :: id_NK1
   INTEGER                        :: id_HNO3,  id_PAN,   id_IHN1
+#ifdef LUO_WETDEP
+  INTEGER                        :: id_H2O2,  id_SO2,   id_NH3
+#endif
 
   ! Arrays for Baldocchi drydep polynomial coefficients
   REAL(fp), TARGET               :: DRYCOEFF(NPOLY    )
@@ -418,6 +421,13 @@ CONTAINS
                 ! (cf. the GOCART model).  NOTE: In practice this will
                 ! only apply to the species SO2, SO4, MSA, NH3, NH4, NIT.
                 DVZ = MAX( DVZ, DBLE( SpcInfo%DD_DvzMinVal(1) ) )
+#ifdef LUO_WETDEP
+                IF ( DBLE( SpcInfo%DD_DvzMinVal(1) ) > 0.0_fp ) THEN
+                IF ( TC0(I,J) < 253.0_fp ) THEN
+                   DVZ = DBLE( SpcInfo%DD_DvzMinVal(1) )
+                ENDIF
+                ENDIF
+#endif
 
              ENDIF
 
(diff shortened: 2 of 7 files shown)


yantosca:
Thanks @ganluoasrc .  We are benchmarking 13.0.0 for release at present but hope to get this into an early 13 series release.

ganluoasrc:
Thanks Bob. Sounds good to me.

Best

Gan
________________________________

jdhask:
I've been interested in running GCClassic v13.0.0 with the Luo et al., wet scavenging updates. I've taken the files added by Gan Luo above, resolved all conflicts between the version they were made on and v13.0.0, and fixed a few bugs I found when trying to compile. Using the uploaded files, I was able to complete a v13.0.0 run with the Luo et al., wet dep scheme after toggling the "Default" in CMakeLists.txt to "ON" for LUO_WETDEP (since I couldn't find a toggle switch anywhere else?). The resulting output did have lower NIT, for example, than an identical run I did with the default wet dep scheme. I believe these updates keep the integrity of the update Gan Lou provided above, but he should probably check to make sure! Not sure how things in the stratosphere are doing, though. Hope this helps speed the integration of this update for regular toggle on/off usage. 

My fork with these updates is here: [feature/gan-luo_wetdep](https://github.com/geoschem/geos-chem/compare/main...jdhask:feature/gan-luo_wetdep)

### Description of Changes: 

#### GeosCore/**sulfate_mod.f90**
    
##### *Resolving Conflicts:*
   The version uploaded above has pointer variables (pHCloud, QLxpHCloud, isCloud) which point to State_Chm variables.  Across v13.0.0 in this file, other pointers have been removed and are referenced simply as their State_Chm variable names, so following this convention, I removed these pointers, and updated their references to be their respective State_Chm variable names. 
    
   These pointer variables were then reset to 0 (for safety). There is a new function within state_chem_mod.f that initializes variables and sets them to 0s there, so this is now redundant and was removed for any 0 values. 
   
   I replaced all occurrences in Gan Lou's code to these pointer variables with their State_Chm variable names. This was only relevant for pHCloud. 
   
##### *Bug fixes* (found when compiling)
   I updated dust calculations to use: `State\_Chm%SpcData(id\_DST1)%Info%MW_g`
instead of `State\_Chm%SpcData(id\_DST1)%Info%emMW_g`
which no longer exists as a State_Chm variable. (new lines # 3077 - 3103)
   
   
#### GeosCore/**drydep_mod.f90**
##### *Resolving Conflicts:*
   The version uploaded has a pointer variable DVEL which points to a State_Chm variable. Similar to updates in sulfate_mod.f90, these pointers were removed and references to them were changed to their respective State_Chm variable. Additionally, the initialization to 0 is removed, as it has already been preformed. This variable is not referenced in the updates provided above, so this was a trivial conflict. 
   
##### *Bug fixes* (found when compiling)
   The variable `TC0(I,J)` in subroutine `UPDATE_DRYDEPFREQ` had not been declared as a local variable, so I added that on new lines # 409-411.
   
   Similarly, the variable `LSNOW(I,J)` had not been declared as a local variable in the subroutine `DEPVEL` on new line # 1141, so I added it.
   
####  Headers/**state_chm_mod.f90**
 ##### *Resolving Conflicts:*
 This is the file with the largest conflict. The problems arise because the subroutine `Register_ChmField` has been superseded by the subroutine `Init_and_Register` in v13.0.0.  Within an `#ifdef LUO_WETDEP` statement, I over-write the value for State_Chm%phCloud (which was initialized prior in the code to 0) to be 5.6_fp (this may be inappropriate here?). Then I call `Init_and_Register` for `KRATE`, `pHRain`, `QQpHRain` , `QQRAin` , and delete the `Init_and_Register` call for `PSO4s`.

ganluoasrc:
Dear Jessica,

Thanks for the updates. I quickly had a look at your updates in GeosCore/drydep_mod.f90

At line 410, you declared TC0(State_Grid%NX,State_Grid%NY) in SUBROUTINE UPDATE_DRYDEPFREQ which was used in GC12.9.3 but is no longer in GC13.0.0. It will give you TC0=0.0 which is not right. The right way is to modify line 511 as following:
IF ( State_Met%TS(I,J) < 253.0_fp ) THEN

At line 1141, you declared LSNOW(State_Grid%NX,State_Grid%NY) in SUBROUTINE DEPVEL but did not assign any value to it as that done in GC12.9.3.
So, you need to assign LSNOW at somewhere as following:
LSNOW(I,J)  = ( State_Met%ALBD(I,J) > 0.4 )

Similar problems may exist in other codes. If you need my help to check them, please let me know.

Best

Gan

jdhask:
Thank you! I updated both of those lines in my fork. They were both lines I added after I got errors when compiling. The other ambiguity I had questions about was about initializing `State_Chm%pHRain` to a value of 5.6 instead of 0. In my fork, this is done in `Headers/state_chm_mod.F90` (line 1659) and also in `sulfate_mod.f90` (line 2488)- I wasn't sure which took precedent or if both were necessary?

ganluoasrc:
Headers/state_chm_mod.F90 (line 1659) is re-initialization of State_Chm%pHCloud from 0 to 5.6. In my opinion, pHs for cloud are generally varied from 4 to 6 in the atmosphere. Therefore, a initial value of 5.6 sounds more reasonable to me.

State_Chm%pHRain needs to be reset to 5.6 at every chemistry time step in sulfate_mod.f90 (line 2488) due to the changing of cloud chemistry in the atmosphere.

yantosca:
@ganluoasrc, I am starting to merge this PR into GEOS-Chem.  Could you provide the metadata (long name, units) for the following fields:

  1. KRATE
  2. pHrain
  3. QQpHrain
  4. QQrain

These are needed so that if these fields are saved to netCDF, they will have the proper attributes.

yantosca:
@ganluoasrc never mind, I found it lower down in the code.

yantosca:
@ganluoasrc: I noticed that there are several lines like this where you use `(I,J,L:)` in SUM statements.  
```F90
IF(SUM(State_Chm%QQ3D(I,J,L:)*State_Met%BXHEIGHT(I,J,L:))>1.D-30&
```
Is this equivalent to I think what you mean is `(I,J,L:State_Grid%NX)`?  The reason why I ask is because the compiler that I am using (gfortran 10.2.0) does not like this syntax.

ganluoasrc:
Yes, the right way is '(I,J,L:State_Grid%NZ)'.

Best,

Gan
________________________________

yantosca:
I will close this PR as it has been now superseded by PR #779.  The merging had to be done manually.

PR #530, opened 2020-11-19
Quality-of-life improvements for createRunDir.sh

sdeastham:
Creating run directories in createRunDir.sh now allows:
1. Tab-completion (using `read -e`) when the run directory (or ExtData) is requested
2. If the run directory does not exist but its parent does, the user can opt to create it

This should help to reduce some frustration around run directory creation without risking users accidentally creating monster paths because of the combination of `mkdir -p` and a typo early in the directory path.

### DIFF
--- a/run/GCHPctm/createRunDir.sh
+++ b/run/GCHPctm/createRunDir.sh
@@ -46,7 +46,7 @@ if [[ -z "${GC_DATA_ROOT}" ]]; then
     valid_path=0
     while [ "$valid_path" -eq 0 ]
     do
-	read extdata
+	read -e extdata
 	if [[ ${extdata} = "q" ]]; then
 	    printf "\nExiting.\n"
 	    exit 1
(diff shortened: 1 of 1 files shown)


yantosca:
Thanks for writing @sdeastham.  I have more or less already applied these updates in https://github.com/geoschem/geos-chem/blob/dev/13.0.0%2BiTests/run/GCClassic/createRunDir.sh, which right now is in dev/13.0.0+iTests (soon to be merged into dev/13.0.0).  

I also added a regexp to replace the ~ in a path with the ${HOME} variable according to this post: https://stackoverflow.com/questions/3963716/how-to-manually-expand-a-special-variable-ex-tilde-in-bash/27485157#27485157

sdeastham:
Thanks Bob! I'd still like to suggest that we use `read -e` for ExtData, and that we have the logic which allows creation of the run directory. I find it fiddly to have to first create a directory that I want to use for my runs, and only _then_ run `createRunDir.sh`.

yantosca:
OK we can pull it in then.  Thanks!

yantosca:
Now merged  into dev/13.0.0

PR #582, opened 2021-01-04
[PULL REQUEST] Add co2 run

TerribleNews:
Companion pull request to https://github.com/geoschem/GCHP/pull/71

This is the submodule I'm least confident about. I had to pull a number of times and each time new conflicts came up.


lizziel:
I am going to work on bringing in all the adjoint updates this week. @TerribleNews, could you change the target branch to dev?

msulprizio:
> could you change the target branch to dev?

I went ahead and did this @lizziel. Anyone can do it by clicking 'Edit' at the top of the PR.

TerribleNews:
> I am going to work on bringing in all the adjoint updates this week.

Great! 

> 
> 
> > could you change the target branch to dev?
> 
> I went ahead and did this @lizziel. Anyone can do it by clicking 'Edit' at the top of the PR.

I was also hoping to get some changes I committed past the pull request included, but I couldn't figure out how to do it. Should I edit the PR or would you like to check it out first? The additional changes should be real easy, they're all to files I created in the run directory.

msulprizio:
@TerribleNews I believe if you push any changes to your fork's `add_co2_run` branch, it should automatically update the PR.

TerribleNews:
Oh, then never mind!

lizziel:
Thanks all! @TerribleNews, a few questions/comments:
1. I noticed a couple instances of `#ifdef FALSE` in your updates. Are those intentional or are they supposed to be `#ifdef ADJOINT`? 
2. Some of the changes appear to be debugging changes. Could you browse through the changes (see "Files Changed" tab above) and make sure no temporary debug updates are still present?
3. It doesn't look like the `USE` statement for `MAPL_CommsMod` is in an `ifdef MAPL` block. Could you double-check that you don't have any MAPL libraries or other GCHP-specific framework code that might be run in GC-Classic?

Let me know when any edits and other updates are all in. I'll hold off on bringing this in until I get the go-ahead from you.

LiamBindle:
Is it possible to have a runtime setting that enables/disables the adjoint? Does it take a lot longer to compile with the adjoint code? Just thinking it's best we avoid preprocessor definitions unless it is necessary.

TerribleNews:
> 
> 
> Thanks all! @TerribleNews, a few questions/comments:
> 

Okay, I'll look into those and let you know when I've sorted them out! Thank you!


> Is it possible to have a runtime setting that enables/disables the adjoint? Does it take a lot longer to compile with the adjoint code? Just thinking it's best we avoid preprocessor definitions unless it is necessary.

In fact, there are both compile-time and run-time flags. The preprocessor flags are less about a longer compilation step and more about runtime performance. There are things the adjoint code needs to to during the forward run to prepare for the reverse integration. Our assumption is that most of the GEOS-Chem community will never need to run the adjoint, and there are memory and performance implications for using the adjoint code in forward mode, so we want to keep those declarations and code blocks out of most people's GEOS-Chem. If you want to do adjoint operations, you build a binary with the adjoint compile-time flags and then use that for the forward and reverse integrations with different runtime configurations.

lizziel:
@TerribleNews, do you have a time estimate for when you will get to this? We would like to include the updates in 13.1 which we will be wrapping up in the next couple weeks. If it is not yet ready we could push it to a later release if needed.

TerribleNews:
> 
> 
> @TerribleNews, do you have a time estimate for when you will get to this? We would like to include the updates in 13.1 which we will be wrapping up in the next couple weeks. If it is not yet ready we could push it to a later release if needed.

I would really like to get this in for the next minor release cycle, so I will do my best to get it back to you sometime this week.

GeosCore/hco_interface_gc_mod.F90
```diff
@@ -423,10 +423,25 @@ SUBROUTINE HCOI_GC_Init( Input_Opt, State_Chm, State_Grid, &
...
+! #ifdef ADJOINT
+!     else
+!     ! Emission, chemistry and dynamics timestep in seconds
+!     HcoState%TS_EMIS = -GET_TS_EMIS()
+!     HcoState%TS_CHEM = -GET_TS_CHEM()
+!     HcoState%TS_DYN  = -GET_TS_DYN()
+!     ! Look into whether we want to change the sign in the body of GET_TS_*()
+!     endif
+! #endif
+
```
lizziel, line 444: Keep this even though it is commented out?

GeosCore/hco_interface_gc_mod.F90
```diff
@@ -1180,6 +1195,7 @@ SUBROUTINE HCOI_GC_WriteDiagn( Input_Opt, Restart, RC )
 
     USE Time_Mod,        ONLY : Get_Year, Get_Month, Get_Day, GET_DAY_OF_YEAR
     USE Time_Mod,        ONLY : GET_HOUR, GET_MINUTE, GET_SECOND
+    USE MAPL_CommsMod,   ONLY : MAPL_AM_I_ROOT
```
lizziel, line 1198: This will need to be in a MAPL ifdef so that GC-Classic does not execute it

GeosCore/mixing_mod.F90
```diff
@@ -206,6 +206,9 @@ SUBROUTINE DO_TEND( Input_Opt, State_Chm,    State_Diag, State_Grid, &
     USE ERROR_MOD,            ONLY : SAFE_DIV
     USE GET_NDEP_MOD,         ONLY : SOIL_DRYDEP
     USE HCO_Interface_Common, ONLY : GetHcoDiagn
+#ifdef FALSE
```
lizziel, line 209: Did you intend for this to use FALSE?

GeosCore/mixing_mod.F90
```diff
@@ -470,7 +502,17 @@ SUBROUTINE DO_TEND( Input_Opt, State_Chm,    State_Diag, State_Grid, &
        ! dry deposition and/or emissions
        !--------------------------------------------------------------------
        IF ( .NOT. DryDepSpec .AND. .NOT. EmisSpec ) CYCLE
-
+       
+#ifdef FALSE
```
lizziel, line 506: Same question as before about FALSE.

Interfaces/GCHP/Chem_GridCompMod.F90
```diff
@@ -4393,6 +4626,7 @@ SUBROUTINE Run_( GC, Import, Export, Clock, Phase, RC )
        _VERIFY(STATUS)
     ENDIF
     CALL CopyGCStates2Exports( am_I_Root, Input_Opt, HistoryConfig, STATUS )
+    IF ( MAPL_am_I_Root() ) WRITE(*,*) "Back from CopyGCSates2Exports"
```
lizziel, line 4629: Do you need this print? If done every timestep it will clutter the log.

Interfaces/GCHP/Chem_GridCompMod.F90
```diff
@@ -4438,6 +4672,8 @@ SUBROUTINE Run_( GC, Import, Export, Clock, Phase, RC )
             '### ', a, '  |  Execution on PET # ',      i5.5, / &
             '###' )
 
+    IF (Mapl_am_i_root()) WRITE(*,*) "Returning from Run_"
+
```
lizziel, line 4676: Same question as above about cluttering the log.

Interfaces/GCHP/Chem_GridCompMod.F90
```diff
@@ -5283,7 +5602,7 @@ SUBROUTINE Extract_( GC,         Clock,    Grid,    MaplCF, GeosCF,    &
        ! Get the solar zenith angle and solar insolation
        ! NOTE: ZTH, SLR are allocated outside of this routine
        CALL MAPL_SunGetInsolation( LONS  = lonCtr,    &
-                                   LATS  = latCtr,    &
+            LATS  = latCtr,    &
```
lizziel, line 5605: Looks like the only the spacing here changed. Was that intended?

Interfaces/GCHP/gchp_chunk_mod.F90
```diff
@@ -1134,4 +1571,97 @@ SUBROUTINE GCHP_Chunk_Run( GC,                                             &
 
   END SUBROUTINE GCHP_Chunk_Run
 !EOC
+
+!BOP
+  SUBROUTINE GCHP_PRINT_MET(I, J, L,         &
```
lizziel, line 1576: This could be reused for GC-Classic and stored in a more generic module. Can keep for now and I'll move it later.

Interfaces/GCHP/gchp_chunk_mod.F90
```diff
@@ -1134,4 +1571,97 @@ SUBROUTINE GCHP_Chunk_Run( GC,                                             &
...
+    CHARACTER(LEN=255) :: ErrorMsg, ThisLoc
+
+
+    !=========================================================================
+    ! GCHP_PRINT_MET begins here!
+    !=========================================================================
+
+    ErrorMsg  = ''
+    ThisLoc   = ' -> at Print_Global_Species_Kg (in module ' // &
+         'GeosUtil/unitconv_mod.F)'
```
lizziel, line 1622: Location and name is wrong here.

Interfaces/GCHP/gchp_historyexports_mod.F90
```diff
@@ -159,7 +159,7 @@ SUBROUTINE Init_HistoryConfig ( am_I_Root, HistoryConfig, configFile, RC )
        RETURN
     ENDIF
     ! Optional debugging
-    !CALL Print_DiagList( am_I_Root, HistoryConfig%DiagList, RC )
+    CALL Print_DiagList( am_I_Root, HistoryConfig%DiagList, RC )
```
lizziel, line 162: This should stay commented out

Interfaces/GCHP/gchp_historyexports_mod.F90
```diff
@@ -813,7 +813,10 @@ SUBROUTINE CopyGCStates2Exports( am_I_Root, Input_Opt, HistoryConfig, RC )
     ! Loop over the History Exports list
     current => HistoryConfig%HistoryExportsList%head
     DO WHILE ( ASSOCIATED( current ) )
-
+       
+       if (MAPL_Am_I_Root()) THEN
+          print *, '  Copying ' // TRIM(current%name)
+       endif
```
lizziel, line 819: Keep but comment out.

Interfaces/GCHP/gchp_historyexports_mod.F90
```diff
@@ -865,6 +868,7 @@ SUBROUTINE CopyGCStates2Exports( am_I_Root, Input_Opt, HistoryConfig, RC )
 
     ! Copy emissions data to MAPL exports via HEMCO
     CALL HCOI_GC_WriteDiagn( Input_Opt, .FALSE., RC )
+    IF ( MAPL_Am_I_Root() ) WRITE(*,*) "Back from HCOI_GC_WriteDiagn, RC = ", RC
```
lizziel, line 871: Keep but comment out.

run/GCHP/GCHP.rc.template
```diff
@@ -37,7 +37,7 @@ DYNAMICS_DT: 600
 SOLARAvrg: 0
 IRRADAvrg: 0
 
-GCHPchem_REFERENCE_TIME: 121000
+GCHPchem_REFERENCE_TIME: 001000
```
lizziel, line 40: Why is this changing?

run/GCHP/GCHP.rc.template
```diff
@@ -60,8 +60,8 @@ BKG_FREQUENCY: 0
 #                             Can exceed 24 hours (e.g. 1680000 for 7 days)
 # Record ref date (YYYYMMDD): Reference date; set to before sim start date
 # Record ref time (HHMMSS)  : Reference time 
-RECORD_FREQUENCY: 1680000 
-RECORD_REF_DATE: 20000101
+RECORD_FREQUENCY: 100000000 
+RECORD_REF_DATE: 20140901
```
lizziel, line 64: Why is this changing?

run/GCHP/input.geos.templates/input.geos.CO2
```diff
@@ -0,0 +1,103 @@
...
+Turn on Cloud Conv?     : T
+Turn on PBL Mixing?     : T
+ => Use non-local PBL?  : T
+------------------------+------------------------------------------------------
+%%% EMISSIONS MENU %%%  :
+HEMCO Input file        : HEMCO_Config.rc
+Switches for UCX        :---
+ => Use CH4 emissions?  : F
+ => Set init. strat. H2O: F
+------------------------+------------------------------------------------------
```
lizziel, line 40: The emissions menu is no longer included in input.geos.

run/GCHP/runConfig.sh.template
```diff
@@ -120,8 +120,8 @@ INITIAL_RESTART=initial_GEOSChem_rst.c${CS_RES}_{SIMULATION}.nc
 #------------------------------------------------------------
 Turn_on_Chemistry=T
 Turn_on_emissions=T
-Turn_on_Dry_Deposition=T
-Turn_on_Wet_Deposition=T
+Turn_on_Dry_Deposition=F
+Turn_on_Wet_Deposition=F
```
lizziel, line 124: Keep as T

TerribleNews:
Thanks @lizziel for the detailed feedback. It's very helpful. In hindsight, several of these were obvious things I should have checked for but since this is my first rodeo I didn't really know to check for them or really how to check for them. I have a better understanding of the interface now, so thank you! I'm working on these and the ones in the other repos and I will let you know when all the fixes are checked in.

CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ endif()
 #-----------------------------------------------------------------------------
 # Print header
 #-----------------------------------------------------------------------------
-get_repo_version(GC_REPO_VERSION ${CMAKE_CURRENT_SOURCE_DIR})
+# get_repo_version(GC_REPO_VERSION ${CMAKE_CURRENT_SOURCE_DIR})
```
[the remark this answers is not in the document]
TerribleNews: Ah, yes. I commented this out because if I don't, I get:
```
CMake Error at src/GCHP_GridComp/GEOSChem_GridComp/geos-chem/CMakeLists.txt:20 (get_repo_version):
  Unknown CMake command "get_repo_version".
``` when I run cmake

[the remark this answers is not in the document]
Comment (TerribleNews) [reply CMakeLists.txt:20]: @LiamBindle has helped me fix this so it's uncommented plus the lines that include the library to avoid the error

**Comment by lizziel** [comment]
Excellent, thanks @TerribleNews.

**Comment by TerribleNews** [comment]
I believe I have dealt with all of @lizziel 's comments on the code. I now have to look into the changes I made to the common run files.

File: run/GCHP/runConfig.sh.template (hunk 16 above)
Comment (lizziel) [reply run/GCHP/runConfig.sh.template:124]: You will also need to remove the line right above this: Turn_on_emissions=T

**Comment by lizziel** [comment]
The conflict in mixing_mod.F90 is due to this. It is simple to fix by including `ErrorMsg` declaration.
```
<<<<<<< HEAD
    CHARACTER(LEN=255) :: ErrMsg, ErrorMsg, ThisLoc
=======
    CHARACTER(LEN=255) :: ErrMsg, ThisLoc
#ifdef ADJOINT
    LOGICAL                 :: IS_ADJ
#endif
>>>>>>> b1076ea20cf9c5e163485facd2f243dcec3bb0a7
```

The conflict in `runConfig.sh.template` is because we have since removed the toggle to turn on/off emissions in `input.geos`, and thus took it out of `runConfig.sh`.

File: run/GCHP/GCHP.rc.template (hunk 13 above)
Comment (TerribleNews) [reply run/GCHP/GCHP.rc.template]: I have removed these and the other changes to common run files for now.

**Comment by lizziel** [comment]
Great, thanks. For the `RECORD` and drydep/wetdep settings, we can handle setting those to what you have here during run directory creation. The backend for run directory creation is about to change so you can leave these as is and we can do the changes. Is there anything else that needs to change in common run directory file templates (`GCHP.rc`, `runConfig.sh`, etc) for CO2?

**Comment by TerribleNews** [comment]
Yes. Also I apologize for shoehorning two different but strongly intertwined features here. It should be possible to separate the CO2 and adjoint run configurations in future, but they kind of got developed together, so it will take some work. For now, if you compile without the `ADJOINT `and `REVERSE_OPERATOR `flags (i.e. regular old GCHP), it should still be able to run with the CMSFlux inputs in CO2 only mode for forward. But currently the adjoint code (i.e. complied with `-DADJOINT=yes -DREVERSE_OPERATORS=yes`) will only work with a CO2 run directory.  

In any event, when creating a rundirectory that might be used for adjoint (which is currently the CO2 option in `createRunDir.sh`) there should be the following added to `GCHP.rc`
```
#
# %%% Adjoint variables
#
MODEL_PHASE: FORWARD
```
This is required for the runConfig.sh and runConfig_adj.sh to have a variable to replace. All the other adjoint options are currently manually defined.

**Comment by TerribleNews** [comment]
Also, thank you @lizziel for guiding me through all this, I really appreciate the extra effort!

**Comment by lizziel** [comment]
No problem. I'm glad you don't mind the pushback. I'm testing GC-Classic and GCHP now (non-adjoint runs) to ensure zero diff. For GC-Classic I had to change part of input_mod.F90. See next review comment.

File: GeosCore/input_mod.F90 [hunk 18]
```diff
@@ -5600,6 +5600,19 @@ SUBROUTINE CHECK_TIME_STEPS( Input_Opt, State_Grid, RC)
...
+         ! TS_DYN and TS_CHEM should always be set to something valid...?
+         TS_DYN = TS_DYN * -1
+         TS_CHEM = TS_CHEM * -1
+         if (LTURB .or. LCONV) TS_CONV = TS_CONV * -1
+         if (LEMIS .or. LDRYD) TS_EMIS = TS_EMIS * -1
+         ! I'm not sure which flag determins radiation on/off?
+         if (LCHEM) TS_RAD = TS_RAD * -1
+      endif
+         
+
```
lizziel, line 5615: For GC-Classic to build and run I needed to change this section to this:
```
    ! If we're doing the reverse integration                                              
    ! multiply all the timesteps by -1 here                                               
    IF ( TS_DYN < 0 ) THEN
       TS_CHEM = TS_CHEM * -1
       TS_EMIS = TS_EMIS * -1
       TS_CONV = TS_CONV * -1
       TS_DYN  = TS_DYN  * -1
       TS_RAD  = TS_RAD  * -1
    ENDIF

```

lizziel: Otherwise this looks good to go. If you could just push the update I will go ahead and merge. Thanks for your work on this!

lizziel:
Hi @TerribleNews, I have a follow-up question about the changes in this PR. There is now `input_opt%is_adjoint`, but only if built with `ADJOINT`. I want to clean up unitconv_mod.F90 so that `IS_ADJOINT` logical is not passed around. Would `Input_Opt%is_adjoint` always be `.TRUE.` if compiled with `ADJOINT` or did you intend it to be flexible based on a config file setting even when compiling with `ADJOINT`?

TerribleNews:
Oh yeah, @lizziel, there's some confusing terminology there, so I apologize. I tried to keep things clear and separate where I could, but the field just kind of has some built-in ambiguities. Tldr: the pre-processor `ADJOINT` flag and `input_opt%is_adjoint` local are separate and both required.

A full adjoint integration consists of a forward model run followed by a inverse or adjoint run, i.e., the reverse-time run with the adjoint calculations. In the old GEOS-Chem adjoint, the forward and inverse runs were done with a single execution, but we decided in this case that we would allow the forward and inverse runs to be two separate executions. In doing it this way, we had the option of creating two binaries (e.g., `gchp` and `gchp_adj`) or one with a runtime option. We chose the latter option, so both the forward and inverse runs execute `gchp` but with different values in the `MODEL_PHASE:` variable in `GCHP.rc`. 

In addition, the forward model run for an adjoint integration requires extra operations from the non-adjoint-enabled forward model, so executing a binary compiled with `-DADJOINT` but `MODEL_PHASE: FORWARD` (i.e., `input_opt%is_adjoint = .false.`) is different from just running `gchp` compiled without `-DADJOINT`. 

The actual potential redundancy is having two pre-processor flags, `-DADJOINT` and `-DREVERSE_OPERATORS`, but I left them because they could be separate changes to the code (although the way they are currently setup, the only `#ifdef REVERSE_OPERATORS` blocks are nested inside `#ifdef ADJOINT` blocks).

Summary: `-DADJOINT` means adjoint-enabled binary and `input_opt%is_adjoint` means "are we running the forward or inverse integration?

lizziel:
Got it, thanks! My change will be to always define `input_opt%is_adjoint` (whether or not built with `-DADJOINT`) and pass `input_opt` to unit conversion subroutines rather than local logical `IS_ADJOINT`. I'll keep the `ADJOINT` ifdefs as is in `unitconv_mod.F90`, but replace `IS_ADJOINT` logical with `input_opt%is_adjoint` within the conversions. Let me know if you see any issues with this for what you are doing, but I think it should work out.

PR #595, opened 2021-01-21
Added new logging system files to run directory creation for GCHP

sdeastham:
Run directory creation with MAPL 2.5.0 now requires that the file `logging.yaml` be present. This file configures the logger (from `pFlogger`), which works almost exactly like the Python `logger`.

### DIFF
--- a/run/GCHP/logging.yaml
+++ b/run/GCHP/logging.yaml
@@ -1,0 +1,174 @@
+schema_version: 1
+
+###############################      
+locks:
+   mpi:
+      class: MpiLock
+      comm:  MPI_COMM_WORLD
+
+###############################      
+formatters:
+   legacy:
+      class: Formatter
+      format:  '%(message)a'
+
+   basic:
+      class: Formatter
+      format:  '%(short_name)a15~: %(level_name)a~: %(message)a'
+
+   mpi:
+      class: MpiFormatter
+      format:  '%(mpi_rank)i4.4~: %(name)~: %(level_name)a~: %(message)a'
+      comm: MPI_COMM_WORLD
+
+   column:
+      class: Formatter
+      format: '(%(i)i3.3,%(j)i3.3): %(level_name)'
+
+
+
+###############################      
+handlers:
+
+   console:
+      class: streamhandler
+#      formatter: legacy
+      formatter: basic
+      unit: OUTPUT_UNIT
+      level: INFO
+      
+   warnings:
+      class:  FileHandler
+      filename: warnings_and_errors.log
+      lock: mpi
+      level: WARNING
+      formatter: basic
+
+   errors:
+      class:  StreamHandler
+      formatter: basic
+      unit: ERROR_UNIT
+      level: ERROR
+
+   mpi_shared:
+      class: FileHandler
+      filename: allPEs.log
+      formatter: mpi
+      comm: MPI_COMM_WORLD
+      lock: mpi
+      rank_keyword: rank
+      level: DEBUG
+
+   mpi_debug:
+      class: MpiFileHandler
+      formatter: basic
+      filename:  debug_%(rank)i3.3~.log
+      comm: MPI_COMM_WORLD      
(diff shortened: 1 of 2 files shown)


lizziel:
I am changing the assignment to me from @LiamBindle since I am doing this update as part of the larger GMAO library updates work for GCHP.

lizziel:
@sdeastham, could you provide information on what you tested in pFlogger within GCHP by adjusting the file? For example, how did you replace the debug level we currently recommend for debugging ExtData? This replacement should be part of the update to the run directory.

sdeastham:
Hi Lizzie - I performed a few different tests, but focused on ensuring that I could reproduce something like our current ExtData behavior. Specifically, setting `root_level` to `DEBUG` in the `CAP.EXTDATA` logger should (following integration of my ExtData updates, which I submitted as separate pull request I believe) result in the root process spitting all that info to the allPEs.log file; changing `level` to `debug` for `CAP.EXTDATA` should then provide the new functionality of having *every* PE report into allPEs.log (although you may prefer to use the `mpi_debug` handler for that purpose).

lizziel:
Ok, great. I am thinking of keeping the ExtData debug entry in `runConfig.sh` exactly the same for now until we have more thoroughly explored all of the pFlogger capabilities. But rather than update `ExtData.rc` it would update the logging file instead. And only the root would print to `allPEs.log` (although I'd like to change the name of that file if it is only going to store prints from root).

Do you have a preference about users configuring commonly changed settings from `runConfig.sh` versus editing `logging.yaml` directly?

sdeastham:
Personally I'd rather have them tweak `logging.yaml`, but that's somewhat dependent on how much we end up using pFlogger. For example, we may want to consider having GEOS-Chem's logging routines pass through to pFlogger when running with GCHP. I'm not particularly attached to either approach though - I acknowledge that for most users they'd rather have to deal with one file rather than ten, and I'm just a bit of a micromanager!

On the question of `allPEs.log`, it should still get error messages from all processes, not just root. At the moment we're just not using pFlogger enough to see it, I think. I need to talk to Tom about the correct way to pass errors etc through pFlogger. You'd need to use the `mpi_debug` handler to generate a file which only gets messages from the root process.

lizziel:
Okay, for 13.1 we will keep the simple ExtData debug toggle in `runConfig.sh`. I think the longer term goal should be to provide a guide for configuring pFlogger, have users edit `logging.yaml` directly, and route GEOS-Chem debug messaging through it.

GCST has talked before about potentially adopting pFlogger for GEOS-Chem. Right now we have if/else statements and a toggle in `input.geos`. The capabilities of pFlogger are more powerful and would make the code a lot cleaner. I'll start a github discussion about this in the geos-chem repo.

run/GCHP/logging.yaml
```diff
@@ -0,0 +1,174 @@
...
+   MAPL:
+       handlers: [mpi_shared]
+       level: INFO
+       root_level: DEBUG
+       
+   CAP.EXTDATA:
+       handlers: [mpi_shared]
+       level: WARNING
+       root_level: WARNING
+       propagate: 0
```
lizziel, line 96: Could you change this from 0 to false and push to your branch? Using 0 results in a type mismatch error in the updated libraries I am using.

lizziel, line 96:
```suggestion
       propagate: false
```

sdeastham: Got it - I think that should now be done?

lizziel: Thanks!

lizziel:
`Okay, for 13.1 we will keep the simple ExtData debug toggle in runConfig.sh.`
I changed my mind on this one. After looking at `runConfig.sh` I realized all other run-time debug print options need to be set manually (`input.geos`, `HEMCO_Config.rc`). Now the debug section of `runConfig.sh` will just be comments listing what files to change. It will point to the docs for `logging.yaml` when they exist.

PR #651, opened 2021-03-12
[PULL REQUEST] Remove all code for RRTMG binary diagnostics

lizziel:
This PR removes RRTMG binary diagnostic code. NetCDF diagnostics are now implemented and validated to replace all RRTMG binary diagnostics, but we are waiting until after the 13.0 release to remove the binary.  I will pull this in to dev/13.1.0 once we start working on it.


msulprizio (approved):

lizziel:
Validation of the netcdf diagnostics was performed by comparing 1-day mean output against the equivalent binary output using GEOS-Chem Classic 13.0. Results are viewable at:
[http://ftp.as.harvard.edu/gcgrid/geos-chem/validation/RRTMG_bpch_vs_netcdf/13.0/gcc_13.0_rrtmg_bpch_vs_nc.pdf](http://ftp.as.harvard.edu/gcgrid/geos-chem/validation/RRTMG_bpch_vs_netcdf/13.0/gcc_13.0_rrtmg_bpch_vs_nc.pdf)

All RRTMG diagnostic differences are small numerical differences introduced within the model due to precision differences in the internal arrays used to calculate the data. GEOS-Chem History calculates the data output to netCDF format and uses REAL*8 internal arrays for accumulating and averaging data. GEOS-Chem legacy code for calculating data output to binary format using REAL*4.

PR #661, opened 2021-03-19
[PULL REQUEST] Bug fixes for GEOS-Chem Classic dry-run simulations

yantosca:
### Overview

This PR fixes several issues in the GEOS-Chem "Classic" dry-run simulation, namely:

  1. Routine `INIT_UCX` in (`GeosCore/ucx_mod.F90`), was printing incorrect dry-run output to the log file, which caused dry-run simulations to hang.  This has now been corrected.
  
  2. The `run/GCClassic/createRunDir.sh` script has the following updates:
    a.  Aerosol simulations now use the fullchem restart file
    b.  POPs simulation now looks for an initial restart file w/ the extra option (BaP, PYR, PHE) in` ExtData/GEOSCHEM_RESTARTS/v2020-02`.  
    c. All other specialty simulations use restart files in `GEOSCHEM_RESTARTS/v2020-02` on the remote server
    d. Send error output from `aws s3 cp` to `dev/null`

  3. The `run/shared/download_data.py` script has the following updates:
    a. Look for NK15 and NK40 in input.geos to decide if we need to download data files for TOMAS15 and TOMAS40 simulations
    b. Use the fullchem restart file (Jan 2019) for the aerosol simulation
    c. Look for specialty simulation restart files in `GEOSCHEM_RESTARTS/v2020-02`
    d. When downloading data from ComputeCanada, first download it to the local `GEOSCHEM_RESTARTS` directory tree, then copy it to the run directory.
    e. AWS restart files should now be placed in the run directory by `createRunDir.sh`.  But if a restart file is missing, running
       `download_data.py` with `-aws` will restore it.
  4. In `state_chm_mod.F90`, we now nullify the `State_Chm%HetInfo` and `State_Chm%Map_SpeciesBC` fields in routine `Init_State_Chm`.  Explicitly setting these fields to `NULL()` will make them be recognized as unallocated by the compiler, which can prevent seg faults during finalization, at least with the later GNU Fortran compilers.

  5. Also note: 2 x 2.5 specialty simulation restart files have been copied to `ExtData/GEOSCHEM_RESTARTS/v2020-02`.  The reference time of these restart files has been edited to either 2019-01-01 or 2019-07-01.

### Testing

The dry-run capability was tested on the AWS cloud.  The proper restart files are now being placed into the run directories.

CAVEAT: There may be still some issues with the POPS simulation, since there are 3 separate restart files.  But since the POPs simulation is basically unused, I think we can address any user-reported issues that come up on a case-by-case basis.

### DIFF
--- a/GeosCore/ucx_mod.F90
+++ b/GeosCore/ucx_mod.F90
@@ -49,7 +49,6 @@ MODULE UCX_MOD
 !
 ! !PUBLIC DATA MEMBERS:
 !
-  PUBLIC :: NOON_FILE_ROOT ! Directory for noontime data
   PUBLIC :: T_STS          ! Max temperature of STS formation (K)
   PUBLIC :: NDENS_AER      ! See below
 !
@@ -144,7 +143,6 @@ MODULE UCX_MOD
   !=================================================================
 
   ! Scalars
-  CHARACTER(LEN=255)   :: NOON_FILE_ROOT
   REAL(fp)             :: SLA_VA
   REAL(fp)             :: SLA_RR
   REAL(fp)             :: SLA_VR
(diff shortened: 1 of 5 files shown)


msulprizio (approved):
I have reviewed this PR and everything looks good to me.

PR #682, opened 2021-04-07
GCHP no diff run directory updates

lizziel:
This pull request features zero diff updates to GCHP run directories. Updates include:
- Remove deletion of gcchem* files from example run scripts. Unlike in previous version, MAPL now allows overwriting checkpoint files.  Previously MAPL would crash if gcchem_internal_restart was present.
- Remove ExtData debug level and setting it from runConfig.sh. ExtData debug prints are now configured from new file `logging.yml`.
- Update and clean up Harvard example run scripts following system changes on the Harvard cluster
- Change ifort18 to ifort19 in example slurm script since GCHP 13.1 no longer supports the ifort18 compiler.
- Change new logging filename from `logging.yaml` to `logging.yml` (IMPORTANT: this will require an update to `GCHPctm.F90` in the GCHP repository to use the new filename)


LiamBindle:
Good to know about the new behaviour of overwriting `gcchem*` files.

LiamBindle (approved):
Thanks for tackling these

msulprizio (approved):

PR #689, opened 2021-04-11
[PULL REQUEST] More stable stratospheric H2O boundary condition

sdeastham:
By selecting the "static" H2O option in input.geos, the user can now request that stratospheric water vapor be specified up to 70 hPa in the tropics. This results in a more stable water vapor signal in the stratosphere. If this option is not selected, the water vapor in the stratosphere may grow continuously over multi-year simulations.

The change significantly reduces the accumulation when running a full-chemistry simulation at C24 in GCHP (see plot; old approach on the right, new approach on the left).

image

### DIFF
--- a/GeosCore/input_mod.F90
+++ b/GeosCore/input_mod.F90
@@ -1961,6 +1961,14 @@ CONTAINS
     ENDIF
     READ( SUBSTRS(1:N), * ) Input_Opt%LSETH2O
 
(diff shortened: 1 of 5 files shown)


msulprizio (approved):

msulprizio:
NOTE: I moved the switch for this option to the Chemistry Menu in input.geos since the Emissions Menu was removed in 13.0.0.

PR #728, opened 2021-05-18
[PULL REQUEST] Changes required for CESM2-GC

fritzt:
Hello!

This PR fixes some of the remaining issues that I had for CESM2-GC.

1. It adds a `run/CESM` folder containing all the input files required for CESM2-GC (`HEMCO_Config.rc`, ...)
2. It removes a lot of older CPP flags that I introduced in prior commits (`#ifdef MODEL_CESM`)
3. It modifies the `Accept_External_Pedge` routine to include a `State_Grid` argument. In CESM, different chunks can have different number of columns. The `EXTERNAL_PEDGE` variable is initialized with the maximum number of columns (`pcols`), but the dimensions of `State_Met%PEDGE` and `State_Grid` are chunk-dependent and are only defined with `ncol` columns.

Let me know if you need anything else!

Regards,
Thibaud Fritz


msulprizio:
Thanks @fritzt. We can include these changes in the next benchmark version, GEOS-Chem 13.2.0. We're wrapping up 13.1.0 now.

lizziel:
Thanks @fritzt! This passed all integration and diff tests and is ready to merge. This is a zero diff update.

msulprizio:
Hi @fritzt. I found an issue with your changes in commit https://github.com/geoschem/geos-chem/commit/1abb448c5a9545a75b6be2e20bfb5290698e2204. Changing `'NO'` to `NO` in species_database.yml causes issues for gcpy. When the quotes are removed, gcpy cannot find the properties for species NO and therefore can't compute any values that depend on those properties (e.g. emissions totals). For GEOS-Chem benchmarking, I need to change this string back to `'NO'`. Will this cause issues with CESM2-GC?

fritzt:
Hi @msulprizio! Feel free to revert this! This is not affecting CESM2-GC.

msulprizio:
Will do. Thanks for the confirmation!

Issue #760, opened 2021-06-24
[FEATURE REQUEST] Update download_data.py script to fetch GCAP2 data from Rochester


yantosca:
The GCAP 2.0 met data is being stored on a server at U. Rochester (see PR #670).  We have to modify the dry-run data download script so that the appropriate met data for GCAP 2.0 simulations are pulled from the FTP site at Rochester.

yantosca:
Pull request #764 has been issued to implement this feature.

NOTE: I made the PR to the main branch but this can be changed, most likely to a patch branch.

yantosca:
This pull request has been merged into the `patch/13.1.2` branch, which will shortly be released as GEOS-Chem patch version 13.1.2.

PR #764, opened 2021-06-29
[PULL REQUEST] Update download_data.py to grab files from U. Rochester server

yantosca:
The GCAP 2.0 met field and related data are stored at the University of Rochester server.  We have modified the Python script
run/shared/download_data.py to download data from this server.

Calling the download_data.py script with:
```
 ./download_data.py log.dryrun -ur
```
will start the download process from University of Rochester:

CAVEAT: There may be some missing data files at the U. Rochester server but these can be added.

This PR addresses the Feature Request #760.

Signed-off-by: Bob Yantosca <yantosca@seas.harvard.edu>

### DIFF
--- a/run/shared/download_data.py
+++ b/run/shared/download_data.py
@@ -135,8 +135,12 @@ def extract_pathnames_from_log(dryrun_log):
     # Close file and return
     # The "sorted" command will return unique values
     f.close()
-    return {"comments": comments, "found": found,
-            "missing": missing, "local_prefix": local_prefix}
+    return {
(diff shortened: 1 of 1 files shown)


yantosca:
A few more notes:
  1. The download_data.py script was the only script that was touched, so no need for integration tests.
  2. I also tested a normal "MERRA-2" data download from AWS, which worked properly.
  3. I didn't test with ComputeCanada download, as the data will be migrated to Wash U. very soon.

msulprizio (approved):

PR #769, opened 2021-07-01
[PULL REQUEST] Trace metals simulation -- brought up to GEOS-Chem 13

yantosca:
This PR supersedes PR #511 (by @junweilovexu).  The modifications from #511 have been manually brought into GEOS-Chem 13.  The merge had to be done manually because the GEOS-Chem Unit Tester (where the run directory generation was done) has now been retired.

NOTES:
  1. Only the Canadian and NEI2011 surface inventories are read by default.  The HEMCO_Config.rc file has entries for the NEI2011 point data, but not all of the data files were present on the FTP server http://rain.ucis.dal.ca/ctm/ForGCST.  We will leave these commented out for the time being.
  2. The trace metals simulation has been made into its own specialty simulation.  This is to avoid the hourly data I/O from impacting other simulation types.  Also, there are only 3 years of data (2011, 2012, 2013) available, whereas most other simulations are using more recent years.
  3. Species names have been updated (e.g. from `Al_F1`, `Al_F2`, ... to `AlF1`, `AlF2, ...).  This should avoid issues with applying wild cards in HISTORY diagnostics.
   
Emissions data are included in the `HEMCO/TRACE_METALS/v2021-06` directory tree.


msulprizio (approved):
Looks good. I would only suggest changing the fullname descriptions for the new trace metals in species_database.yml to use the full element names instead of abbreviations.

PR #776, opened 2021-07-08
[PULL REQUEST] Bug fix: add carbon to species scale factors to AEIC VOC emissions.

christophkeller:
The AEIC VOC emissions are still being emitted in units of kgC instead of kg species. This bug fix adds the carbon to species scale factors to all AEIC VOC emissions. This requires the definition of two new scale factors, CtoMACR and CtoRCHO. I assigned scale factor IDs 83 and 84 to them.
This bug fix results in an increase in aircraft emissions of ACET, ALD2, ALK4, C2H6, C3H8, PRPE, MACR, and RCHO by up to a factor of 1.6. The resulting changes in modeled trace gas concentrations are small. In a test simulation using GEOS with GEOS-Chem at c90 resolution, changes in species concentrations were less than 1% after one month.

### DIFF
--- a/run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.fullchem
+++ b/run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.fullchem
@@ -3985,8 +3985,8 @@ Warnings:                    1
 43 CtoBENZ MATH:78.12/(6.0*12.0)   - - - xy unitless 1
 44 CtoC2H4 MATH:28.05/(2.0*12.0)   - - - xy unitless 1
 45 CtoC2H6 MATH:30.08/(2.0*12.0)   - - - xy unitless 1
-46 CtoC3H8 MATH:46.08/(3.0*12.0)   - - - xy unitless 1
-47 CtoEOH  MATH:46.06/(2.0*12.0)   - - - xy unitless 1
+46 CtoC3H8 MATH:44.11/(3.0*12.0)   - - - xy unitless 1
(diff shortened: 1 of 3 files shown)


yantosca:
I think it should probably be OK to change EOH to 46.07.  Going to consult with @msulprizio to see if that would have any side-effects.

yantosca:
Going to merge this up to dev.  Changed the MW of EOH to 46.07 as per the reference.

yantosca (approved):
Updated the scale factors to use proper MW's for C3H8 and EOH.  Changed EOH MW to 46.07 in the species database.

flavioquadros:
Hi, I am a bit unsure if the carbon correction was needed, could you please check this issue again? I was under the impression that the HC variable in AEIC was in CH4-equivalent grams, as mentioned in the papers describing the inventory (https://doi.org/10.1016/j.trd.2013.07.001, https://doi.org/10.1016/j.atmosenv.2011.07.012). The speciation factors seem to be those adopted by the EPA (https://www.epa.gov/regulations-emissions-vehicles-and-engines/organic-gas-speciation-profile-aircraft), which are also implemented by the FAA's emissions model (https://aedt.faa.gov/Documents/AEDT3d_UserManual.pdf, item 5.1.1.8).

In GEOS-Chem, we have factor `114` (=1.16e-3) which does (g)->(kg) and the conversion of THC (total hydrocarbons, base CH4) to TOG (total organic gases, in overall mass). And then factors `101` ... `109` do the speciation on that as per the EPA profile. So, unless I am missing something, weren't the resulting numbers prior to the patch already in total kg of each species?

PR #780, opened 2021-07-15
[PULL REQUEST] Meng offline dust

YanshunLi-washu:
Hi,

In this version. The offline dust emissions are updated from v2019-01 to v2020-05. The scaling factors for make the total annual dust emission to 2000 Tg were updated as:

GEOS-FP: 5.7141e-04
(North America dust source optimizing (reducing 30% emission) by applying scale factor 666, which includes a mask file of "DustEmis_scalefactor_NA_optimized.nc". See the example "HEMCO_Config_w_NA_optimizing.rc")

MERRA2: 4.7376e-04
(North America dust source optimizing (reducing 30% emission) by applying scale factor 666, which includes a mask file of "DustEmis_scalefactor_NA_optimized.nc". See the example "HEMCO_Config_w_NA_optimizing.rc")

Thanks,
Yanshun

### DIFF
--- a/run/CESM/HEMCO_Config.rc
+++ b/run/CESM/HEMCO_Config.rc
@@ -106,7 +106,7 @@ Warnings:                    1
 #   OFFLINE_SEASALT     - SeaSalt
 #   OFFLINE_SOILNOX     - SoilNOx
 #--------------------------------------------------------------
-    --> OFFLINE_DUST           :       false    # 1980-2021
+    --> OFFLINE_DUST           :       true     # 1980-2021
     --> OFFLINE_BIOGENICVOC    :       false    # 1980-2021
     --> OFFLINE_SEASALT        :       false    # 1980-2021
     --> OFFLINE_SOILNOX        :       true     # 1980-2021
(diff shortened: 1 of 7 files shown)


msulprizio (commented):
I'm concerned that there are many files listed here as "Empty file." Why is that and will merging that in remove the contents of those files?

In addition, I thought we had decided at a GCST meeting with Daniel and Randall that we would no longer apply scaling factors to the offline emissions in HEMCO_Config.rc but instead apply them in the preprocessing stage so they're already included in the data files read in by HEMCO. I'll follow up with an email to Randall and others to check on this.

YanshunLi-washu:
Thanks. I'll check on the "Empty file" issue ASAP. For the scaling factor issue, I followed Jun Meng's development kit for the offline dust emissions. Yes, thanks for following up with Randall. I think Jun will also need to know this. Because currently the v2020-05 data is still using scaling factors. If we like to using no scaling factors, then there could be another new version.

YanshunLi-washu:
@msulprizio Hi Melissa, those empty files seem to related to permission-related issues when I did git clone on my local system. It was fixed now. Could you please take a look and let me know if it's all good? Thanks.

msulprizio (approved):
The empty files are now gone, so I now approve this PR.

msulprizio:
> In addition, I thought we had decided at a GCST meeting with Daniel and Randall that we would no longer apply scaling factors to the offline emissions in HEMCO_Config.rc but instead apply them in the preprocessing stage so they're already included in the data files read in by HEMCO. I'll follow up with an email to Randall and others to check on this.

Randall Martin wrote:

>Thanks for catching this.  Yes, I agree that the dust emissions should be scaled by the scale factors so that a direct read of the offline emissions will yield the desired emission rate. 

@YanshunLi-washu will work on reprocessing the offline dust emissions to include the scale factors. I will close this pull request for now, and a new pull request can be opened for the updated dataset.

PR #789, opened 2021-07-22
GCHP Harvard run script updates

lizziel:
This PR makes bug fixes and improvements to the run scripts stored in `run/GCHP/runScriptSamples/operational_examples/harvard_gcst`.

- Fix incompatibilities of `gchp.multirun.*` scripts with 13.1.2
- Customize `gchp.multirun.*` scripts for GCHP benchmarking
- Add error handling in `gchp.local.run` for case of mismatch of total cores in `gchp.local.run` and `runConfig.sh`
- Remove lower memory suggestion for Transport Tracers (leave up to the user instead)

### DIFF
--- a/run/GCHP/createRunDir.sh
+++ b/run/GCHP/createRunDir.sh
@@ -435,9 +435,9 @@ sed -i -e "s|{DATE2}|${enddate}|"       ${rundir}/CAP.rc
 
 # Special handling for benchmark simulation
 if [[ ${sim_extra_option} = "benchmark" || ${sim_name} == "TransportTracers" ]]; then
-    total_cores=48
+    total_cores=96
     num_nodes=2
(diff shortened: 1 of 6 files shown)


lizziel:
I kept the scripts to use 48 cores over 2 nodes for benchmarking. I can change this before merging if we want to instead go to 96 cores over 2 nodes.

msulprizio (approved):

msulprizio:
Looks good to me. I think 96 cores over 2 nodes would be good.

lizziel:
Fyi, this update will break integration tests since the test package has expected settings in runConfig.sh hard-coded. I am applying a fix for that in the next merge for 13.1.3, along with some other GCHP integration test updates.

PR #797, opened 2021-07-27
Updating adjoint code with bug fixes and accounting for structural changes

TerribleNews:
There were a few bugs in the previously checked in adjoint code and there were some
big structural changes to gchp_chunk_mod that meant the adjoint code was being called
at the wrong point in the run subroutine so I had to move some of the code around
a bit.
There were also changes to the run directory that needed to be incorporated into the adjoint run directory files.

### DIFF
--- a/Interfaces/GCHP/gchp_chunk_mod.F90
+++ b/Interfaces/GCHP/gchp_chunk_mod.F90
@@ -285,7 +285,7 @@ CONTAINS
     ENDIF
 
 
-    if (.not. FD_STEP == -1 .and. input_opt%IS_ADJOINT) THEN
+    if (.not. FD_STEP == -1 .or. input_opt%IS_ADJOINT) THEN
        Input_Opt%FD_STEP = FD_STEP
 
        call ESMF_ConfigGetAttribute(CF, FD_SPEC, &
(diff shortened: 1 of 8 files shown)


Interfaces/GCHP/gchp_chunk_mod.F90
```diff
@@ -1504,16 +1505,6 @@ SUBROUTINE GCHP_Chunk_Run( GC,                                             &
...
 
-    CALL MAPL_TimerOff( STATE, 'GC_DIAGN' )
-    if(Input_Opt%AmIRoot.and.NCALLS<10) write(*,*) ' --- Diagnostics done!'
-
-    !=======================================================================
-    ! Convert State_Chm%Species units
-    !=======================================================================
-    CALL Convert_Spc_Units ( Input_Opt, State_Chm, State_Grid, State_Met, &
-                             OrigUnit, RC )
-    _ASSERT(RC==GC_SUCCESS, 'Error calling CONVERT_SPC_UNITS')
```
LiamBindle, line 1515: Could you explain what this change is? Is it supposed to be outside of an `#ifdef ADJOINT` block?

Interfaces/GCHP/gchp_chunk_mod.F90
```diff
@@ -1552,6 +1545,15 @@ SUBROUTINE GCHP_Chunk_Run( GC,                                             &
...
 
+    CALL MAPL_TimerOff( STATE, 'GC_DIAGN' )
+    if(Input_Opt%AmIRoot.and.NCALLS<10) write(*,*) ' --- Diagnostics done!'
+
+    !=======================================================================
+    ! Convert State_Chm%Species units
+    !=======================================================================
+    CALL Convert_Spc_Units ( Input_Opt, State_Chm, State_Grid, State_Met, &
+                             OrigUnit, RC )
+    _ASSERT(RC==GC_SUCCESS, 'Error calling CONVERT_SPC_UNITS')
```
LiamBindle, line 1556: Oops, I see it here. The diff showed up weird.

LiamBindle (commented):
Looks good to me.

Interfaces/GCHP/gchp_chunk_mod.F90
LiamBindle: See next comment.

PR #814, opened 2021-08-07
[PULL REQUEST] CH3O2 + OH reaction

msl3v:
Added MO2 + OH reaction per DJJ request.


yantosca (commented):
Looks straighforward.  Are there FAST-JX config file updates too that go with this?  I see EHTP is a new photolyzed species so we probably need to update FJX_j2j.dat etc.

yantosca:
This reaction is now included in pull request #904.  We can close this PR now.

PR #832, opened 2021-08-23
updated download scripts to include wustl extdata mirror

LiamBindle:


msulprizio (commented):
@LiamBindle These changes look fine to me. However, I would suggest changing the base branch to [`patch/13.2.1`](https://github.com/geoschem/geos-chem/tree/patch/13.2.1) since @yantosca made changes there to read input for download_data.py from a YML file. See commit https://github.com/geoschem/geos-chem/commit/fa5919c0c36862d88ec995cea87f126323816141.

LiamBindle:
@msulprizio Thanks for point that out. It looks like @yantosca already updated `download_data.py`. I opened #832 which is this PR rebased on 13.2.1.

PR #835, opened 2021-08-23
Minor GCHP run directory updates

lizziel:
This PR is for small updates for GCHP related to run directories. So far it includes expanding clarifying comments in example GCHP SLURM run scripts (discussed in https://github.com/geoschem/GCHP/issues/129) and ensuring GCHP option to write restart files via o-server is disabled if using less than 1000 cores (discussed in https://github.com/geoschem/GCHP/issues/133). This PR is for merge into GEOS-Chem 13.2.1.


LiamBindle (approved):
Thanks for adding this @lizziel

PR #846, opened 2021-08-26
[WIP] Logging updates for GCHP

LiamBindle:
This is a work in progress (WIP).

### 1. Improve error reporting for common pitfalls

The reporting of errors related to reading config files could be improved. The config files are parsed in `Chem_GridCompMod` so that's the module I'm focusing on. Two updates:

1. Replace `__RC__` calls with `rc=STATUS` followed by `_ASSERT_RC()` followed by `_VERIFY()`. 
2. Report what file is about to be parsed before it is actually read. 

#### Details on 1.1.

`__RC__` is a PP macro defined by 
 
```
#define __RC__  RC=STATUS); _VERIFY(STATUS
```

in [MAPL/include/MAPL_Exceptions.h](https://github.com/GEOS-ESM/MAPL/blob/72d5d4edbe67e548ffd050e3c7984294873ce3fb/include/MAPL_Exceptions.h#L126). In places like loading config files (e.g., [here](https://github.com/geoschem/geos-chem/blob/3d82b3a6ad2dcb0843e752616f7f761820a1a1aa/Interfaces/GCHP/Chem_GridCompMod.F90#L435)) these \_\_RC\_\_ calls mean that errors aren't reported very well. For example, if parsing GCHP.rc fails the following *non-fatal* error is reported

```
pe=00001 FAIL at line=00573    Chem_GridCompMod.F90                     <status=8>
```

and there is no information about the offending file. Note that the staus is actually helpful, `status=8` refers to `MAPL_STRING_TOO_SHORT` defined [here](https://github.com/GEOS-ESM/MAPL/blob/b101095c1a1b780b4be5382aeb927d0c999f524f/shared/MAPL_ErrorHandling.F90#L43). We just need it to be printed.

The call we want to make is

```
MAPL_Assert_return_code(COND, return_code, filename, line, rc)
```

where `condition=(STATUS ==MAPL_SUCCESS)`, `return_code=STATUS`, and `rc=MAPL_ERROR_PARSING_CONFIG_FILE`.

Soon there will be the `_ASSERT_RC()`  PP macro defined by

```
#  ifdef I_AM_MAIN
#    define __return call MAPL_abort()
#    define __rc(rc) 
#  else
#    define __return return
#    define __rc(rc) ,rc
#  endif
#    define _ASSERT_MSG_AND_LOC_AND_RC(A,msg,stat,file,line,rc)  if(MAPL_Assert(A,msg,stat,file,line __rc(rc))) __return
#    define _ASSERT_MSG_AND_LOC(A,msg,stat,file,line) _ASSERT_MSG_AND_LOC_AND_RC(A,msg,stat,file,line,rc)
#    define _ASSERT_RC(A,msg,stat) _ASSERT_MSG_AND_LOC(A,msg,stat,_FILE_,__LINE__)
``` 

which is close to what we want. It would be  better to use `__FILE__` though so the full path is printed, and `_ASSERT_RC` isn't actually available yet.

The call we want to make is `_ASSERT_MSG_AND_LOC_AND_RC(STATUS==MAPL_SUCCESS, 'Error parsing GCHP.rc', __FILE__, __LINE__, MAPL_ERROR_PARSING_CONFIG_FILE)`.


LiamBindle:
When I was looking into this I came across GEOS-ESM/MAPL#1002. After that fix the errors are quite a bit better. Closing this for now.

PR #877, opened 2021-09-01
Changed default log level for MAPL to WARNING (all PEs) and INFO (root PE)

LiamBindle:


LiamBindle:
FYI @lizziel. This way the start/end of stages is only logged to `allPEs.log` by the root PE.

lizziel:
Sounds good. I definitely think we need documentation somewhere on using the logger since users should know they can configure how much info they can get out at run-time. We have [this](url). But I wonder if there should be an entire section on debugging in the User Guide menu below "Running GCHP". Sorry if there's already a github issue on this - my github notifications are out-of-control and I haven't kept on top of what's out there.

PR #894, opened 2021-09-15
Feature/integration tests: script for interactively running integration tests (for usage on aws)

laestrada:


yantosca (changes requested):
Hi @laestrada, great job with this.  One very minor point, I think the sentence:
```bash
echo "Execution   tests submitted as SLURM job ${EXE_ID}"`
```
can just be changed to "Execution tests finished".   The `$EXE_ID` variable was the SLURM job ID when running on cannon.

Otherwise we can go ahead and merge!

Issue #906, opened 2021-09-21
[BUG/ISSUE] Typo in CloudHet


viral211:
There is a typo in CloudHet in gckpp_Hetrates.F90. The line
https://github.com/geoschem/geos-chem/blob/ba58da4a6e447e47714fe5e7da488d15432a6fdc/KPP/fullchem/gckpp_HetRates.F90#L1282
should be:
      `ff = min( ff, 1.0e30_dp )`

I have run it by Chris Holmes @cdholmes. I think this would change the benchmark output significantly.

yantosca:
Thanks Viral.  Also tagging @msl3v on this.  We can try to get this into 13.3.0 and benchmark.

yantosca:
This has been fixed with commit c6adf4b2f3772caa8d2dedc8acb9990bf03a3016.  Awaiting PR review and merge.

yantosca:
This has now been merged into the 13.3.0 development branch and tagged with 13.3.0-alpha.4.

yantosca:
Hi @viral211 and @cdholmes.  I merged this PR as 13.3.0-alpha.4 but @msulprizio and I noticed that it caused fullchem sims to choke on the 2nd chemistry timestep with a could not integrate error:
```
---> DATE: 2019/07/01  UTC: 00:20  X-HRS:      0.333333
 Forced exit from Rosenbrock due to the following error:
 --> Step size too small: T + 10*H = T or H < Roundoff
 T=   962.73884987080180      and H=   5.0591750291364415E-013
 ### INTEGRATE RETURNED ERROR AT:           57          39          31
 Forced exit from Rosenbrock due to the following error:
 --> Step size too small: T + 10*H = T or H < Roundoff
 T=   966.10882298003025      and H=   5.1668987654245023E-013
 ## INTEGRATE FAILED TWICE !!! 
===============================================================================
GEOS-CHEM ERROR: Integrator error code : -7
STOP at INTEGRATE_KPP
===============================================================================
```

I am wondering if that MIN value is even needed.  In my new hetrates code I now use a SafeDiv function to set the result at 1e30 if the division cannot be done.  So we may be able to get away with this:
```F90
    ! Ratio of volume inside to outside cloud
    ! ff has a range [0,+inf], so cap it at 1e30
    ff = SafeDiv( H%CldFr, H%ClearFr, 1.0e+30_dp )
```

yantosca:
[...]

PR #912, opened 2021-09-24
[PULL REQUEST] Fix typo in CloudHet functions: Use MIN instead of MAX for ff

yantosca:
This is the corresponding PR to bug report #906.

These updates are 1-line fixes (changing MAX to MIN), but we will run "alpha" benchmarks on this update, as it has the potential to cause significant changes in the chemistry output.


msulprizio (approved):
This is a simple enough fix and looks good to me. We will evaluate with an internal 1-month benchmark.

Issue #547, opened 2020-11-20
[FEATURE REQUEST] US: NEI 2016 (P. Campbell et al)


yantosca:
This update by co-authors Patrick Campbell, Barron Henderson (@barronh), and Lyssa Freese (@lfreese) is listed as almost ready for inclusion into GEOS-Chem.

This feature request will track the progress of this item's inclusion into GEOS-Chem.

lfreese:
@yantosca I have the benchmark analysis that @barronh and I completed for the new files if that is needed for this incorporation. I think we were also hoping to find a way to make the methods publicly accessible (with the hopes of it being used again for future inventories), but Barron may want to confirm that given that he did the majority of the work on the notebook to create the files.

barronh:
I confirm that this is ready, but I urge initial users to look for any issues that we may have missed.

barronh:
Lyssa Freese and I developed a Jupyter Notebook that can prepare gridded monthly NEI files for use by HEMCO. It can work in the cloud with Google Colab, or on any system that has the necessary libraries. It can be tweaked for different versions.

https://colab.research.google.com/drive/1Z7Btq_upCyCcHWaKk-NFhS9P_Rbiddne?usp=sharing

I'm not sure where all this is documented, so I wanted to tag it here.

barronh:
Also wanting to note an ftp site for the emissions.

https://gaftp.epa.gov/Air/aqmg/global/NEI2016fh/output/

msulprizio:
Hi @barronh. I'm working on bringing these emissions into GEOS-Chem 13.2.0 and have a few questions. Do you recommend using the files in `EPA_Config.rc` or `EPAMERGED_Config.rc`? Do you have a README file for the files in NEI2016fh/output/? 

Note: I was able to greatly reduce file size by chunking and compressing the files in NEI2016fh/output/ using: 
`nccopy -c lon/900,lat/400,lev/1,time/1 -d1 2016fh_xxx.ncf tmp.nc; mv tmp.nc 2016fh_xxx.ncf`

barronh:
I can work on a README. The merged files are useful for faster reading, but
have limited utility for sectoral applications. I can address this in a
README. I'm glad that the chunking reduced the files so greatly. I though
we were already chunking, so I'll have to check that.

-Barron

msulprizio:
[...]

PR #920, opened 2021-09-28
[PULL REQUEST] NEI2016 emissions

msulprizio:
This PR corresponds to Feature Request https://github.com/geoschem/geos-chem/issues/547.

This inventory will be optional (off by default) and CEDS emissions (1750-2019) will remain the default anthropogenic emissions inventory globally.


yantosca (approved):
This all looks straightforward.  Will have to test with NEI emissions turned on but it looks like a straight-up swap of NEI2016 for NEI2011.

Question: why does Ordonez_Ionocarb now use a hierarchy of 1 instead of 50?  Is it because the EPA2016 does not have the same emissions?  I am sure it is OK but just wanted to know.

msulprizio:
>Question: why does Ordonez_Ionocarb now use a hierarchy of 1 instead of 50? Is it because the EPA2016 does not have the same emissions?

This appeared to be a typo in the original HEMCO_Config.rc and/or isn't needed anymore. This only applies to the `ORDONEZ_CH3I` entry and caused issues saving out the `InvORDONEX_CH3I` diagnostic in HEMCO_Diagn.rc. The remaining `ORDONEZ_IODOCARB` entries have always used hierarchy of 1. Changing the hierarchy value to 1 for `ORDONEZ_CH3I` helps us save out all species for this inventory in HEMCO_Diagn.rc consistently.

I believe the hierarchy of `ORDONEZ_CH3I` was 50 before to avoid conflicts with the `CH3I_SEAWATER` entry which also uses Cat=1 and Hier=1. The `CH3I_SEAWATER` entry is commented out so there should not be conflicts with the `ORDONEZ_IODOCARB` entry now.

PR #719, opened 2021-05-04
[PULL REQUEST]Updates to IONO2 hydrolysis as described in Wang et al 2021

tsherwen:
Hello GCST, 

Please can I request that this minor commit be included in the timescales of the updated Cl+Br+I paper [Wang et al 2021]? the Wang et al [2021] paper is to be submitted to ACPD very very soon. This is a scientific change with small impacts. It includes IONO2 hydrolysis in a way that is more consistent with other XNOy species. A description of the changes will be included in Wang et al [2021]. 

Screenshot 2021-05-04 at 17 06 03

I can write an accompanying issue for this pull request too. 

Cheers, 

Tomas

PR #923, opened 2021-09-28
[PULL REQUEST] Update IONO2 hydrolysis following Wang et al 2021 (supersedes PR #719)

yantosca:
This PR implements the update to IONO2 hydrolysis originally implemented in PR #719.  

The code in the original PR no longer exists in the current GEOS-Chem (after implementation of the new hetchem rate functions), so I had to take the algorithms from the original PR and manually apply them into the new code base.

This PR now supersedes #719.  Please see the original PR for the discussion thread.


yantosca:
Integration tests pending

Issue #941, opened 2021-10-05
[BUG/ISSUE] Incorrect indexing for State_Chm%DryDepVel and related drydep quantities


yantosca:
@lizziel found an error in the registering of the `State_Chm%DryDepVel` field.  

The list of CHEM_DryDepVel species in the log looks wrong
```
 CHEM_SURFACEFLUX_ALK4          | Surface flux (E-D) f | xy    | kg m-2 s-1
 CHEM_SURFACEFLUX_ALD2          | Surface flux (E-D) f | xy    | kg m-2 s-1
 CHEM_SURFACEFLUX_AERI          | Surface flux (E-D) f | xy    | kg m-2 s-1
 CHEM_SURFACEFLUX_ACTA          | Surface flux (E-D) f | xy    | kg m-2 s-1
 CHEM_SURFACEFLUX_ACET          | Surface flux (E-D) f | xy    | kg m-2 s-1
 CHEM_DRYDEPVEL_MPAN            | Dry deposition veloc | xy    | m s-1
 CHEM_DRYDEPVEL_MP              | Dry deposition veloc | xy    | m s-1
 CHEM_DRYDEPVEL_MONITU          | Dry deposition veloc | xy    | m s-1
 CHEM_DRYDEPVEL_MONITS          | Dry deposition veloc | xy    | m s-1
 CHEM_DRYDEPVEL_MONITA          | Dry deposition veloc | xy    | m s-1
 CHEM_DRYDEPVEL_MOH             | Dry deposition veloc | xy    | m s-1
 CHEM_DRYDEPVEL_MGLY            | Dry deposition veloc | xy    | m s-1
```
The drydep species start at MPAN and then go alphabetically.   Some of those are not drydep species, e.g. `MP`.  

We think that this is related to incorrect species indexing, and/or the omission of the wild card for drydep species.  We are working on a fix.

NOTE: For dry-deposition diagnostics, we typically use the `State_Diag%DryDepVel`` diagnostic instead, so this should not affect benchmark results.

msulprizio:
This has been resolved by PR https://github.com/geoschem/geos-chem/pull/942 and is now in the `dev` branch for inclusion in 13.3.0.

PR #942, opened 2021-10-05
Fix registering species-based quantities in state_chm_mod.F90

yantosca:
This commit fixes the issue described in geoschem/geoschem #941.

When registering species-based quantities, there was an indexing
error for quantities with less than the total number of species
(i.e. drydep and wetdep quantities).

We have updated the algorithm in the routine Get_Diagnostic_Name
to compute the modelId from the drydep or wetdep ID, and then use that
to look up the name of the species.

Signed-off-by: Bob Yantosca <yantosca@seas.harvard.edu>


yantosca:
Integration tests are currently pending.

msulprizio (approved):

PR #170, opened 2019-12-11
[PULL REQUEST] Oceanic ozone dry deposition update

r-pound:
Model updates described in accompanying paper (https://www.atmos-chem-phys-discuss.net/acp-2019-1043/ - currently under discussion). Only the changes in drydep_mod.f are required for this update.

PR #985, opened 2021-11-02
[PULL REQUEST] Fix numerical issues that halted GCHP benchmark simulations (closes geoschem/gchp #170)

yantosca:
This PR addresses the issues brought up in https://github.com/geoschem/GCHP/issues/170 (GCHP 13.3.0-alpha.9 benchmark fails).

We have modified some error traps that should now prevent certain equations from blowing up due to very small denominators.  In particular:

  1. In routine `CHEM_SO2`, we skip computing in-cloud sulfur chemistry if liquid water content (`LWC`) is less than 1e-20.
  2. In hetchem utility routine `kIIR1Ltd`, we exit if `concEduct` (i.e. the "abundant" species) is less than 1e-8 molec/cm3.

@msulprizio, @lizziel, I have this PR going into the "release" branch so that we can use these fixes to run 13.0.0-rc.1 benchmarks.  Feel free to change the destination branch.

### DIFF
--- a/GeosCore/sulfate_mod.F90
+++ b/GeosCore/sulfate_mod.F90
@@ -2998,19 +2998,18 @@ CONTAINS
        L8      = 0.e+0_fp !(jmm, 06/26/18)
        L8S     = 0.e+0_fp !(jmm, 06/26/18)
 
-       ! If we are not using KPP to compute reaction rates, then
-       ! compute SO2 cloud chemistry (1) if there is cloud, (2) if there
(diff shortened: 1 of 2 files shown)


msulprizio (approved):

msulprizio:
I see this PR is for merging into the `release` branch. Can we also merge into `dev` so they're at the same point? Then I think we can tag with `13.3.0-rc.1`.

yantosca:
OK great, I can do both.  Am on it now.

PR #990, opened 2021-11-04
[PULL REQUEST] Change 4D State_Chm%Species array to vector of 3D concentration arrays

lizziel:
This is a no-diff structural update that changes the container in which we store species concentrations throughout GEOS-Chem. Species concentrations are currently stored in a 4D array called State_Chm%Species which is allocated with dimensions State_Grid%NX, State_Grid%NY, State_Grid%NZ, and State_Chm%nSpecies. This PR changes that to store species concentrations in one 3D array per species (derived type SpcConc). Pointers to the concentration arrays are stored in a 1D vector (State_Chm%SpeciesVec) of size State_Chm%nSpecies. This work is in progress and names are subject to change.

The motivation of this update is to make the species concentration containers the same size as those stored in MAPL when using GEOS or GEOSgcm. Having them the same size will allow (1) pointing to the concentration arrays in the internal state and (2) avoidance of allocating 3D concentration arrays to contain duplicate information. This will reduce the memory requirement for the model. The update to point to the internal state concentration arrays from the new species vector will be a separate PR.


yantosca:
Thanks @lizziel, now I understand how this is being implemented.  When we add this in, this will likely necessitate a version number change to 14.0.0, as we've broken backwards compatibility.

lizziel:
This PR is now good to go, pending review. I did integration tests for both GCHP and GC-Classic before and after the updates and get zero diffs with the exception of APM (Metrics and SpeciesConc collections) and RRTMG (RRTMG collection only) simulations. I also did two integration test runs of my ref (13.4.0-alpha.27) and compared them. Those also had diffs in the same places, indicating a parallelization issue separate from my updates.

GeosCore/hco_interface_gc_mod.F90
```diff
@@ -4816,7 +4820,6 @@ SUBROUTINE Compute_Sflx_for_Vdiff( Input_Opt,  State_Chm, State_Diag,      &
     RC      =  GC_SUCCESS
     dflx    =  0.0_fp
     eflx    =  0.0_fp
-    spc     => State_Chm%Species(:,:,1,1:State_Chm%nAdvect)
```
yantosca, line 4819: The pointer spc here is not in the new code.

GeosCore/seasalt_mod.F90
```diff
@@ -418,11 +419,11 @@ SUBROUTINE ChemSeaSalt( Input_Opt,  State_Chm, State_Diag,                 &
     !----------------------------------------
     ! Sea salt emissions for extra APM bins
     !----------------------------------------
-    SEASALTS = &
-     State_Chm%Species(:,:,:,APMIDS%id_SEABIN1:(APMIDS%id_SEABIN1+NSALTBIN-1))
-    CALL SRCSALTBIN( SEASALTS, State_Grid, State_Met )
-    State_Chm%Species(:,:,:,APMIDS%id_SEABIN1:(APMIDS%id_SEABIN1+NSALTBIN-1)) &
-      = SEASALTS
+
+    DO N = 1, NSALTBIN
```
yantosca, line 423: Is State_Chm%Species(N)%Conc now referenced in the SRCSALTBIN routine?

yantosca (approved):
This is a straightforward PR that swaps out State_Chm%Species(I,J,L,N) for State_Chm%Species(N)%Conc(I.J,L).  Good to merge.

GeosCore/carbon_mod.F90
```diff
@@ -117,7 +117,7 @@ MODULE CARBON_MOD
 !  (14) Pye. H.O.T., A.W.H Chan, M.P. Barkley, and J.H. Seinfeld, "Global
 !        modeling of organic aerosol: The importance of reactive nitrogen (NOx
 !        and NO3)", Atmos. Chem. Phys., Vol 10, 11261-11276,
-!        doi:10.5194/acp-10-11261-2010, 2010.
+!        doi:10.5194/acp-10-11261-2010, 2010.spc(
```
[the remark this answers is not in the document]
lizziel: Good catch!

GeosCore/seasalt_mod.F90
lizziel: Yes, the species IDs are passed to SRCSALTBIN and used to retrieve concentrations from State_Chm%Species. See [here](https://github.com/geoschem/geos-chem/blob/1085010733aa9fc0276d3adfdb918970895b788a/GeosCore/seasalt_mod.F90#L1670-L1672).

GeosCore/hco_interface_gc_mod.F90
lizziel: Right, I removed the spc pointer since State_Chm%Species is used directly in Compute_Sflx_for_Vdiff.

Issue #1012, opened 2021-11-15
[BUG] ExtData OFFLINE_DUST 


sdeastham:
Although the entries in HEMCO_Config.rc for OFFLINE_DUST now point to v2021-08 (since v13.3.0), it looks like the entries in ExtData.rc still point to v2019-01 (https://github.com/geoschem/geos-chem/blob/e9a4f1f9873f6e12ef3b9af7f823368a97339aa7/run/GCHP/ExtData.rc.templates/ExtData.rc.fullchem#L1687-L1693).

yantosca:
Thanks for bringing this up @sdeastham.  Wonder if there is going to be an update for dust emissions that would supersede this soon?

sdeastham:
No problem! It sounds like this is a problem with the GCHP v13.3.0/1 release; it appears that offline dust was updated in HEMCO_Config.rc, but not in ExtData.rc (https://github.com/geoschem/geos-chem/pull/816).

lizziel:
I am doing a PR on this now so we can get into 13.3.2 in time for GEOS-Chem newsletter release.

LiamBindle:
Closing. Lizzie fixed this in geos-chem#1017 and it's release in 13.3.2.

PR #1017, opened 2021-11-18
Fix outdated path in GCHP fullchem config file ExtData.rc

lizziel:
This PR addresses the bug fix reported in https://github.com/geoschem/geos-chem/issues/1012. It corrects a bug where the GCHP standard simulation was using an old offline dust inventory. The benchmark simulation is not impacted because it uses online dust.


lizziel:
I am going to do a quick test run for this now.

yantosca (approved):
This is a straightforward fix of just changing directories in ExtData.rc.  Also, this should not affect the benchmarks because we use online dust.

PR #1042, opened 2021-12-03
[PULL REQUEST] Allow negative pKa but skip missing values in routine CALC_HEFF (Closes #1001)

### LINKED ISSUE #1001: [BUG/ISSUE] Bug in henry_mod/CALC_HEFF
### Problem
The calc_heff routine calculates the effective Henry's law constant accounting for acid dissociation. calc_heff is called from wetscav_mod/compute_l2g and gckpp_hetrates/compute_l2g_local. 
In the current code (version 13.2) we compute Heff only when the pKa is greater than 0, but pKa can be negative. 
https://github.com/geoschem/geos-chem/blob/ba58da4a6e447e47714fe5e7da488d15432a6fdc/GeosUtil/henry_mod.F90#L196-L200
This condition was set to bypass the calculation when pKa is not defined, which is true for all of the species in the species database, since the Henry's law constants specified in the species database for use in wetscav_mod are already the effective ones. So calc_heff does not need to be called from wetscav_mod/compute_l2g at all. However, when calc_heff is called from gckpp_hetrates/compute_l2g_local, pKa is defined and is negative. This returns an incorrect Heff, which affects the partitioning of acids (HCl and HBr) to cloud droplets.

### Proposed solution:
(i) Remove the pKa > 0 condition in calc_heff, and (ii) remove the call to calc_heff in wetscav_mod/compute_l2g.

yantosca:
This is the corresponding PR for #1001.  It fixes an error that had caused pH to be neglected in the computation of Henry's law in cloud water (for heterogeneous chemistry).


yantosca:
A rerun of the GCHP integration tests had all tests pass.  We are waiting for a rerun of GCC integration tests to complete.

laestrada (approved):

yantosca:
The GCC fullchem integration tests now all pass.  See issue #1001 for details.

PR #1102, opened 2022-01-19
Updates from GEOS developers

lizziel:
This PR includes all changes made to GEOS-Chem by GEOS developers, including myself, since GEOS-Chem 13.1 was incorporated into GEOS. It is starting off as a draft PR since I have not yet looked at merge conflicts.

Here is brief description of changes, organized into categories:

**GEOS-only source code changes**
1. Use area normalized units for `WetLossConv` and budget diagnostics if using GEOS
2. Add new diagnostic `NOxTau` if using GEOS (still undergoing testing at GMAO)
3. Skip setting HEMCO time in `gchp_chunk_mod.F90` subroutine `GCHP_Chunk_Run` if using GEOS (whether we need it in GCHP needs to be examined)

**Standard model source code changes**
1. Move setting dry dep velocity diagnostics out of the main loop in subroutine `Update_DryDepFreq` to avoid a segmentation fault when using GEOS (and CESM) with ifort21.
2. Add new diagnostics for concentration before and after chemistry (`ConcBeforeChem` and `ConcAfterChem`)
3. Add optional argument `Vdotout` to KPP subroutine `Fun` to return species rates
4. Add support for 91 levels (currently being used for tests in GEOS)
5. Bug fix to limit `State_Chm%nDryDep` to advected species. This fixes a segmentation fault in GEOS when deallocating `State_Chm%DryDepVel` during finalize.
6. Consolidate the budget metadata section of `state_diag_mod.F90` subroutine `Get_Metadata_State_Diag`

**Run directory changes**
1. Add propane entries for QFED in GEOS run config files (whether these should be in standard model to be determined)
2. Add experimental settings in GEOS run config file `GEOSCHEMchem_GridComp.rc`
3. Remove QFED3 from GEOS `HEMCO_Config.rc` options
4. Add missing GFED extension option "GFED_subgrid_coag" to GEOS `HEMCO_Config.rc`
5. Make volcano extension settings updates for 13.3 in the GEOS `HEMCO_Config.rc`
6. Add VFEI extension to GEOS HEMCO config (still experimental in GEOS)


yantosca (changes requested):
Here are my notes on this PR:

  1. convection_mod.F90: diff only affects GEOS, this is OK
  
  2. diagnostics_mod.F90: diff only affects GEOS, this is OK
 
  3. drydep_mod.F90: Has this been shown to result in zero-diffs in GCClassic?
 
  4. fullchem_mod.F90: I think the  removal of the USE statements will cause a conflict with the latest updates for fullchem and Hg chemistry via KPP in 13.4.0.  The interface to DO_FULLCHEM  needs several state objects but the references to these types have been removed.

 5. pressure_mod.F90: Can; the DATA statements be replaced with F90-style array constructors e.g. `a91_loc(92) = (/1.0000000, 1.7500000, 2.7500000, 4.1000000, 5.9895105, ... /)`.  I think gfortran might balk at the DATA statement, which is now obsolete in F90 and later standards.

  6. diaglist_mod.F90: Lots of changes, would be good to test with GCClassic to see if it breaks backwards compatibility.

  7. species_database_mod.F90: Lizzie's fix is good.  I will have a PR (currently in prep) that will go further in preventing e.g. drydep & wetdep parameters from being assigned to non-advected species

  8. state_diag_mod.F90: Most of the updates seem to be adding new diagnostics, so shouldn't conflict with GCClassic or GCHP.  

  9. Chem_GridCompMod.F90: I am assuming this is OK, as I am not familiar with the GCHP interface.
 
  10. gchp_chunk_mod.F90: This is OK

  11. KPP/fullchem/gckpp_Function.F90: We will ideally need to add this extra text that adds VdotOut into the KPP preprocessor so that it can be generated when the mechanism is rebuilt.  Also, having VdotOut as an OPTIONAL variable will cause some extra overhead, as KPP function FUN is called for every box in the chemistry grid.  Could this be done in post-processing for the time being until we can modify KPP?

  12. Extdata.rc: OK

  13. GridComp.rc: OK

  14. run/GEOS/HEMCO_Config.rc: Looks OK, seems like the latest emissions updates are here.

  15. run/GEOS/HEMCO_DiagnFile.rc: OK

  16. run/GEOS/input.geos.rc: Now points to proper FAST_JX folder.

I think the biggest issues are the interface with the KPP code (gckpp_Function.F90 and fullchem_mod.F90), which will definitely break GEOS-Chem Classic.

yantosca:
Also wondering if there is a reason why many of the derived types for states were removed from DO_FULLCHEM?  In GEOS, are these included somewhere else?  We could block this out with e.g.

```F90
#ifndef MODEL_GEOS
    ! GEOS already includes these so skip re-including them here
    USE Species_Mod,              ONLY : Species
    USE State_Chm_Mod,            ONLY : ChmState
    USE State_Diag_Mod,           ONLY : DgnState
    USE State_Grid_Mod,           ONLY : GrdState
    USE State_Met_Mod,            ONLY : MetState
    USE Time_Mod,                 ONLY : Get_Day
    USE Time_Mod,                 ONLY : Get_Month
    USE Time_Mod,                 ONLY : Get_Year
    USE Time_Mod,                 ONLY : Get_Ts_Chem
#endif
```

yantosca:
Also tagging @christophkeller

lizziel:
The removal of those USE statements was a merge error. I fixed it in the branch that will be merged into geos/develop but haven't pushed it here yet.

christophkeller:
Good point regarding the extra overhead for VdotOut in gckpp_Function. I added this for chemical lifetime calculations (diagnostics NOxTau). This feature is still optional and I'm ok with leaving it out for now. If we remove VdotOut from gckpp_Function, we will have to create a dummy entry for it in flexchem_mod.F90 to make sure that the NoxTau diagnostics structure still works (or we would have to remove all NoxTau diagnostics references, which is more cumbersome in my opinion).

lizziel:
I'll remove Vdotout from gckpp_Function.F90 in this branch and add something to fullchem_mod.F90 to keep divergence with geos/develop minimal. This will keep the NOxTau diagnostic which is now in MODEL_GEOS brackets. Note that this PR branch will not be merged into geos/develop since we don't want 13.4.0 yet (not yet done or benchmarked).

lizziel:
Following up on earlier comments:
- I found no differences in State_Diag%DryDepVel with these updates. I only did a 1-hr test so we should verify with a 1-month benchmark run (can just look at species concentrations).
- I changed the F77-style arrays for F90 and moved them so they are local rather than module level.
- Vdotout is still used in MODEL_GEOS blocks but this will be okay in the GEOS-ESM branches in which I will make sure it is still included as an optional argument in Fun.
- The missing USE statements are back. I actually applied the above changes on a commit prior to bringing in 13.4.0 dev and re-merged, forcing a push. 
- diaglist_mod.F90 changes have not been fully tested since we don't yet have a testing suite to capture all the possible problems with HISTORY.rc and permutations of diagnostics configuration in GC-Classic and GCHP. I think if there are bugs not caught by the benchmark then we should fix them as they are encountered.

New integration tests are currently in progress.

lizziel:
All integration tests passed. @yantosca, if the changes look good to you then feel free to merge.

yantosca:
Thanks @lizziel, I'll have a look soon.

yantosca (approved):
Good to merge, we'll get this in 13.4.0-alpha.18

Issue #128, opened 2019-11-09
Potentially masked error state [1] in Config_ReadFile, hco_config_mod


kilicomu:
Wondering about your thoughts on this:

In subroutine `Config_ReadFile` from `hco_config_mod`, passing an unassociated `HcoConfig` argument is permitted - the subroutine calls `ConfigInit` on the programmer's behalf, without passing the optional argument `nModelSpecies` (which is explicitly passed when `ConfigInit` is called from `HCOI_GC_Init`, for example).

Is there ever a case wherein `Config_ReadFile` should be called with an unassociated `HcoConfig`? If not, does the implicit initialisation of the HEMCO config object mask an error state?

PR #1106, opened 2022-01-24
[PULL REQUEST] GCHP restart updates for compatibility with newer MAPL

yantosca:
**[UPDATE: This PR goes with GCHP PR #[214](https://github.com/geoschem/GCHP/pull/214)]**

This PR continues the work originally begun in #128, and merges that on top of the GEOS-Chem 13.4.0-alpha.18 tag.

This update originally failed a GCHP integration test with;
```CONSOLE
At line 256 of file /n/holyscratch01/jacob_lab/ryantosca/tests/GCHP_alpha.18/CodeDir/src/GCHP_GridComp/HEMCO_GridComp/HEMCO/src/Core/interpreter.F90
Fortran runtime error: Recursive call to nonrecursive procedure 'add_sub'
```
using the following software versions:
```
Currently Loaded Modules:
  1) git/2.17.0-fasrc01      7) zlib/1.2.11-fasrc01
  2) gmp/6.2.1-fasrc01       8) szip/2.1.1-fasrc01
  3) mpfr/4.1.0-fasrc01      9) hdf5/1.10.7-fasrc02
  4) mpc/1.2.1-fasrc01      10) netcdf/4.8.0-fasrc03
  5) gcc/10.2.0-fasrc01     11) netcdf-fortran/4.5.3-fasrc03
  6) openmpi/4.1.0-fasrc01  12) cmake/3.16.1-fasrc01
```
As @lizziel suggested, this might be related to issue: https://github.com/GEOS-ESM/MAPL/issues/589, in which a `--no-recursion` flag had to be added for GNU Fortran builds.

This issue is under investigation by @LiamBindle

### DIFF
--- a/run/GCHP/GCHP.rc.template
+++ b/run/GCHP/GCHP.rc.template
@@ -56,6 +56,10 @@ NUM_WRITERS: 1
 # ----------------------------
 BKG_FREQUENCY: 0
 
+# Enable bootstrapping restart file fields missing from restart
+#-----------------------------------------------------------------
+MAPL_ENABLE_BOOTSTRAP: YES
+
 # Settings for production of restart files
 #---------------------------------------------------------------
 # Record frequency (HHMMSS) : Frequency of restart file write
--- a/run/GCHP/GCHP.rc.template
+++ b/run/GCHP/GCHP.rc.template
@@ -71,18 +71,18 @@ RECORD_REF_DATE: 20000101
 RECORD_REF_TIME: 000000
 
 # Chemistry/AEROSOL Model Restart Files
-# Enter +none for GCHPchem_INTERNAL_RESTART_FILE not use an initial restart file
 # -------------------------------------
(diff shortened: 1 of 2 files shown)


lizziel:
I changed this PR to draft since it needs changes to avoid having to use a new restart file with all required restart variables. HEMCO restart fields and files used for post-processing or continuity across consecutive runs are not needed at start-up. At first we thought newer MAPL could no longer handle missing internal state variables (e.g. not in the restart file) but we have since learned that it can. I need to look into how to get that to work.

lizziel:
I added in a config file setting that allows restart bootstrapping in the newer version of MAPL. Since the updates in this PR are very minor I applied them on top of latest dev as new commits. I omitted the change to the restart file path since we no longer need to use a restart file with zeros added for missing fields. I then force-pushed to GitHub so that the PR history is cleaned up.

Note that this update has branch name containing MAPL 2.16 but is part of a GCHP super-project PR that upgrades MAPL to 2.18.3.

PR #1112, opened 2022-02-03
Fix occasional RRTMG compile error encountered during linking stage

lizziel:
During repeated integration testing of a development branch off of GEOS-Chem dev I noticed that building with RRTMG sometimes failed. I investigated and found that file `rrtmg_sw_k_g.F90` contained subroutines that were not wrapped in a module. It appears this caused occasional compile problems for the initialization subroutine within `rrtmg_sw_init.F90` which used the subroutines found in `rrtmg_sw_k_g.F90`.

This PR contains a fix that I believe corrects the problem. Since the failure only occurs sporadically only time will tell.


yantosca (approved):
Thanks @lizziel.  When subroutines are placed in a module the compiler automatically creates the interfaces (similar to function prototypes in C/C++) and stores them in the *.mod files).  For subroutines not in modules you have to explicitly add the interfaces.  This can cause linking issues.

The easy solution is what you did is to make sure all subroutines are in a module.

PR #1164, opened 2022-03-11
[PULL REQUEST] Updates for CESM in GEOS-Chem

lizziel:
This PR merges in the latest updates compatibility with CESM. Updates were developed by Thibaud Fritz (@fritzt), Haipeng Lin (@jimmielin), and myself. The changes include the following:

1. Set `EmisSpec` and `DepSpec` to `.False.` for GEOS-Chem in CESM.
2. Update run config files for use in CESM, including new file `geoschem_config.yml`, deletion of `input.geos`, and preliminary updates of `HEMCO_Config.rc` from Haipeng to prevent merge conflict with his PR https://github.com/geoschem/geos-chem/pull/1344. 
3. Move aerosol_mod module variable `SOILDUST` to `State_Chm` to avoid horizontal stripes in the local copies of `SOILDUST` and `DUST` in `dust_mod.F90`, possibly due to corruption. Moving `SOILDUST` to `State_Chm` resolved the issue.
4. Updates to set Input_Opt fields from new config file `geoschem_config.yml` if using CESM. Certain fields used in GEOS-Chem Classic are not used in CESM.
5. Bug fix in `input_mod.F90` for case of using GEOS and GCHP. Previously it was applied to any other model that defined `EXTERNAL_FORCING`, including CESM.


lizziel:
These updates passed all GC-Classic and GCHP integration tests as of two weeks ago, with the exception of GC-Classic CO2 which failed due to an unrelated CEDS issue that has since been fixed in the dev branch.

yantosca (approved):
This all looks good.  I approve of this PR. 

NOTE: It would be great if we can start pulling out the other fields of aerosol chemistry modules into state_chm.  It makes sense to keep these in the state variable rather than as use-associated module variables.  In due time... :-)

PR #1167, opened 2022-03-14
Add tagged O3 run directory for GCHP

xpye98:
Hi, all,

I recently put together a tagged O3 run directory for GCHP. Thanks to the good compatibility of GCC and GCHP, it only involves some changes and addition of configuration files. I have tested the code and run tagO3 simulation for several days , and the result seems to be similar to GCC's.

One thing I also did is to create a directory, `ExtData/HEMCO/GCHP_Output/13.0.0/`, to archive my daily prod/loss rates from fullchem simulation (just like the path in GCC tagO3). I think we can generate this directory and put some default prod/loss rates there for users to test.

https://github.com/geoschem/GCHP/issues/201

### DIFF
--- a/run/GCHP/ExtData.rc.templates/ExtData.rc.tagO3
+++ b/run/GCHP/ExtData.rc.templates/ExtData.rc.tagO3
@@ -1,0 +1,344 @@
+
+Ext_AllowExtrap: .true.
+#
+PrimaryExports%%
+#--------|-------|------|------------|---------------|--------|-------|---------------------|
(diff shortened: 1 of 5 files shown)


lizziel:
Thanks @xpye98! This looks good to me. There are some changes in createRunDir.sh that are going into 13.4.0 which cause conflicts, but we can take care of those.

Jourdan-He:
The input.geos file has been moved under geoschem_config.yml.templates. New PR #1380  has been created.

msulprizio:
This PR is now superseded by https://github.com/geoschem/geos-chem/pull/1380.

PR #1182, opened 2022-03-18
Extend QFED range end from 2020 to 2022

lizziel:
All HEMCO_Config.rc files that use QFED2 are now updated to use 2022 as the end year range. QFED data is not available for the future, but using an end year of 2022 does not break the model. New 2022 months will be made available as they are ready for download, approximately monthly.


yantosca (approved):
This is just a straight swap-out of 2020 for 2022 in the HEMCO_Config.rc file.  OK to merge!

PR #1189, opened 2022-03-22
[PULL REQUEST] Apply bug fix for dust aciduptake simulation from Shixian Zhai

yantosca:
The commit resolves the issue raised Shixian Zhai in Github discussion #874.  This only affects the fullchem simulation with acid uptake on dust, but will not affect the benchmark simulation.  Therefore, we can bundle this into the same alpha benchmark as the AEIC 2019 emissions.

### DIFF
--- a/GeosCore/sulfate_mod.F90
+++ b/GeosCore/sulfate_mod.F90
@@ -2753,22 +2753,28 @@ CONTAINS
 
        ! Isolate H2SO4 for reaction with dust    tdf 3/6/2K9
        IF ( LDSTUP ) THEN
-          ! Compute gas phase SO4 production again, as in offline case
-          ! RK1: SO2 + OH(g) [s-1]  (rjp, bmy, 3/23/03)
-          M    = State_Met%AIRDEN(I,J,L) * F
-          KK   = K0 * M / Ki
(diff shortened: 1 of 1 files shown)


zsx-GitHub:
@yantosca Thanks! The revision suggested in #874 requires turning on the "ProdLoss" output in HISTORY.rc, and it should be better if it can be revised in a way that "LDSTUP" can work properly regardless of "ProdLoss" being output or not.

msulprizio (approved):
Looks good to merge

yantosca:
Integration tests failed due to this compilation error:
```console
 2769 |              KK       = State_Diag%Prod(I, J, L, id_PSO4)
      |                                                        1
Error: Symbol ‘id_pso4’ at (1) has no IMPLICIT type
make[2]: *** [src/GEOS-Chem/GeosCore/CMakeFiles/GeosCore.dir/build.make:785: src/GEOS-Chem/GeosCore/CMakeFiles/GeosCore.dir/sulfate_mod.F90.o] Error 1
```

yantosca:
I've decided to merge this with PR #1190, so that we can test everything in one set of integration tests and benchmarks.  This PR does not affect fullchem benchmarks because it is only used for the acid dust uptake simulations.

PR #1198, opened 2022-03-28
[WIP] Remove non-zero diffs introduced by splitting up simulations in time

lizziel:
This PR will enable GCHP and GEOS-Chem Classic runs to be split up in time without introducing differences in output. The introduction of differences when splitting up a run is a long-standing problem in the model that will be fixed in 14.0.


lizziel:
My changes to date enable no differences in one 2-hr versus two 1-hr runs in GCHP unless megan and paranox extensions are turned on. This notably fixes the differences we have been seeing in chemistry. I will look into the problems in the HEMCO extensions soon. Fixing them may require the introduction of more restart variables, or there may need to be fixes applied in HEMCO.

I have not yet added the required restart variables to GC-Classic.

This update reorders several calls in chemistry, specifically moving the calls to `Calc_Strat_Aer`, `Aerosol_Conc`, and `RDAER` to before the first call to `ChemSulfate` in `Do_Chemistry`. More generally, the aerosol chemistry is now out of `Do_Fullchem` and is in `Do_Chemistry` instead, including the call to compute dust online. Isorropia is also moved, from after `Do_Fullchem` to before it.  These changes minimize the number of arrays that need to be included in the restart file for continuity across timesteps.

This update also removes code that skipped gravitational settling in the first timestep.

Discussion of these updates with the relevant workings will happen after I do before and after internal benchmarks to assess the differences introduced by the changes.

GeosCore/chemistry_mod.F90
```diff
@@ -102,6 +102,7 @@ SUBROUTINE DO_CHEMISTRY( Input_Opt,  State_Chm, State_Diag, &
 #ifdef TOMAS
     USE TOMAS_MOD,       ONLY : DO_TOMAS  !(win, 7/14/09)
 #endif
+
```
yantosca, line 105: This is a cosmetic change, in dev

yantosca: also the rest

GeosCore/fullchem_mod.F90
```diff
@@ -1583,6 +1494,15 @@ SUBROUTINE Do_FullChem( Input_Opt,  State_Chm, State_Diag,                 &
        CALL DEBUG_MSG( '### CHEMDR: after UCX_H2SO4PHOT' )
     ENDIF
 
+    ! Set State_Chm arrays for surface J-values used in HEMCO and
```
yantosca, line 1497: These updates (1) pull out many calls into chemistry_mod and (2) save JNO2 and JOH for subsequent use.

Interfaces/GCHP/Chem_GridCompMod.F90
```diff
@@ -409,6 +409,12 @@ SUBROUTINE SetServices( GC, RC )
     LOGICAL                       :: useCFMaskFile
 #endif
 
+    ! Manual internal state entries
```
yantosca, line 412: These updates are necessary to add the extra fields to the GCHP and/or GEOS internal state (and to the restart file).

yantosca (approved):
These updates are OK to merge.  Further modifications may be necessary, pending testing and validation.

yantosca:
This is now superseded by PR #1212.  We have merged this branch up to 13.4.0-rc.3 to take advantage of some more recent updates & fixes.

lizziel:
This PR is superseded by https://github.com/geoschem/geos-chem/pull/1229.

PR #1229, opened 2022-04-26
[PULL REQUEST] Bug fixes to remove GCHP diffs if splitting up simulations in time

lizziel:
This PR fixes several unrelated bugs that each introduced differences upon simulation restart in both GC-Classic and GCHP. Combined with a couple fixes that went into 13.4.0 (https://github.com/geoschem/HEMCO/pull/140 and https://github.com/geoschem/geos-chem/pull/1178), these updates make GCHP benchmark simulation output the same regardless of how the run is split up. 

All fixes extend to GC-Classic as well, but the analogous problem in GC-Classic is not corrected with this pull request. Updates to correct the problem in GC-Classic will be submitted as a separate pull request, possibly for a later version. For more information on the impact of this problem and the fixes, see http://ftp.as.harvard.edu/gcgrid/geos-chem/validation/GC_time_regression_fixes/.

Changes include:

1. Reorder several calls in chemistry, specifically moving the calls to `Calc_Strat_Aer`, `Aerosol_Conc`, and `RDAER` from within `Do_Fullchem` to before the first call to `ChemSulfate` in `Do_Chemistry`. More generally, the aerosol chemistry is now out of `Do_Fullchem` and is in `Do_Chemistry` instead, including the call to compute dust online. This update allows `State_Chm%AeroArea` and `State_Chm%AreaRadi` to be computed prior to their use in `ChemSulfate` in the first timestep. Prior to this change they were used in `ChemSulfate` as zero in the first timestep.

4.  Move Isorropia to earlier in chemistry, to before the call to `do_fullchem`. Several Isorropia output arrays are used in KPP but previously Isorropia was called after KPP, resulting in zero values used in the first timestep.

5. Add new restart variable `AeroH2O_SNA` which is computed in Isorropia but used earlier in subroutine `RDAER`. This fixes the problem of using a zero value in the first timestep within `RDAER`.

6. Enable wet settling in the first timestep.  Previously this was turned off for both the stratosphere in `Calc_Strat_Aer` and the troposphere in `Grav_Settling` which introduced differences upon restart.
 
7. Add `State_Chm` arrays `JOH` and `JNO2` for surface J-values of reactions `O3 + hv --> O2 + O` and `NO2 + hv --> NO + O`, and add them as new restart variables. These surface J-values are used in the HEMCO `paranox` extension which is called before J-values are calculated. This fixes the problem of using zero values for surface J-values in `paranox` in the first timestep.

8. Move `carbon_mod` module variable `ORVC_SESQ` to `State_Chm` array `ORVCsesq`, and store the array as a restart variable for both GCHP and GC-Classic. This array is the sesquiterpenes mass per grid box and is used in SOA chemistry prior to it being set. Using a zero value in the first timestep caused differences in `ASOA*`, `ASOG*`, `TSOA*`, and `TSOG*` concentrations.

9. Set the `TotalOC` diagnostic to zero before it is set every timestep to correct a bug resulting in a cumulative sum across timesteps.

10. Compute cos SZA before emissions are run. This is done in GEOS but previously was not done in GCHP. I did not isolate the differences caused by this update and it is possible it did not have an impact.


lizziel:
@theloniuspunk, I am pinging you as a member of the Aerosol Working Group since this update has a minor impact on aerosols. Besides fixing several problems of using zero values in the first timestep, it rearranges a few of the aerosol subroutines. Please let me know if you see any issues or have any questions.

theloniuspunk:
Thanks! Do you think it might affect TOMAS simulations (either in Classic or GCHP)?

lizziel:
None of the problems were in TOMAS, although I limited the testing to the benchmark simulation so if there are additional sources of restart differences in TOMAS I would not have found them. For the update I did make all full chemistry simulations will have slight differences due to the rearrangement of the aerosol subroutines.

msulprizio (approved):
These changes are consistent with the detailed description for this PR.

PR #1230, opened 2022-04-26
Added native versions of GEOS-FP as menu options in GCHP run directory creation

LiamBindle:
This is a WIP. I'll add more in the coming days.

Things I need to check
- [x] fullchem + native GEOS-FP 
- [x] fullchem + native GEOS-FP + mass fluxes
- [x] fullchem + normal GEOS-FP
- [x] fullchem + normal MERRA2


LiamBindle:
This is ready to merge @msulprizio. 

This adds "GEOS-FP native data" as an option to "Choose meteorology source:". If you select it, it also asks 
```
Do you want to use mass fluxes for advection? (yes/no, default=no):
Do you want to use mass fluxes derived winds for advection? (yes/no, default=no):
```

Based on your response, it will initialize the run directory appropriately.

msulprizio (approved):
Looks good to me.

@lizziel Will this cause any conflicts with https://github.com/geoschem/geos-chem/pull/1259? Any preference on the order of merging this PR and that one? Both can probably go into 14.0.0-alpha.5.

LiamBindle:
Sorry about that @msulprizio. I forgot to add an update to the copy of `init_rd.sh`. One of the updates was removing the `sort` command in `./createRunDir.sh` so that multi-line template values can be used. After fixing that, I found `RUNDIR_MET_DIR` and `RUNDIR_GRID_DIR` were being defined in the wrong order but `sort` was happening to reorder them "correctly". Now that I removed the `sort`, I had to include the shared met settings (e.g., `geosfp.txt`) after the grid definition section. 

I don't have the capability to run integration tests. Are you guys able to run that easily?

msulprizio:
@LiamBindle Thanks for pushing those fixes. I tested on my end and this looks good to merge now.

Issue #1244, opened 2022-05-11
[BUG/ISSUE] Incorrect meteorology written to GC-Classic restart


lizziel:
### What institution are you from?
Harvard University (GCST)

### Description of the problem
The output restart values for TMPU1, SPHU1, PS1_WET, and PS1_DRY are incorrect if written at a time divisible by 3 hours, including midnight. The values saved correspond to 3 hrs previously, e.g. SPHU1 for 21:00 through 23:50 is written for restart file at 00:00. This introduces error if splitting up a GC-Classic into multiple segments, including in the 1-year benchmark simulations of GC-Classic.

### Description of troubleshooting performed
I diagnosed this problem by running one GC-Classic run continuously for 6 hrs and another split into two 3 hr runs. I printed TMPU1, SPHU1, PS1_WET, and PS1_DRY at each timestep. I compared the values from the run logs and saw that the restart values used in the second part of the 2-segment run correspond to the same values used in the first part of the 1-segment run. These values are supposed to change every 3 hours, where TMPU1 becomes TMPU2, etc and a new value for TMPU1 et al. is read in from file.

### GEOS-Chem version
14.0.0-alpha.0. However, this problem has been present for some time.

### Description of modifications
I am looking into a fix now. If we don't need these variables in the restart I will remove them. If we do need them then TMPU2 etc should be saved to TMPU1 if the restart is output at a time divisible by 3 hours. If it is not divisible by 3 hrs then outputting TMPU1 would be okay since it should stay the same in the next timestep when the restart values would be used.

### Related issues
https://github.com/geoschem/geos-chem/issues/1225

msulprizio:
This fix has now been merged into `dev` and will be included in 14.0.0.

PR #1250, opened 2022-05-12
Remove TMPU1, SPHU1, PS1_WET, and PS1_DRY from GC-Classic restart file

lizziel:
This update fixes a bug in which the output restart values for TMPU1, SPHU1, PS1_WET, and PS1_DRY were incorrect if written at a time divisible by 3 hours, including midnight, thereby introducing error if starting a run from the output restart file of a previous run. The values saved corresponded to 3 hrs previously, e.g. SPHU1 for 21:00 through 23:50 was written for restart file at 00:00. Since GEOS-Chem Classic now requires met-fields to have the exact time of the run and the fields are therefore away available via read, these fields are now removed from the restart file rather than fix the values written.

This update has no impact on GCHP. It will cause small differences in the GC-Classic 1-month benchmark because the `*1` met-fields in the restart file will not be used and the true meteorology at the start time will be used instead.

Closes https://github.com/geoschem/geos-chem/issues/1244


msulprizio (changes requested):
The changes so far are straightforward and simply remove the four fields from being read from the GEOS-Chem restart file.

I'm wondering, should we also remove saving out these fields to the Restart collection in HISTORY.rc? For example:

https://github.com/geoschem/geos-chem/blob/9a6cb65b2f50a99441405be655cc83ae10703f55/run/GCClassic/HISTORY.rc.templates/HISTORY.rc.fullchem#L80-L83

msulprizio (approved):
Great. Now approved!

PR #1255, opened 2022-05-17
Use modified EDGAR diurnal scale factors file to correctly read data

lizziel:
This PR updates the GC-Classic and GCHP config files to use new file `EDGARv42/v2015-02/NO/EDGAR_hourly_NOxScale.modified.nc`, which is a duplicate of the file currently used but with all missing data removed and replaced with value 1.0.

The currently used file, `EDGARv42/v2015-02/NO/EDGAR_hourly_NOxScale.nc`, uses missing values and `_FillValue` attribute value of 1 to indicate where scale factors are 1.0. This causes a silent bug in GC-Classic where near-zero values are used instead of 1.0 since HEMCO treats all missing values as 0 rather than using the _FillValue (see [here](https://github.com/geoschem/HEMCO/blob/4a66bae48f33e6dc22cda5ec9d4633192dee2f73/src/Core/hco_error_mod.F90#L94-L97), [here](https://github.com/geoschem/HEMCO/blob/4a66bae48f33e6dc22cda5ec9d4633192dee2f73/src/Core/hcoio_read_std_mod.F90#L766-L782), [here](https://github.com/geoschem/HEMCO/blob/4a66bae48f33e6dc22cda5ec9d4633192dee2f73/src/Shared/NcdfUtil/hco_ncdf_mod.F90#L646-L648), and [here](https://github.com/geoschem/HEMCO/blob/4a66bae48f33e6dc22cda5ec9d4633192dee2f73/src/Shared/NcdfUtil/hco_ncdf_mod.F90#L1185-L1201).

This also causes a run fail in GCHP if a new `_FillValue` update is enabled in the new version of MAPL we just updated to. This is because data with values equivalent to `_FillValue` are now treated as missing and not included in regridding. For this reason that fix is currently disabled MAPL inGCHP. We will enable it following this update (https://github.com/geoschem/MAPL/pull/21).

This update causes small differences in GC-Classic. No differences are expected for GCHP.

### DIFF
--- a/run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.aerosol
+++ b/run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.aerosol
@@ -2203,7 +2203,7 @@ ${RUNDIR_TES_CLIM_N2O}
 #==============================================================================
 # --- Diurnal scale factors ---
 #==============================================================================
-25 EDGAR_TODNOX $ROOT/EDGARv42/v2015-02/NO/EDGAR_hourly_NOxScal.nc NOXscale 2000/1/1/* C xy unitless 1
+25 EDGAR_TODNOX $ROOT/EDGARv42/v2015-02/NO/EDGAR_hourly_NOxScal.modified.nc NOXscale 2000/1/1/* C xy unitless 1
 26 GEIA_TOD_FOSSIL 0.45/0.45/0.6/0.6/0.6/0.6/1.45/1.45/1.45/1.45/1.4/1.4/1.4/1.4/1.45/1.45/1.45/1.45/0.65/0.65/0.65/0.65/0.45/0.45 - - - xy unitless 1
 
 #==============================================================================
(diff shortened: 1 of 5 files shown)


msulprizio (approved):
This update is simply replacing the file used for EDGAR_TODNOX. I have no objections, but would like to request that you attach an emissions table and difference plots here quantifying the impact of this fix on GCClassic output.

lizziel:
I changed this PR to draft status since I would like to investigate the HEMCO missing value handling to better. I specifically want to better understand the differences introduced by using scale factor of 1 rather than missing value. I found code in HEMCO that assigns a value of 1 for scale factor values equal to the HEMCO definition of missing value. This would suggest the original file should work okay. The differences introduced using the modified file may be due to handling of missing values in regridding. If this is the case there needs to be further investigation of which handling is better for scale factors, using missing values or data values of 1.

I also found a kludge in HEMCO for one of the missing values handling if using GCHP (see [here](https://github.com/geoschem/HEMCO/blob/4a66bae48f33e6dc22cda5ec9d4633192dee2f73/src/Core/hco_calc_mod.F90#L1040-L1046)). This should be revisited given the new missing value handling in MAPL. It is also possible additional special handling of missing values for GCHP is needed in HEMCO.

lizziel:
I am closing this draft PR. After further investigation and discussion with @christophkeller, I decided that we should not merge it in. Using data values of 1.0 instead of missing values when reading and regridding the EDGAR scale factors will dilute the scale factors that are adjacant to cells marked as missing. This is particularly true along the coast.

The current GC-Classic handling is correct by setting all input values that are equal to __FillValue_ to `HCO_MISSVAL`, and then skipping those grid cells during regrid interpolation. Later, during the GEOS-Chem run, all values equal to `HCO_MISSVAL` are set to either 0 (for masks and emission levels) or 1 (for scale factors) every timestep prior to emissions retrieval.

We therefore should keep using the current EDGAR diurnal scale factors file which uses missing values rather than data value 1.0. It happens that the `_FillValue` is 1.0 for this file, but that value is not actually used beyond identifying the cells with missing data.

GCHP will require a separate fix  to accompany the MAPL update that identifies input values equal to `_FillValue` as missing. This will be an update in HEMCO only.

PR #1257, opened 2022-05-18
Bug fixes for CH4 emissions

msulprizio:
A bug in the implementation of Scarpelli_Canada CH4 emissions in 13.4.0
resulted in incorrect oil, gas, and coal emissions over the US. The hierarchy
for the CAN_*_COAST entries was accidentally set to 100, so zero values
over the US overwrote the GEPA emissions.

The following fixes are applied:

1. Set the hierarchies for the USA, Canada, and Mexico inventories to
   the same value (100), so the emissions are added. There should be no
   overlap between the inventories, so this will prevent prioritizing
   emissions from one country over the other along borders.

2. Set hierarchies for coastal oil, gas, and coal emissions from the GEPA,
   Scarpelli_Mexico, and Scarpelli_Canada inventories to 5 to add to GFEI.
   The previous hierarchy of 1 was meant to add to EDGAR, but GFEI now
   overwrites EDGAR for oil, gas, and coal sectors.

Signed-off-by: Melissa Sulprizio <mpayer@seas.harvard.edu>

### DIFF
--- a/run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.CH4
+++ b/run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.CH4
@@ -120,57 +120,58 @@ Warnings:                    1
 # --- Gridded EPA (Maasakkers et al., Environ. Sci. Technol., 2016) ---
 #
 # NOTES:
-# - Do not use GEPA forest fire emissions. Use QFED instead.
+# - Use Hier=100 to add to Canada and Mexico regional inventories
+# - Do not use GEPA forest fire emissions. Use QFED or GFED instead.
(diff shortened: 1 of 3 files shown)


msulprizio:
The impact on methane emissions is shown below from a 4x5 global simulation and a 0.25x0.3125 North America nested-grid simulation. Changes are over the US and coastal regions as expected.

- [13.4.0_CH4_EmisFix_4x5.pdf](https://github.com/geoschem/geos-chem/files/8718105/13.4.0_CH4_EmisFix_4x5.pdf)
- [13.4.0_CH4_EmisFix_025x03125_NA.pdf](https://github.com/geoschem/geos-chem/files/8718106/13.4.0_CH4_EmisFix_025x03125_NA.pdf)

PR #1259, opened 2022-05-23
[WIP] GCHP run directory changes to easily segment runs in time

lizziel:
GCHP runs can now be broken up in time with zero differences but the procedure to break up runs is cumbersome and error-prone. This PR features changes to GCHP configuration files and example run scripts to make the process of setting up and submitting consecutive GCHP runs easy (what we have previously called multi-run). The primary changes are as follows:

1. Set start date and time in file cap_restart. Previously start was set in runConfig.sh. Along with this update is removing both start and end from runConfig.sh and configuring them in CAP.rc. Start is now way in the past (year 1900) and end is now way in the future (year 2200). MAPL ignores the CAP.rc values if cap_restart is present.
2. Store restart files in a subdirectory called Restarts. Previously restart files were in the main level of the run directory. The symbolic links to initial restart files for July 1, 2019 (fullchem sims) and Jan 1, 2019 (transport tracer sim) are also in the Restarts directory after creating a run directory.
3. Use symbolic link gchp_restart.nc4 to point to restart files in Restarts. Along with this is the removal of setting initial restart file in runConfig.sh.
4. Use set file format for all restart files. The new file format is similar to that of GC-Classic and includes the simulation start date and time as well as grid resolution. For example, GEOSChem.Restart.20190701_000000z.c24.nc4. The new shell script in the run directory called setRestartLink.sh will automatically set the restart symlink gchp_restart.nc4 to the appropriate restart file based on the start time in cap_restart and the grid resolution set in runConfig.sh.

The description above and the code is still a work-progress. Please contact me directly with comments rather than starting a conversation here since my description is not yet compelte.


lizziel:
This update also requires https://github.com/geoschem/gcpy/pull/151 which adds a logical for whether version is before 14.0. After this update the GEOS-Chem restart files have a new name and are in the Restarts subdirectory. Therefore the benchmark script needs to specify version as 14.0 to get the right restart file name and location.

LiamBindle:
I'm afraid I won't have time to take a look at this before IGC10.

msulprizio (approved):
This PR looks straightforward. I have no major comments at this time, but may circle back after testing it out.

I did have a question about the GCHP integration test scripts. I noticed we now only have scripts customized for Harvard's Cannon cluster now. Is the intention that we only run these tests on Cannon? It would be nice if we could make these scripts more generic to run on any system. Perhaps we can address this if/when we add integration testing capabilities on AWS.

run/GCHP/createRunDir.sh
```diff
@@ -388,99 +389,51 @@ elif [[ ${sim_extra_option} =~ "APM" ]]; then
...
-    enddate='20190801'
-fi
-RUNDIR_VARS+="RUNDIR_SIM_START_DATE=$startdate\n"
-RUNDIR_VARS+="RUNDIR_SIM_END_DATE=$enddate\n"
-RUNDIR_VARS+="RUNDIR_SIM_START_TIME='000000'\n"
-RUNDIR_VARS+="RUNDIR_SIM_END_TIME='000000'\n"
 RUNDIR_VARS+="RUNDIR_SIM_DUR_YYYYMMDD='00000100'\n"
 RUNDIR_VARS+="RUNDIR_SIM_DUR_HHmmSS='000000'\n"
 
 # Use monthly diagnostics by default
```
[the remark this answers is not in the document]
lizziel: I checked the HISTORY.rc defaults and it is already monthly diagnostics. I can remove the RUNDIR_VARS for freq, duration, and monthly diag but wonder if these are needed for anything that I'm not aware of?

LiamBindle:
Hi all, I looked at the new run directory yesterday and I think it's great. It makes running multi-segment runs WAY easier! Great work @lizziel!

LiamBindle (approved):

run/GCHP/runScriptSamples/gchp.batch_job.sh
```diff
@@ -0,0 +1,157 @@
...
+
+#################################################################
+#
+# PRE-RUN COMMANDS
+#
+# Print loaded modules
+module list     
+
+# Define log name to include simulation start date
+start_str=$(echo $(cat cap_restart) | sed 's/ /_/g')
```
kilicomu, line 113: @lizziel I think this (and similar occurrences) is a little neater written as:

```
sed 's/ /_/g' cap_restart
```

run/GCHP/createRunDir.sh
```diff
@@ -545,16 +498,15 @@ RUNDIR_VARS+="$(cat ${gcdir}/run/shared/settings/gmao_hemco.txt)\n"
...
 
 #--------------------------------------------------------------------
 # Print run direcory setup info to screen
 #--------------------------------------------------------------------
 
-printf "\n  -- This run directory has been set up for $startdate - $enddate.\n"
-printf "\n  -- The default frequency and duration of diagnostics is set to monthly.\n"
-printf "\n  -- You may modify these settings in runConfig.sh.\n"
+printf "\n  -- This run directory has been set up to start on $start_date. Edit this in file cap_restart.\n"
+printf "\n  -- Edit other commonly changed run settings in setCommonRunSettings.sh, geoschem_config.yml, and HEMCO_Config.rc.\n"
```
kilicomu, line 509: @lizziel Super minor thing - the new run directory setup info doesn't match the line width of the rest of the script, which comes out a little jarring when running it.

run/GCHP/runScriptSamples/gchp.batch_job.sh
```diff
@@ -0,0 +1,157 @@
...
+# POST-RUN COMMANDS
+#
+# If new start time in cap_restart is okay, rename and move restart file
+# and update restart symlink
+new_start_str=$(echo $(cat cap_restart) | sed 's/ /_/g')
+if [[ "${new_start_str}" = "${start_str}" || "${new_start_str}" = "" ]]; then
+   echo "ERROR: cap_restart either did not change or is empty."
+   exit 1
+else
+    N=$(grep "CS_RES=" setCommonRunSettings.sh | cut -c 8- | xargs )
```
kilicomu, line 154: @lizziel Does the empty `xargs` do anything? Not seen that pattern before if so.

kilicomu:
Other than a couple of minor things, this looks really cool! And will hopefully make running my longer simulations less manually troublesome. Thanks for doing it.

run/GCHP/runScriptSamples/gchp.batch_job.sh
lizziel: Indeed. Thanks for the tip! I'll push that update.

run/GCHP/runScriptSamples/gchp.batch_job.sh
lizziel: I used xargs here to remove whitespace if any.

run/GCHP/createRunDir.sh
lizziel: Good catch. I'll push a fix to that.

run/GCHP/runScriptSamples/gchp.batch_job.sh
lizziel: Let me know if you have a better way.

lizziel:
All GCHP integration tests now pass with zero differences compared to 14.0.0-rc.0.

run/GCHP/runScriptSamples/gchp.batch_job.sh
kilicomu: No that's cool, didn't think about using it that way and it's shorter than what I would do (`tr` or `sed`)!

msulprizio:
>Use a fixed file format for all restart files. The new file format is similar to that of GC-Classic and includes the simulation start date and time as well as grid resolution. For example, GEOSChem.Restart.20190701_000000z.c24.nc4.

@lizziel Is there a reason the restart filename for GCHP includes hhmmss instead of just hhmm as is done for the GCClassic restart file? If not, could we remove the use of seconds so the filenames between models are consistent?

msulprizio:
Thanks for clarifying. For now, users who generate GCHP restart files from GCClassic ones will just need to remember to include seconds in the GCHP restart filename. This may be fine or we can think about making it consistent between GCClassic and GCHP restart files at some point.

lizziel:
I checked the filename of the optional restarts that can be output mid-run and they do not include seconds. Unlike the end-of-run restart those files are automatically named within MAPL to include the time. Is it too late to remove the seconds from the restart filename in 14.0? This would then make it consistent with both GC-Classic restart and the MAPL checkpoints.

lizziel:
Following up, two PRs came out of this discussion to remove seconds:

1. https://github.com/geoschem/geos-chem/pull/1302
2. https://github.com/geoschem/geos-chem/pull/1303

PR #1260, opened 2022-05-23
Fix HEMCO diagnostic counter zero warnings in full chemistry simulation

lizziel:
In the dev version of 14.0.0 I noticed that several HEMCO warnings about diagnostic counters being zero are printed every timestep to the HEMCO log file. Updates in this PR will prevent the initialization of HEMCO diagnostics that are never updated if using the full chemistry simulation.


msulprizio (approved):
Looks good. This PR comments out the BrSALA and BrSALC emissions diagnostics. If users enable the sea salt debromination option (CalcBrSeasalt) in HEMCO_Config.rc, they will need to manually uncomment those lines in HEMCO_Diagn.rc.

PR #1262, opened 2022-05-25
Reduce root logging level for MAPL to WARNING

sdeastham:
The INFO logging level still generates a large number of messages which are excessive unless performing debugging. I suggest we reduce this to WARNING for MAPL, which results in much cleaner log files.


lizziel:
This looks fine to me. This will be a zero diff update, only impacting the MAPL log file.

PR #1279, opened 2022-06-14
Feature/user registration: optionally register first time users with dynamodb database

laestrada:
This is a workin progress pr that will allow for registration of new users of geoschem. For first time users, few additional questions are asked (email, research interests, affiliation), which are then sent to a dynamodb database via a POST api request. 

A list of registered users is then available from a [GC User dashboard](https://gc-dashboard.org/users), which I have mocked up with some sample data.

Is there additional info that would be useful for registration? FYI, the dashboard currently uses a crazy ephemeral url that I plan to switch to a more static custom domain for the actual pr (eg. geoschem.org), but suggestions for domain names are appreciated!

Here is an example of the `./createrundir` flow for first time users:
Screen Shot 2022-06-27 at 6 06 41 PM

### DIFF
--- a/run/GCClassic/createRunDir.sh
+++ b/run/GCClassic/createRunDir.sh
@@ -19,6 +19,7 @@ function post_registration() {
     site="$3"
     git_username="$4"
     research_interest="$5"
+    env_name="$6"
     curl --location --request POST "https://gc-dashboard.org/registration" \
         --header "Content-Type: text/plain" \
         --data-raw "{
@@ -27,6 +28,8 @@ function post_registration() {
             \"site\": \"${site}\",
             \"git_username\": \"${git_username}\",
             \"research_interest\": \"${research_interest}\"
+            \"model_type\": \"GCClassic\",
+            \"env_name\": \"${env_name}\"
         }"
 }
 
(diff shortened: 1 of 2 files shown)


LiamBindle (approved):

yantosca (approved):
I think it looks good.  It'd also be nice to ask if they plan on using GC on a computer cluster or the AWS cloud (as Daniel often asks for that type of information).  

I would almost change "What is your application for GEOS-Chem" to "Please tell us how you plan on using GEOS-Chem in your research.  (We will add this information to the GEOS-Chem website.)".  That will make sure that we get a good blurb for the web page.

laestrada:
> Ask "What modes of GEOS-Chem do you plan to use?" with options for GC-Classic, GCHP, and GCHP adjoint. We'll need to allow multiple in the response.

My current thought is that we can hardcode in GCClassic, GCHP (and the adjoint?) based on the run directory they create to avoid asking too many questions.

laestrada:
> I also wonder if we'll want to know what types of simulations people are using. We occasionally get asked how many users are using a specialty simulation. I think that most recently came up for POPs. If we had that information we could contact people if, say, we wanted to retire a simulation type.

Happy to add this if people think it's important -- do you have a specific wording in mind?

djxjacob:
I don’t know, it seems like too much information to request –  Daniel

laestrada:
The wording/ format has now been approved by Randall and Daniel. I consider it ready to merge, perhaps for 14.1

msulprizio:
Hi @laestrada. I merged this up to patch/14.0.1 and tried to execute but after the last prompt, I get:

```
curl: option --data-raw: is unknown
curl: try 'curl --help' or 'curl --manual' for more information
```

Then createRunDir.sh proceeds to the next prompt to select simulation type. I checked ~/.geos-chem/config after that and it still indicates `GC_FIRST_TIME_USER=true`. I'll hold off on merging this PR until this error is resolved.

laestrada:
@msulprizio I believe it should be fixed -- cannon's default curl version doesn't have the --data-raw flag. I've updated it to use -d instead, which should be more universal! I also added a prompt asking for a users name and updated the env variable in .geoschem to be GC_USER_REGISTERED, since I think the previous variable name was a little unclear. Should be all set!

djxjacob:
Excellent!

PR #1302, opened 2022-07-11
Change GCHP restart filename convention to exclude seconds

lizziel:
This update changes the GCHP end-of-run restart filename format to exclude seconds in the time. This makes the time format in GCHP restart files match the format used in GC-Classic restart files and GCHP mid-run checkpoint files.

NOTE: This PR also removes two files from the GCHP run directory folder for general cleanup.
1. `cap_restart` template since not used when creating a run directory.
2. `gchp.local.run` in the Harvard operational examples folder, since it is identical to the generic `gchp.local.run` provided in `runScriptSamples`.


run/GCHP/runScriptSamples/operational_examples/harvard_cannon/gchp.submit_consecutive_jobs.sh
```diff
@@ -8,7 +8,7 @@
 # One argument: number of runs
 numRuns=${1}
 if [[ "x${numRuns}" == "x" ]]; then
-    echo "ERROR: Specify number of runs as argument, e.g. ./gchp.multirun.submit 2"
+    echo "ERROR: Specify number of runs as argument, e.g. ./gchp.submit_consecutive_jobs.sh 2"
     exit
 fi
```
msulprizio, line 14: I think line 17  below also needs to be changed from:

```
echo "Duration: $(grep "DURATION=" setCommonRunSettings.sh | cut -c 8- | xargs)"
```

 to:

```
 echo "Duration:  $(grep "Run_Duration=" setCommonRunSettings.sh | cut -c 14- | xargs)"
 ```

lizziel: Good catch. All set.

PR #1316, opened 2022-07-25
Add capability to write metadata for defined species to a YAML file

yantosca:
## Overview
This PR now gives GEOS-Chem the capability to write metadata for defined species (as opposed to all possible species listed in `species_database.yml`) to a YAML file.  This was identified as a need by @jimmielin for interfacing GEOS-Chem to external models such as CESM.

## How it works:
A new tag `species_metadata_output_file` has been added to all of the template `geoschem_config.yml` files:

```YAML
simulation:
  name: fullchem
  start_date: [20190701, 000000]
  end_date: [20190701, 010000]
  root_data_dir: /n/holyscratch01/external_repos/GEOS-CHEM/gcgrid/data/ExtData
  met_field: MERRA2
  species_database_file: ./species_database.yml
  species_metadata_output_file: OutputDir/geoschem_species_metadata.yml  # <-- this is new
  debug_printout: false
  use_gcclassic_timers: true 
```
Here you can specify a path where the YAML file will be created, or type `none` to skip.  A YAML file will be created (only on the root CPU) containing all of the metadata for defined species.  Species metadata that are undefined will not be printed.

This should go into 14.1.0, as it contains a change to the run directory.  But you can pull from this branch if you need it for CESM development in the meantime.


Headers/species_database_mod.F90
```diff
@@ -838,6 +838,21 @@ SUBROUTINE Init_Species_Database( Input_Opt, SpcData, SpcCount, RC )
 32  FORMAT( a30, " | ", 3f10.2 )
 40  FORMAT( a30, " | ", i10    )
 
+    !=======================================================================
+    ! Print metadata for only the species that are defined in this
+    ! simulation (but not the entire species database) to another YAML
+    ! file.  This facilitates coupling GEOS-Chem to external models.
```
lizziel, line 844: I don't think we need to put that this facilitates coupling to external models since I think the new file will be useful as a reference for any GEOS-Chem user.

lizziel (changes requested):
Looks good. I think we should also turn off the species database prints if debug prints are on to avoid redundant information in the new yaml and the log file. Is it correct that the new file will contain the same information that we previously printed to log when debug prints were enabled, via spc_print?

yantosca:
Yes, it should be the same information.  I can toggle the qfyaml_print output off if we input_opt%lprt is true.  Will push an update soon.

lizziel:
I also want to verify this only included advected species. So if a species is in KPP and the species database, but not in the advected list, it won't be included.

jimmielin:
> I also want to verify this only includes advected species. So if a species is in KPP and the species database, but not in the advected list, it won't be included.

I think we need both advected- and non-advected species. CESM needs a list of non-advected species after aerosols in `solsym` and `adv_mass`, and WRF will output these as well for diagnostic purposes. So we definitely need the full list - the advected species list can already be obtained from `input.geos` (or yaml equivalent) so it would not be as useful.

lizziel:
I don't think we include all non-advected species in the species database. @yantosca, is that right?

lizziel:
I just realized including the is_advected entry would be sufficient to not have to cross-reference with input.geos.

yantosca:
> I don't think we include all non-advected species in the species database. @yantosca, is that right?

We include all species in the database as we at least need formula, longname, and MW even for KPP-only species.  They all get added to the list of species names after the advected species.

jimmielin:
Thanks @yantosca, that’s exactly what we need!

yantosca:
Commit #3c50dce5950132293e10cfe20d54025e9c2d449a has added logic to:
  1. Block out the `IF ( Input_Opt%amIRoot )` for MODEL_CESM only
  2. Skip calling `QFYAML_Print` when `prtDebug == .TRUE.`

jimmielin:
Thanks @yantosca! I've updated the companion code (https://github.com/jimmielin/geos-chem-coupling-tools) to support YAML output now.

For some reason, `pyyaml` that I am using parses `NO: ...` as `False`. What a strange bug...

yantosca:
@jimmielin: I believe "NO" is a YAML alternative for "false".  I might have to add a test so that if we print "NO" on the first line, we put it in quotes.  I can look at that in a bit.

Headers/species_database_mod.F90
yantosca: That's fine.  I'll make the edit.

yantosca:
> @jimmielin: I believe "NO" is a YAML alternative for "false". I might have to add a test so that if we print "NO" on the first line, we put it in quotes. I can look at that in a bit.

This is now fixed in commit 5c211aaa35e89b688917fc376dab1ffd24240803.  We now get this output in the `OutputDir/geoschem_species_metadata.yml` file:
```YAML
'NO':
  Background_VV: 4.0e-13
  Formula: 'NO'
  FullName: Nitrogen oxide
  Is_Advected: true
  Is_Gas: true
  Is_Photolysis: true
  MW_g: 30.01
```

Headers/species_database_mod.F90
yantosca: @lizziel The comment has been edited in 05704322001daf770881422339666ea3594b6caf.

lizziel:
I think this line that prints the species database information to the standard out is still present: https://github.com/geoschem/geos-chem/blob/63987bf019e0d9a0fe35e469ff7c50cfef52de93/Headers/species_database_mod.F90#L827

I think it would be better to only have one option for printing out the species database information to avoid confusion and duplication.

yantosca:
> I think this line that prints the species database information to the standard out is still present:
> 
> https://github.com/geoschem/geos-chem/blob/63987bf019e0d9a0fe35e469ff7c50cfef52de93/Headers/species_database_mod.F90#L827
> 
> I think it would be better to only have one option for printing out the species database information to avoid confusion and duplication.

@lizziel This is now addressed in commit db06d3ab5cc056a006af5db29115165c96405faf.

lizziel (approved):

yantosca:
I've merged `dev/14.1.0` into the `feature/yaml-spc-output` branch.  The original branch was far behind the current state of the 14.1.0 development.  The `gc_2x25_CO2_merra` integration test failed due to what I think was trying to read some files that have since been removed.

yantosca:
After merging 14.1.0 into this feature/yaml-spc-output branch, rerunning the GEOS-Chem Classic integration tests resulted in all tests passing:
```console

Using 24 OpenMP threads
Number of execution tests: 48
 
Execution tests:
------------------------------------------------------------------------------
gc_025x03125_CH4_geosfp_47L_na...................Execute Simulation.....PASS
gc_025x03125_fullchem_geosfp_47L_na..............Execute Simulation.....PASS
gc_05x0625_CH4_merra2_47L_na.....................Execute Simulation.....PASS
gc_05x0625_fullchem_merra2_47L_na................Execute Simulation.....PASS
gc_2x25_aerosol_merra2...........................Execute Simulation.....PASS
gc_2x25_CH4_merra2...............................Execute Simulation.....PASS
gc_2x25_CO2_merra2...............................Execute Simulation.....PASS
gc_2x25_fullchem_merra2..........................Execute Simulation.....PASS
gc_2x25_Hg_merra2................................Execute Simulation.....PASS
gc_2x25_metals_merra2............................Execute Simulation.....PASS
gc_2x25_POPs_BaP_merra2..........................Execute Simulation.....PASS
gc_2x25_tagCH4_merra2............................Execute Simulation.....PASS
gc_2x25_tagCO_merra2.............................Execute Simulation.....PASS
gc_2x25_tagO3_merra2.............................Execute Simulation.....PASS
gc_2x25_TransportTracers_merra2..................Execute Simulation.....PASS
gc_2x25_TransportTracers_merra2_LuoWd............Execute Simulation.....PASS
gc_4x5_aerosol_geosfp............................Execute Simulation.....PASS
gc_4x5_aerosol_merra2............................Execute Simulation.....PASS
gc_4x5_CH4_geosfp................................Execute Simulation.....PASS
gc_4x5_CH4_merra2................................Execute Simulation.....PASS
gc_4x5_fullchem_aciduptake_merra2................Execute Simulation.....PASS
gc_4x5_fullchem_APM_merra2.......................Execute Simulation.....PASS
gc_4x5_fullchem_benchmark_merra2.................Execute Simulation.....PASS
gc_4x5_fullchem_complexSOA_merra2................Execute Simulation.....PASS
gc_4x5_fullchem_complexSOA_SVPOA_merra2..........Execute Simulation.....PASS
gc_4x5_fullchem_geosfp...........................Execute Simulation.....PASS
gc_4x5_fullchem_marinePOA_merra2.................Execute Simulation.....PASS
gc_4x5_fullchem_merra2...........................Execute Simulation.....PASS
gc_4x5_fullchem_merra2_47L.......................Execute Simulation.....PASS
gc_4x5_fullchem_merra2_LuoWd.....................Execute Simulation.....PASS
gc_4x5_fullchem_RRTMG_merra2.....................Execute Simulation.....PASS
gc_4x5_fullchem_TOMAS15_merra2_47L...............Execute Simulation.....PASS
gc_4x5_fullchem_TOMAS40_merra2_47L...............Execute Simulation.....PASS
gc_4x5_Hg_geosfp.................................Execute Simulation.....PASS
gc_4x5_Hg_merra2.................................Execute Simulation.....PASS
gc_4x5_metals_merra2.............................Execute Simulation.....PASS
gc_4x5_POPs_BaP_geosfp...........................Execute Simulation.....PASS
gc_4x5_POPs_BaP_merra2...........................Execute Simulation.....PASS
gc_4x5_tagCH4_geosfp.............................Execute Simulation.....PASS
gc_4x5_tagCH4_merra2.............................Execute Simulation.....PASS
gc_4x5_tagCO_geosfp..............................Execute Simulation.....PASS
gc_4x5_tagCO_merra2..............................Execute Simulation.....PASS
gc_4x5_tagO3_geosfp..............................Execute Simulation.....PASS
gc_4x5_tagO3_merra2..............................Execute Simulation.....PASS
gc_4x5_TransportTracers_geosfp...................Execute Simulation.....PASS
gc_4x5_TransportTracers_geosfp_LuoWd.............Execute Simulation.....PASS
gc_4x5_TransportTracers_merra2...................Execute Simulation.....PASS
gc_4x5_TransportTracers_merra2_LuoWd.............Execute Simulation.....PASS
 
Summary of test results:
------------------------------------------------------------------------------
Execution tests passed: 48
Execution tests failed: 0
Execution tests not yet completed: 0

```

yantosca:
Rerunning the GCHP integration tests resulted in all tests passing as well:
```console
GCHP: Execution Test Results

Number of execution tests: 3
 
Execution tests:
------------------------------------------------------------------------------
gchp_fullchem_benchmark_merra2_c48...............Execute Simulation.....PASS
gchp_fullchem_standard_merra2_c24................Execute Simulation.....PASS
gchp_TransportTracers_geosfp_c24.................Execute Simulation.....PASS
 
Summary of execution test results:
------------------------------------------------------------------------------
Execution tests passed:        3
Execution tests failed:        0
Execution tests not completed: 0

```

lizziel:
I am continuing this PR thread post-merge since I am now testing GEOS-Chem 14.1 in CESM.

I have to comment out writing the species yaml file in CESM since it is causing an error. Was the intention to output the yaml file as part of a GC-Classic dry-run (rather than CESM) and then use the output with (https://github.com/jimmielin/geos-chem-coupling-tools) to create the information needed in CESM files? If we don't need to generate it from CESM I am thinking we could archive it for each benchmarked GEOS-Chem version and link to it on the wiki so that it is available to use for pre-processing as needed in other models. I think it is created for all GC-Classiv runs by default.

@yantosca, @jimmielin Do you have any thoughts on this?

jimmielin:
Hi @lizziel, the intention was to run this with GC-Classic to facilitate coupling, since CESM doesn't run at all if the species list is incorrect and WRF-GC requires a full clean recompile to update the species list. So this would be part of the update procedure for GEOS-Chem in each model. I'm curious to hear everyone's thoughts on this as well.

Thanks!
Haipeng

lizziel:
Thanks Haipeng. I'll bring up wiki posting of the species yaml file per version at the next GCST meeting. I'll put a CESM fix in 14.1.1.

Issue #1309, opened 2022-07-18
[BUG/ISSUE] 550nm AOD used when scaling aerosol/dust for J-values computation


lizziel:
I discovered a bug in 14.0,0 in which the incorrect AOD wavelength logical is used in subroutine `PHOTO_JX`, resulting in calculating the 550 nm scaling rather than the 999 nm scaling for aerosol/dust.

This bug was introduced in 14.0. Prior to that version the AOD wavelength indicator was set to 0 (999 nm) for use as an input argument to `RDAER` within `Do_Fullchem`. In 13.4 this block of code was moved out of `Do_Fullchem` and into `Do_Chemistry`. The AOD wavelength indicator was still declared in `Do_Fullchem` following this update but was never set. As a result, a garbage value of it was passed to `FAST_JX`, ultimately causing the logical `AOD999` passed to `PHOTO_JX` to be `.FALSE.`, unless the garbage value happened to be 0.

The fix is simply setting `WAVELENGTH=0` prior to calling `FAST_JX`. I will create a companion PR with that fix. I will also do a 1-month benchmark to assess the impact of this bug.

lizziel:
There is a delay in releasing 14.0 so this bug fix can now be included in that version. This means the bug will not appear in any GEOS-Chem releases.

PR #1319, opened 2022-07-26
Specify AOD wavelength 999 nm for use in FAST-JX

lizziel:
This PR fixes a bug in 14.0.0 in which the incorrect AOD wavelength logical is used in subroutine PHOTO_JX. That bug resulted in calculating the 550 nm scaling rather than the 999 nm scaling for aerosol/dust. The fix is simply setting WAVELENGTH=0 prior to calling FAST_JX.

I did a 1-month benchmark comparison of this fix using 14.0. Differences are small. Benchmark tables with ref being 14.0.0-rc.1 and dev being 14.0.0-rc.1 plus the fix are as follows:

[OH_metrics.txt](https://github.com/geoschem/geos-chem/files/9192275/OH_metrics.txt)
[GlobalMass_Trop.txt](https://github.com/geoschem/geos-chem/files/9192281/GlobalMass_Trop.txt)
[GlobalMass_TropStrat.txt](https://github.com/geoschem/geos-chem/files/9192277/GlobalMass_TropStrat.txt)
[Emission_totals.txt](https://github.com/geoschem/geos-chem/files/9192278/Emission_totals.txt)

closes https://github.com/geoschem/geos-chem/issues/1309

### DIFF
--- a/GeosCore/fullchem_mod.F90
+++ b/GeosCore/fullchem_mod.F90
@@ -334,6 +334,7 @@ CONTAINS
     ENDIF
 
     ! Do Photolysis
+    WAVELENGTH = 0
     CALL FAST_JX( WAVELENGTH, Input_Opt,  State_Chm,                         &
                   State_Diag, State_Grid, State_Met, RC                     )
 


lizziel:
There is a delay in releasing 14.0.0 so we can now include this fix in that version.

PR #1328, opened 2022-08-01
Updates for GEOS from GMAO

lizziel:
Updates included in this pull request are from Christoph Keller and are relevant only for running GEOS-Chem in GEOS.  These updates are zero diff for GEOS-Chem Classic and GCHP. The primary update is abstracting GEOS-only code out of file `GCHP/Interfaces/Chem_GridCompMod.F90` so that the file mostly contains code used by both GCHP and GEOS. It also adds a GEOS-only diagnostic for troposphere NOx lifetime.


lizziel:
I requested two reviewers: @christophkeller to double-check that the updates include everything that was intended, and @yantosca as a general GCST reviewer.

yantosca (approved):
A couple of general comments:

  1. I think in KPP 2.5.0 we (in 14.0.0) the `Fun()` routine of KPP returns Vdotout, but this was removed in KPP 3.0.0.  So this might crash when we update to 14.1.0.
  2. The GEOS updates add a lot of extra code to `fullchem_mod.F90`.  Was wondering if some of these updates can be abstracted into subroutines of `fullchem_mod.F90` eventually so as to keep the `DO_FULLCHEM` routine as clean as possible.  Food for thought for future development.

lizziel:
Thanks for the feedback @yantosca.

> I think in KPP 2.5.0 we (in 14.0.0) the Fun() routine of KPP returns Vdotout, but this was removed in KPP 3.0.0. So this might crash when we update to 14.1.0.

Why was returning Vdotout (time derivative of variable species) removed in KPP 3.0.0? This is needed for the NOx lifetime diagnostic in GEOS. 

> The GEOS updates add a lot of extra code to fullchem_mod.F90. Was wondering if some of these updates can be abstracted into subroutines of fullchem_mod.F90 eventually so as to keep the DO_FULLCHEM routine as clean as possible. Food for thought for future development.

It looks like in general there is a lot of diagnostic code in `Do_Fullchem`, including both GEOS as well as the offline model. Maybe all of that diagnostic code could be abstracted to clean it up. We have subroutine Diag_OH_HO2_O1D_O3P in the module as precedent for that. I'll make a feature request for that.

PR #1343, opened 2022-08-09
Fix bug in setCommonRunSettings.sh that cause "RECORD_REF_DATE: cap_restart" and "RECORD_REF_TIME: cap_restart"

LiamBindle:


lizziel:
@LiamBindle, could you add a description to this PR with what the problematic behavior is, in what version, and when this fix should be pulled in?

lizziel (approved):

lizziel:
This PR should be pulled into the `feature/minor_rundir_improvements` branch before release of 14.0.

PR #1354, opened 2022-08-21
Add per-domain FIRST switch in ExtState_SetFields for WRF-GC

jimmielin:
Hi GCST,

This is a WRF-GC (`MODEL_WRF`)-only fix to make `FIRST` scalar in `ExtState_SetFields` for `HCO_Interface_GC_Mod` specific to each domain in WRF-GC (each domain is numbered in `State_Grid%ID` from 1 up to 8).

This affects the same file as https://github.com/geoschem/geos-chem/pull/1348 but serves a different purpose.

Thanks!
Haipeng


yantosca (approved):
This looks good to merge.  We can get this into GEOS-Chem 14.0.1 and HEMCO 3.5.1 as it is a simple fix that shouldn't affect GCClassic or HEMCO.

PR #1362, opened 2022-08-29
Fix gcap2 run

ltmurray:
Needed to adjust some additional files and HEMCO won't parse setting variables from HEMCO_Config.rc with numbers in them, e.g., GCAP2SCENARIO needed to be changed to GCAPSCENARIO, as well as point users to the GCAP2-regridded restart files.

No changes should have impacted the standard run directory creation. The one to keep an eye out on is a new setting for turning AEIC on or off by default (RUNDIR_USE_AEIC), which should still be true for all the normal runs and false for GCAP2 runs.


yantosca (approved):
Approved!  We'll merge it into 14.0.1.  The modifications are as described in the conversation.

msulprizio:
This PR has been merged into `patch/14.0.1` and `dev` and will be included in the 14.0.1 release scheduled imminently. I will close the PR for now, but it won't be reported as merged until it has been brought into main (i.e. when 14.0.1 has been released).

Issue #1386, opened 2022-09-19
[BUG/ISSUE] GC-Classic bug causing high ozone bias over poles since 13.4


lizziel:
### What institution are you from?
Harvard University

### Description of the problem
There has been a discrepancy shown in the GCHP and GEOS-Chem Classic (GCC) benchmark comparisons since 13.4. It is most noticeable as an increase in tropospheric ozone over the poles in GC-Classic between 13.3 and 13.4 that is not present in GCHP. See plots [here](http://ftp.as.harvard.edu/gcgrid/geos-chem/1mo_benchmarks/GC_13/13.4.0-rc.4/GCHP/BenchmarkResults/GCHP_GCC_comparison/Oxidants/Oxidants_Surface.pdf) and [here](http://ftp.as.harvard.edu/gcgrid/geos-chem/1mo_benchmarks/GC_13/13.4.0-rc.4/GCClassic/BenchmarkResults/Oxidants/Oxidants_Surface.pdf) for the 13.4 1-month benchmark comparisons that show this.

Further investigation shows the discrepancy is due to changes in the GCC pressure fixer. The pressure fixer file was modified during 13.4 development to set variables and arrays to zero prior to assigning values. The modifications were not intended to introduce changes.

### Description of troubleshooting performed
I did a 1-month full-chemistry benchmark run using 13.4 but swapping out the pressure fixer file with the version used in 13.3. I then compared to the 13.4 1-month benchmark data. The elevated ozone signature goes away with the change. See a comparison of GCC before and after the fix [here](http://ftp.as.harvard.edu/gcgrid/geos-chem/validation/GCC_pressure_fixer_bug/1mo_benchmark_13.4_comparison/). 

### GEOS-Chem version
14.0.0 (still in development)

### Description of modifications
Revert file `pjc_pfix_mod.F90` to the version in the 13.3 release.

PR #1387, opened 2022-09-19
Bug fix: revert GC-Classic pressure fixer to v13.3

lizziel:
This pull request reverts the GEOS-Chem Classic pressure fixer (file `pjc_fix_mod.F90`) to the version used in GEOS-Chem 13.3. This addresses a problem introduced in 13.4 in which several variables were set to zero prior to assignment, inadvertently introducing changes to the model.

The impact of the pressure fixer changes are visible in the 13.4 benchmark plots but went under the radar due to other changes introduced in that version. It is most noticeable as an increase in tropospheric ozone over the poles in GC-Classic between 13.3 and 13.4 that is not present in GCHP. See plots [here](http://ftp.as.harvard.edu/gcgrid/geos-chem/1mo_benchmarks/GC_13/13.4.0-rc.4/GCHP/BenchmarkResults/GCHP_GCC_comparison/Oxidants/Oxidants_Surface.pdf) and [here](http://ftp.as.harvard.edu/gcgrid/geos-chem/1mo_benchmarks/GC_13/13.4.0-rc.4/GCClassic/BenchmarkResults/Oxidants/Oxidants_Surface.pdf) for the 13.4 1-month benchmark comparisons that show this.

This PR will cause changes in the benchmark simulation.

closes https://github.com/geoschem/geos-chem/issues/1386

### DIFF
--- a/GeosCore/pjc_pfix_mod.F90
+++ b/GeosCore/pjc_pfix_mod.F90
@@ -72,21 +72,21 @@ MODULE PJC_PFIX_MOD
   REAL(fp), ALLOCATABLE :: SINE_FV(:)
 
   ! Scalar variables
-  LOGICAL               :: PR_DIAG
-  INTEGER               :: LOC_PROC
-  REAL(fp)              :: GEOFAC_PC
-  REAL(fp)              :: DLON_FV
+  LOGICAL             :: PR_DIAG
+  INTEGER             :: LOC_PROC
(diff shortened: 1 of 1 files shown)


yantosca (approved):
This PR restores the GC-Classic pressure fixer (and undoes several zeroing of variables, such as dps_ctm) that had caused the behavior described in issue #1386.

Issue #1388, opened 2022-09-19
[BUG/ISSUE] Special handling for MOH in dry deposition is missing


lizziel:
### Description of the problem
Special handling for MOH in GEOS-Chem dry deposition is missing. Species that are included in the SeaFlux extension of HEMCO should not be dry deposited by GEOS-Chem unless over land. This is because the deposition is handled by HEMCO instead. Special handling to skip dry deposition unless over land for other SeaFlux species is present. The same special handling was omitted when MOH was added in GEOS-Chem 12.7.

### GEOS-Chem version
14.0 (still in development)

### Description of modifications
Add the same special handling for MOH as currently done for acetone in `drydep_mod.F90`.

### What institution are you from?
Harvard University

NOTE: This issue was identified by Kelvin Bates.

msulprizio:
This fix (PR #1389) has now been merged into `dev` and will be included in 14.0.0.

PR #1389, opened 2022-09-19
Add special treatment for MOH in dry deposition if not over land

lizziel:
This update adds the same special handling for MOH as currently done for acetone and other HEMCO SeaFlux extension species in `drydep_mod.F90`. This fixes a bug where special handling for MOH in GEOS-Chem dry deposition was missing.

Species that are included in the SeaFlux extension of HEMCO should not be dry deposited by GEOS-Chem unless over land. This is because the deposition is handled by HEMCO instead. Special handling to skip dry deposition unless over land for other SeaFlux species is present. The same special handling was omitted when MOH was added in GEOS-Chem 12.7.

closes https://github.com/geoschem/geos-chem/issues/1388

### DIFF
--- a/GeosCore/drydep_mod.F90
+++ b/GeosCore/drydep_mod.F90
@@ -159,7 +159,7 @@ MODULE DRYDEP_MOD
   INTEGER                        :: NUMDEP,   NWATER
   INTEGER                        :: DRYHg0,   DRYHg2,   DryHgP
   INTEGER                        :: id_ACET,  id_ALD2,  id_O3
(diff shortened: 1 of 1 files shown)


msulprizio (approved):
This is a straightforward fix to remove dry deposition of MOH over oceans from drydep_mod.F90. Looks good to merge.

lizziel:
Tagging the Surface-Atmosphere Exchange Working Group Co-Chairs on this bug fix going into 14.0.0.
@jgeddes85, @ltmurray, @amospktai

PR #1400, opened 2022-09-25
[PULL REQUEST] Fix incorrect RETURN for deallocation in Cleanup_Vdiff

jimmielin:
This fixes a bug where `Cleanup_Vdiff` would not properly deallocate all variables because there was a `RETURN` statement which did not check for a bad RC value and instead was always evaluated.


msulprizio (approved):
Thanks @jimmielin! We can include this fix in our next version release.

PR #1408, opened 2022-09-30
[PULL REQUEST] Update registration to allow for user to input computational environment

msulprizio:
@lizziel wrote:

>Hi all, I just created a run directory from 14.0 and filled out the registration. It looks good except for this part:
```
-----------------------------------------------------------
How do you plan to run GEOS-Chem?
-----------------------------------------------------------
   1. Local Cluster
   2. AWS
```
>It wouldn't let me say 1 and 2, and it doesn't have the option of other remote clusters, such as Pleiades
>Could we make this be a text response, with question "Where do you plan to run GEOS-Chem (e.g. local compute cluster, AWS, other supercomputer)?"


laestrada (approved):
Looks good!

PR #1425, opened 2022-10-06
Fix bug in computing State_Met fields IsLand, IsWater, IsOcean, IsIce

lizziel:
This PR is a minor bug fix for computing `State_Met%IsLand`, `State_Met%IsWater`, `State_Met%IsOcean`, and `State_Met%IsIce`. These logical masks are computed based on fraction land with no snow or ice. 
https://github.com/geoschem/geos-chem/blob/412ba5906b2aa9dcd72a2188c76c50220c3697b7/GeosCore/calc_met_mod.F90#L306-L310

The masks are determined based in part on fraction of land without snow or ice. However, this quantity is calculated by subtracting fraction land ice from fraction land when input met-field for fraction land does not include land ice.
https://github.com/geoschem/geos-chem/blob/412ba5906b2aa9dcd72a2188c76c50220c3697b7/GeosCore/calc_met_mod.F90#L295

I verified that FRLANDIC is indeed separate from FRLAND by printing out the fractions per grid box. Here is an example of printed fractions of a single all land ice grid cell:

>  Constants file
>     => FRLAND    :    0.0000000000000000     
>     => FRLANDIC  :    1.0000000000000000     
>     => FRLAKE    :    0.0000000000000000     
>     => FROCEAN   :    0.0000000000000000     
>  A1 file
>     => FRSEAICE  :    0.0000000000000000     
>     => FRSNO     :    0.0000000000000000     
>  Computed values
>     => FRLAND_NoSno_NoIce :   -1.0000000000000000     
>     => FrWater             :    0.0000000000000000     
>     => FrIce               :    1.0000000000000000     
>     => FrSnow              :    0.0000000000000000 

The result is that land fraction is underestimated (and potentially negative) for cells in which both land and land ice are present. The fix is to remove the subtraction of FRLANDIC from FRLAND.

### DIFF
--- a/GeosCore/calc_met_mod.F90
+++ b/GeosCore/calc_met_mod.F90
@@ -292,15 +292,16 @@ CONTAINS
        IsLocNoon(I,J) = ( LocTimeSec(I,J)          <= 43200  .and. &
                           LocTimeSec(I,J) + Dt_Sec >= 43200 )
 
-       FRLAND_NOSNO_NOICE = State_Met%FRLAND(I,J) - State_Met%FRSNO(I,J) - State_Met%FRLANDIC(I,J)
+       ! Land without snow or ice
+       FRLAND_NOSNO_NOICE = State_Met%FRLAND(I,J) - State_Met%FRSNO(I,J)
 
        ! Water without sea ice
        FRWATER = State_Met%FRLAKE(I,J) + State_Met%FROCEAN(I,J) - State_Met%FRSEAICE(I,J)
       
-       ! Land and sea ice 
+       ! Land ice and sea ice
        FRICE = State_Met%FRLANDIC(I,J) + State_Met%FRSEAICE(I,J)
       
-       ! Land snow
+       ! Snow
        FRSNO = State_Met%FRSNO(I,J)
       
        ! Set IsLand, IsWater, IsIce, IsSnow based on max fractional area


lizziel:
All integration tests (GC-Classic and GCHP) passed. 1-month benchmark comparison of GC-Classic before and after the fix shows very small differences.

```
###############################################################################
### OH Metrics
### Ref = GCC-14.0; Dev = GCC-14.0+SfcTypeBugFix
###############################################################################

------------------------------------------------------------
Global mass-weighted OH concentration [10^5 molec cm^-3]
------------------------------------------------------------
Ref      : 11.97997512832
Dev      : 11.97999014605
Abs diff :  0.00001501774
%   diff :  0.000125

------------------------------------------------------------
CH3CCl3 (aka MCF) lifetime w/r/t tropospheric OH [years]
------------------------------------------------------------
Ref      :  5.153399
Dev      :  5.153397
Abs diff : -0.000002
%   diff : -0.000032

------------------------------------------------------------
CH4 lifetime w/r/t tropospheric OH [years]
------------------------------------------------------------
Ref      :  8.684106
Dev      :  8.684104
Abs diff : -0.000002
%   diff : -0.000018

```

msulprizio (approved):
This fix looks good to merge into 14.1.0.

Issue #1413, opened 2022-10-06
[FEATURE REQUEST] Retire met-field LWI in GC-Classic


lizziel:
This feature request is to (1) remove use of land-water-ice index (LWI) meteorology input in GC-Classic, and (2) remove the GEOS-Chem `State_Met%LWI`. 

The motivation for this is the land-water-ice index in the GEOS-Chem input meteorology will not be available in the GEOS-IT dataset. Updates have already removed the dependency in GCHP by using other input met-fields to determine if each grid cell is predominantly land, water, or ice. https://github.com/geoschem/geos-chem/blob/412ba5906b2aa9dcd72a2188c76c50220c3697b7/Interfaces/GCHP/Includes_Before_Run.H#L65-L74

There is only one location in GEOS-Chem that uses LWI for computation, in dry deposition. 
https://github.com/geoschem/geos-chem/blob/412ba5906b2aa9dcd72a2188c76c50220c3697b7/GeosCore/drydep_mod.F90#L2271-L2273

GEOS-Chem also passes LWI to HEMCO which uses it in several extensions. HEMCO will thus need to be updated to use the available land type met-fields to do this instead, and I'll make a separate feature request in HEMCO for this.

I will work on getting this update into 14.1. It will impact the benchmark but the impact should be minor.

yantosca:
For reference, here is the definition of LWI in GEOS-5. 

>**LWI**: A Land-Water-Ice mask provided for backward compatibility with the GEOS-4 product. It is 1 over continental areas, 0 over open ocean, and 2 over sea-ice covered ocean. Since in GEOS-5 a grid box can be a combination of these, continental areas are arbitrarily defined as those where FRLAND+FRLANDICE>=0.5. The remaining grid boxes are designated as sea-ice if the ice cover exceeds 50%; otherwise they are open (ice-free) ocean.

Note, this field was introduced in GEOS-5 for backwards compatibilty with the GEOS-4 `SURFTYPE` variable.  `LWI` is going to be retired from the newer GEOS-IT dataset as it can be created from a combination of other fields:

yantosca:
@lizziel, we probably shouldn't be too tied to GEOS-4.

lizziel:
I have a couple draft PRs almost ready for review. Integration tests are in progress.
https://github.com/geoschem/geos-chem/pull/1435
https://github.com/geoschem/HEMCO/pull/170

msulprizio:
[...]

PR #1435, opened 2022-10-11
Retire LWI met-field

lizziel:
**This PR is 1 of 2 for super-projects. See instructions here:**
https://github.com/geoschem/GCHP/issues/264
https://github.com/geoschem/GCClassic/issues/26

This pull request retires use of input meteorology field land-water-ice (LWI) index and the State_Met array it is stored in (`State_Met%LWI`). We are removing it as a dependency because it will not be included in the GEOS-IT dataset and determining whether a grid cell is over ocean, ice, or land is possible through other input meteorology fields. It is currently only used for computation in (1) dry deposition if using optional Luo wet deposition, and (2) HEMCO extensions.

Note that the LWI input field is currently computed during GEOS-Chem input pre-processing from GMAO meteorology constants FRLAND, FRLANDIC, FRLAKE, and FROCEAN, and time-varying fields FRSEAICE and FRSNO. Changing to using these fields at run-time will result in differences in the model. In addition, the new way for determining if mostly land, ice, or ocean evaluates the maximum fraction of land (including lake), ice (over both land and ocean), and ocean, which is slightly different than the algorithm for computing LWI. This will also cause differences.

HEMCO PR https://github.com/geoschem/HEMCO/pull/170 should be merged at the same time as this PR.

Closes https://github.com/geoschem/geos-chem/issues/1413


lizziel:
Results of a 1-month benchmark comarison with and without this update show differences in the land/ocean/ice boundaries, as shown here:
Screen Shot 2022-10-17 at 3 57 03 PM
Screen Shot 2022-10-17 at 3 56 51 PM

jimmielin:
Thanks for this PR. Tagging @fengx7. We will have to make corresponding changes in WRF-GC as well.

Interfaces/GEOS/geos_interface.F90
```diff
@@ -1271,10 +1270,18 @@ SUBROUTINE GEOS_Diagnostics( GC, IMPORT,  EXPORT, Clock, Phase, &
...
+                + State_Met%FRLAKE(I,J),                          &
+                State_Met%FRSEAICE(I,J),                          &
+                State_Met%FROCEAN(I,J) - State_Met%FRSEAICE(I,J)  &
+             /) )
+
              ! If there are HEMCO lightning emissions in current grid box set
              ! lightning potential accordingly
              IF ( ASSOCIATED(PtrEmis) ) THEN
-                IF ( LWI(I,J) == 1 ) THEN
+                IF ( SfcTypeIndex(1) == 1 ) THEN
```
msulprizio, line 1284: Do we need the `(1)` here? It looks like `SfcTypeIndex` is a scalar

Interfaces/GEOS/geos_interface.F90
```diff
@@ -1284,7 +1291,7 @@ SUBROUTINE GEOS_Diagnostics( GC, IMPORT,  EXPORT, Clock, Phase, &
 
              ! Lightning flash rate
              IF ( LFR(I,J) > 0.0 ) THEN
-                IF ( LWI(I,J) == 1 ) THEN
+                IF ( SfcTypeIndex(1) == 1 ) THEN
```
msulprizio, line 1294: Same question here

msulprizio (commented):
@lizziel These updates look good to merge. I just had one question denoted here.

Interfaces/GEOS/geos_interface.F90
yantosca: I think `SfcTypeIndex(1)` can just be reduced to `SfcTypeIndex`.  Functionally they are the same.

msulprizio: Thanks @yantosca. If there are no objections from @lizziel then I can make this change before merging.

lizziel: I think that's okay. I originally had it as vector. If it ends up being a problem I can take care of it when I test GEOS.

msulprizio: Thanks for confirming!

PR #1446, opened 2022-10-19
[FEATURE REQUEST] Add CHANGELOG.md

msulprizio:
This PR adds `CHANGELOG.md` to the geos-chem repository. This file follows the format outlined at https://keepachangelog.com/en/1.0.0/.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,42 +1,45 @@
 # Changelog
 
-All notable changes to this project will be documented in this file.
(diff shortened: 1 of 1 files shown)


lizziel (approved):
I made additions to the content, and did some editing. I added in the last version by name to be clear where the file starts, but we could remove that if others don't agree with it. I did the same in GC-Classic and GCHP wrapper CHANGELOGs. I also changed the tense of the verbs to past to match the section titles.

PR #1456, opened 2022-10-24
Improve GCHP crash error message in run scripts

lizziel:
This PR makes GCHP run script messages about GCHP crashing easier to understand. Previously the error message stated the problem in the run script, which was that file `cap_restart` either did not change or was empty. The underlying problem, however, is actually that GCHP failed to complete. The error message now states this with direction to look at the log file, which should be less confusing to users.


msulprizio (approved):
This is a straightforward update changing the GCHP error messages. Good to merge.

PR #1458, opened 2022-10-24
Point GCHP species to internal state and remove unused GCHP code

lizziel:
This commit removes allocation of `State_Chm%Species(N)%Conc` arrays to avoid duplication in memory when using GCHP. Species concentration arrays are now stored in one place only, in the MAPL internal state. To do this, the species values in the internal state are vertically flipped at the start and end of the GEOS-Chem `Run_` method. For safety, the GEOS-Chem species concentration array pointers are only associated during this time. This update remove the necessity of copy to/from the internal state species concentration arrays.

Please note that this PR does not include analogous updates for GEOS. Those updates will come in within a later pull request.

Other no-diff updates in this PR:
- Remove GCHP placeholder code for GEOS provider services
- Fix typo in GEOS pre-processor name
- Remove updating internal state in Finalize_ since redundant


lizziel:
All integration tests for GCHP and GC-Classic passed. I verified this is a zero diff update.

lizziel:
After discussion with @yantosca about performance and Fortran pointers, I decided to revisit doing a direct pointing to the internal state with the 3rd dimension ordered in reverse. Originally when I tried this I had run into an issue, but I am not seeing any issues now. I am doing a 1-month benchmark run now to compare with ref. If it is zero diff then I will push this update to include in 14.0.1.

christophkeller:
I wonder how we would make this work in GEOS, where the internal state fields are R4? I'm not sure if GEOS-Chem is ok with fp = f4 (I know that this is a compiler option, but does it produce reasonable results)? If we want to keep fp = f4, I don't see a simple solution other than keep making a copy.

lizziel:
I think we should be able to get it to work by setting R4 as the precision in a MODEL_GEOS block here (species_mod.F90):
```
  TYPE, PUBLIC :: SpcConc
#if defined( MODEL_GCHPCTM )
     REAL(ESMF_KIND_R8), POINTER :: Conc(:,:,:)
#else
     REAL(fp), POINTER :: Conc(:,:,:)
#endif
```

msulprizio (approved):
These updates look good to merge

PR #1473, opened 2022-10-31
Update Restart collection in HISTORY.rc to include BXHEIGHT and TROPLEV for all simulations

msulprizio:
The BXHEIGHT and TROPLEV fields are required in the GEOS-Chem restart file when generating global mass tables with GCPy. Previously, these fields were only enabled in the Restart collection for the fullchem and Hg simulations. We now include them for all GEOS-Chem Classic simulation types for consistency.


yantosca (approved):
This is a straighforward PR that adds 2 fields to each GCClassic restart file collection,  Approved and ready to merge!

PR #1480, opened 2022-11-03
Update CESM/HISTORY.rc

jimmielin:
Hi GCST,

This is a draft PR as @lizziel is still reviewing the **companion PR in CESM-GC:** https://github.com/CESM-GC/CAM/pull/11. Once the review is complete, this PR likely needs to be merged before the companion PR, so it will use the correct `HISTORY.rc`.

Because we will be adopting the GCHP HistoryExports component for CESM-GC, we need to write out the wildcards explicitly in HISTORY.rc similarly to GCHP.

Because not all processes in CESM-GC are handled within GEOS-Chem, the collections of processes not handled by GEOS-Chem routines, and as such diagnostics would be unavailable, are deleted from HISTORY.rc in CESM to reduce confusion. This includes drydep, wetdep, advection.

Thanks!
Haipeng


lizziel:
I am reviewing this update this week, testing with the other updates in CESM. I am going to change the target branch dev/14.0.2 since I it should be ready to go for that release and it does not impact offline GEOS-Chem or GCHP.

lizziel:
@jimmielin, there are lots of conflicts since your base is 14.1.0 dev. Do you need these updates to be on 14.1? I don't plan on using 14.1 in the ESCOMP/CAM PR for GEOS-Chem.

jimmielin:
Oops, sorry! Perhaps this commit could be cherry-picked to `14.0.1` base for inclusion in `dev/14.0.2`? Not sure how I would do this in my branch since it could mess up the history. But it's just one file change.

lizziel:
That works too. Once I bring it in I'll close this PR.

lizziel:
This PR is superceded by https://github.com/geoschem/geos-chem/pull/1509 so that it can go into 14.0.2.

PR #1509, opened 2022-11-16
GEOS-Chem diagnostics in CESM

lizziel:
This pull request contains Haipeng Lin's (@jimmielin) HISTORY.rc updates for use with CESM. I added changes on top of that.  It supercedes https://github.com/geoschem/geos-chem/pull/1480/ so that the update can go into 14.0.2 rather than 14.1. Here is the message from the original PR:

> Hi GCST,
> 
> This is a draft PR as @lizziel is still reviewing the companion PR in CESM-GC: https://github.com/CESM-GC/CAM/pull/11. Once the review is complete, this PR likely needs to be merged before the companion PR, so it will use the correct HISTORY.rc.
> 
> Because we will be adopting the GCHP HistoryExports component for CESM-GC, we need to write out the wildcards explicitly in HISTORY.rc similarly to GCHP.
> 
> Because not all processes in CESM-GC are handled within GEOS-Chem, the collections of processes not handled by GEOS-Chem routines, and as such diagnostics would be unavailable, are deleted from HISTORY.rc in CESM to reduce confusion. This includes drydep, wetdep, advection.
> 
> Thanks!
> Haipeng


run/CESM/HISTORY.rc
```diff
@@ -104,32 +97,6 @@ COLLECTIONS: 'Restart',
                               'OHwgtByAirMassColumnFull      ', 'GIGCchem',
```
yantosca, line 97: Do we need the 'GIGCchem' for CESM?  If so we can leave it.  We have since remeoved that from GEOS-Chem Classic HISTORY.rc templates.

run/CESM/HISTORY.rc
```diff
@@ -364,12 +315,199 @@ COLLECTIONS: 'Restart',
   JValues.frequency:          {FREQUENCY}
   JValues.duration:           {DURATION}
   JValues.mode:               'time-averaged'
-  JValues.fields:             'Jval_?PHO?                    ', 'GIGCchem',
-                              'JvalO3O1D                     ', 'GIGCchem',
-                              'JvalO3O3P                     ', 'GIGCchem',
-                              #'UVFluxDiffuse_?UVFLX?         ', 'GIGCchem',
-                              #'UVFluxDirect_?UVFLX?          ', 'GIGCchem',
-                              #'UVFluxNet_?UVFLX?             ', 'GIGCchem',
+  JValues.fields:             'JvalO3O3P              ', 'GCHPchem',
```
yantosca, line 318: Also do we need the GCHPchem here?

run/CESM/HISTORY.rc
```diff
@@ -421,7 +559,13 @@ COLLECTIONS: 'Restart',
   ProdLoss.frequency:         {FREQUENCY}
   ProdLoss.duration:          {DURATION}
   ProdLoss.mode:              'time-averaged'
-  ProdLoss.fields:            'Prod_?PRD?                    ', 'GIGCchem',
+  ProdLoss.fields:            'Prod_Ox                       ', 'GIGCchem',
```
yantosca, line 562: Also is GIGCchem needed here?

run/CESM/HISTORY.rc
```diff
@@ -654,45 +727,4 @@ COLLECTIONS: 'Restart',
                           'Met_BXHEIGHT                      ', 'GIGCchem',
```
yantosca, line 727: Also is GIGCchem needed here?

yantosca (changes requested):

yantosca (approved):
OK to merge!  Thanks @lizziel

run/CESM/HISTORY.rc
lizziel: Now removed.

run/CESM/HISTORY.rc
lizziel: Now removed.

PR #1512, opened 2022-11-16
Add fixes to restore the analytical inversion option for CH4 simulations

msulprizio:
The conversion from input.geos to geoschem_config.yml broke several of the switches for the CH4 analytical inversion options. To fix this, `ch4_simulation_options%` needed to be added to the key passed to QFYAML in order to read the correct value and pass it back to Input_Opt.


yantosca (approved):
This PR fixes a YAML input error (wrong search key) for the IMI/analytical inversion.  Good to merge.

PR #1513, opened 2022-11-16
Disallow wildcards in HISTORY.rc for CESM model

jimmielin:
Hi GCST, this is a simple change accompanying https://github.com/CESM-GC/CAM/pull/11 to disallow wildcards in HISTORY.rc when running within CESM.

We've adopted similar HistoryExports diagnostics to GCHP so when GCHP starts supporting this, the check can be removed for CESM as well. 

Thanks!
Haipeng


lizziel (approved):
Looks good.

PR #1519, opened 2022-11-18
Replace HEMCO time slice selection from 'I' to 'C' in CESM

lizziel:
This pull request replaces the time selection flags that are currently `I` (interpolation) in CESM `HEMCO_Config.rc` with `C` (cycle). Interpolation is done only for MODIS LAI and therefore this change is limited to that field. This update is a bug fix since the time interpolation feature of HEMCO currently does not work in CESM.


yantosca (approved):
This PR only changes the HEMCO_Config.rc file used by CESM and thus will be zero-diff for all other GEOS-Chem simulations.  Good to merge.

PR #1520, opened 2022-11-18
Turn off GEOS-Chem seasalt emissions in CESM

lizziel:
This pull request is a bug fix to turn off seasalt emissions in GEOS-Chem within CESM. Seasalt emissions are handled in CESM external to GEOS-Chem.


yantosca (approved):
This PR just toggles seasalt off in the HEMCO_Config.rc file used with CESM.  It will not affect other GEOS-Chem simulations (zero-diff).  Good to merge.

Issue #1523, opened 2022-11-19
[BUG/ISSUE] Invalid time:units values in Maasakers_Lakes*.nc files (CH4/carboncycle simulations)


yantosca:
While developing the carboncycle simulation in GCHP, I received this error message:

```console
pe=00001 FAIL at line=01980    ExtDataGridCompMod.F90  ./HcoDir/CH4/v2017-10/Lakes/Maasakkers_Lakes_$y4.01x01.nc has invalid reference date format>
```
So I looked at the `time:units` fields of the `Maasakers_Lakes*.nc` files.  I discovered that most of them have the wrong reference time.  

```console
$ for f in *.nc; do ls $f; ncdump -ct $f | grep "hours since"; done
Maasakkers_Lakes_2009.01x01.nc
                time:units = "hours since 2015-01-01 00:00:00" ;
Maasakkers_Lakes_2010.01x01.nc
                time:units = "hours since 2015-01-01 00:00:00" ;
Maasakkers_Lakes_2011.01x01.nc
                time:units = "hours since 2015-01-01 00:00:00" ;
Maasakkers_Lakes_2012.01x01.nc
                time:units = "hours since 2012-01-01 00:00:00" ;
Maasakkers_Lakes_2013.01x01.nc
                time:units = "hours since 2015-01-01 00:00:00" ;
Maasakkers_Lakes_2014.01x01.nc
                time:units = "hours since 2015-01-01 00:00:00" ;
Maasakkers_Lakes_2015.01x01.nc
                time:units = "hours since 2015-01-01 00:00:00" ;
```
This is not so much a problem for GCClassic but GCHP requires that the first time slice in the file equal the reference time.

I am not sure if these fields are still used for CH4/carboncycle.  If they are still needed then we should fix the timestamps with the `cdo setreftime` command.

Tagging @msulprizio

Also tagging @Jourdan-He and @SaptSinha for reference

msulprizio:
Thanks @yantosca. The CH4 emissions from lakes should be off by default in HEMCO_Config.rc because they likely double count emissions from wetlands. The option remains in case users want to investigate. However, this means we should comment out the lakes emissions file in ExtData.rc for GCHP simulations. 

Corrected files generated by @yantosca are now in `ExtData/HEMCO/CH4/v2022-11/Lakes`. 

PR #1526 addresses this issue and updates the CH4_LAKES path in HEMCO_Config.rc for CH4 simulations.

yantosca:
[...]

PR #1526, opened 2022-11-22
Update CH4 lake emission files to correct for wrong time units

msulprizio:
The original lake emission files in HEMCO/CH4/v2017-10 had the wrong time units as reported in https://github.com/geoschem/geos-chem/issues/1523. Here the CH4_LAKES file paths are updated to use the corrected files.


yantosca (approved):
This is a simple PR that updates the file path to corrected files with the proper time values for GCHP.  Good to merge!

Issue #1538, opened 2022-11-29
[BUG/ISSUE] EDGAR v6 CH4 rice emissions are missing


msulprizio:
In the CH4 simulation, rice emissions are extremely low starting in 13.4.0 when EDGAR v6 emissions were introduced (replacing EDGAR v4.3.2). 

```
###################################################################################
### Emissions totals for species CH4 [Tg]                                       ###
### Ref = CH4_13.3.3; Dev = CH4_13.4.0                                          ###
###################################################################################
                                    Ref                 Dev     Dev - Ref    % diff
CH4 BiomassBurn    :          19.412779           19.412779      0.000000     0.000
CH4 Coal           :          31.352809           30.415008     -0.937802    -2.991
CH4 Gas            :          24.919700           14.307448    -10.612253   -42.586
CH4 Lakes          :           0.000000            0.000000      0.000000       nan
CH4 Landfills      :          30.891649           36.385579      5.493931    17.785
CH4 Livestock      :         116.916247          120.381024      3.464778     2.963
CH4 Oil            :          42.265931           24.677758    -17.588174   -41.613
CH4 OtherAnth      :          24.629235           53.667464     29.038229   117.901
CH4 Rice           :          37.789759            0.480895    -37.308864   -98.727
CH4 Seeps          :           1.598907            1.598907      0.000000     0.000
CH4 SoilAbsorb     :         -33.631385          -33.631385      0.000000       nan
CH4 Termites       :          12.016941           12.016941      0.000000     0.000
CH4 Wastewater     :          37.746193           43.828258      6.082065    16.113
CH4 Wetlands       :         147.251804          147.251804      0.000000     0.000
-----------------------------------------------------------------------------------
CH4 Total          :         493.160531          470.792239    -22.368292    -4.536
```

It appears that the entry for CH4_RICE in HEMCO_Config.rc is missing in 13.4.0-14.0.1.

msulprizio:
It appears rice emissions are still included in 13.4.0+, however they're inadvertently lumped with the "Other Anthropogenic" sector. This also explains the increase in "CH4 OtherAnth" seen in the table above.

This confusion was caused by the change in file names from using `IPCC_4C_4D...` in EDGAR v4.3.2 to `AGS` (agricultural soils) in EDGAR v6:

HEMCO_Config.rc for 13.3.3:

```
### Rice ###
[...]

PR #1540, opened 2022-11-29
Restore CH4_RICE emissions in HEMCO_Config.rc.CH4

msulprizio:
CH4 rice emissions were previously read into HEMCO as field CH4_OTHER__4C_4D, when they should have been CH4_RICE__4C_4D and saved to HEMCO Cat=7 instead of 8. This error caused rice emissions to appear very low, when they were simply being counted in the other anthropogenic emissions sector. Here we restore the line for CH4_RICE__4C_4D in HEMCO_Config.rc for CH4 simulations and remove the lines for CH4_OTHER__4C_4D.

This addresses Github issue https://github.com/geoschem/geos-chem/issues/1538.


yantosca (approved):
These are simple modifications to the HEMCO_Config.rc file. Good to merge!

PR #1264, opened 2022-05-26
Add "SpeciesConcMND_?ADV?" diagnostic to output species concentrations in molec cm-3

jimmielin:
Hi GCST, this is the PR for the commit mentioned in https://github.com/geoschem/geos-chem/issues/1022. It may be necessary to rename this to `SpeciesMND_` along with renaming `SpeciesConc_` to `SpeciesVMR_`, as suggested by @msulprizio.

Thanks!

Best,
Haipeng

PR #1572, opened 2022-12-22
Add capability to save SpeciesConc diagnostic in units v/v or molec/cm3

msulprizio:
This PR supersedes https://github.com/geoschem/geos-chem/pull/1264.

The SpeciesConc diagnostic collection has been updated to save species concentrations in other units. The existing SpeciesConc variable has been renamed SpeciesConcVV and a new option SpeciesConcMND (units molec/cm3) has been added by @jimmielin.

This change also requires updating GCPy. See corresponding PR https://github.com/geoschem/gcpy/pull/187.


msulprizio:
This update causes differences in model output:

```
------------------------------------------------------------
Global mass-weighted OH concentration [10^5 molec cm^-3]
------------------------------------------------------------
Ref      : 12.41740390220
Dev      : 12.41739581682
Abs diff : -0.00000808537
%   diff : -0.000065
```

See also [GlobalMass_TropStrat.txt](https://github.com/geoschem/geos-chem/files/10289400/GlobalMass_TropStrat.txt).

I believe this is due to removal and/or moving of unit conversions in modules such as [GeosCore/diagnostics_mod.F90](https://github.com/geoschem/geos-chem/pull/1572/files).

GeosCore/diagnostics_mod.F90
```diff
@@ -133,6 +134,20 @@ SUBROUTINE Set_Diagnostics_EndofTimestep( Input_Opt,  State_Chm,           &
...
+    !-----------------------------------------------------------------------
+    ! Set species concentration for diagnostics in units of
+    ! molec/cm3 (hplin, 11/21/21)
+    !-----------------------------------------------------------------------
+    CALL Set_SpcConc_Diags_MND  ( Input_Opt,  State_Chm, State_Diag,         &
+                                  State_Grid, State_Met, RC                 )
+
+    ! Trap potential errors
+    IF ( RC /= GC_SUCCESS ) THEN
+       ErrMsg = 'Error encountered setting species concentration diagnostic 2'
```
yantosca, line 146: Suggest renaming error message to "Error encountered setting SpeciesConcMND diagnostic"

yantosca (changes requested):
Looks good to merge, with one update (error message update).  I did not see anything that would cause numerical drift (but if the ordering of unit conversions are changed then this might do it.

GeosCore/diagnostics_mod.F90
```diff
@@ -818,14 +809,115 @@ SUBROUTINE Set_SpcConc_Diags_VVDry( Input_Opt,  State_Chm, State_Diag,     &
...
+       ErrMsg = 'Incorrect species units in Set_SpcConc_Diags_MND!'
        CALL GC_Error( ErrMsg, RC, ThisLoc )
        RETURN
     ENDIF
 
-  END SUBROUTINE Set_SpcConc_Diags_VVDry
+    ! Convert to molec/cm3
+    CALL Convert_Spc_Units( Input_Opt, State_Chm, State_Grid, State_Met, &
+                            'molec/cm3', RC, OrigUnit=OrigUnit )
+
```
lizziel, line 893: The above call, along with the later one to convert back, will slightly change the values of species concentrations. I previously went through and removed all instances of using convert_spc_conc for diagnostics to avoid unnecessary changes. This should be deleted and replaced by inline conversation when setting the state_diag arrays.

GeosCore/diagnostics_mod.F90
```diff
@@ -818,14 +809,115 @@ SUBROUTINE Set_SpcConc_Diags_VVDry( Input_Opt,  State_Chm, State_Diag,     &
...
+       ! Point to mapping obj specific to SpeciesConcMND diagnostic collection
+       mapData => State_Diag%Map_SpeciesConcMND
+
+       !$OMP PARALLEL DO       &
+       !$OMP DEFAULT( SHARED ) &
+       !$OMP PRIVATE( N, S   )
+       DO S = 1, mapData%nSlots
+          N = mapData%slot2id(S)
+          State_Diag%SpeciesConcMND(:,:,:,S) = State_Chm%Species(N)%Conc(:,:,:)
```
lizziel, line 907: This is where the conversation should be. The equation is the same as used in unitconv_mod for the case of kg/kg -> molec/cm3, which is in subroutine ConvertSpc_KgKgDry_to_MND.

       ! Molecular weight for the species [kg]                                                    
       MW_kg = State_Chm%SpcData(N)%Info%MW_g * 1.e-3_fp

       ! Loop over grid boxes and do the unit conversion                                          
       DO L = 1, State_Grid%NZ
       DO J = 1, State_Grid%NY
       DO I = 1, State_Grid%NX
          State_Chm%Species(N)%Conc(I,J,L) =                             &
                                     State_Chm%Species(N)%Conc(I,J,L)    &
                                     * State_Met%AIRDEN(I,J,L)              &
                                     * ( AVO / MW_kg ) / 1e+6_fp

lizziel (changes requested):
Changes are needed to remove calling Convert_Spc_Conc and replace with direct computation from State_Chm%Species%Conc in kg/kg when setting the State_Diag array.

GeosCore/diagnostics_mod.F90
msulprizio: Thanks for pointing that out! I've pushed these changes to commit [1666335b](https://github.com/geoschem/geos-chem/commit/1666335bda77e786388cc19f214b4d203cbc3117).

GeosCore/diagnostics_mod.F90
msulprizio: Now fixed in [1666335b](https://github.com/geoschem/geos-chem/commit/1666335bda77e786388cc19f214b4d203cbc3117).

msulprizio:
After removing the call to `Convert_Spc_Units` (and converting units locally) as suggested by @lizziel during the review process, I now get zero differences with this update.

```
------------------------------------------------------------
Global mass-weighted OH concentration [10^5 molec cm^-3]
------------------------------------------------------------
Ref      : 12.41740390220
Dev      : 12.41740390220
Abs diff :  0.00000000000
%   diff :  0.000000
```

GeosCore/diagnostics_mod.F90
msulprizio: Thanks! This is now fixe in commit [1666335b](https://github.com/geoschem/geos-chem/commit/1666335bda77e786388cc19f214b4d203cbc3117)..

lizziel (approved):

PR #1582, opened 2023-01-06
Add missing J-value diagnostics to GCHP HISTORY.rc

msulprizio:
In 13.3.0, new species were added for the aromatic chemistry updates. These new photolysis species were automatically output in GCClassic which uses tokens, but needed to be explicitly added to HISTORY.rc for GCHP.


lizziel (changes requested):
Looks good. My only comment is reminder to update CHANGELOG.

msulprizio:
@lizziel Thanks for catching the CHANGELOG omission. I just pushed a commit for that.

lizziel (approved):

PR #1584, opened 2023-01-06
Add GCHP upwards mass flux diagnostic to config files

lizziel:
The upwards mass flux of air was added as a MAPL export in GCHP in 13.4.0. It is computed within the `GCHPctmEnv` gridded component by calling a subroutine in FV3 to retrieve it. That gridded component is used as an interface between GEOS-Chem and FV3 advection.

Users may be unaware this diagnostics exists or how to use it since the diagnostic is not part of the `GEOSchem` gridded component. This pull request solves this issue by adding it to the GCHP config files by default. It is now part of the `LevelEdgeDiags` collection for all simulations.


lizziel:
All GCHP integration tests passed on the branch.

lizziel:
I changed this to draft since I need to check if the output is vertically flipped.

lizziel:
I confirmed that the UpwardsMassFlux diagnostic is vertically flipped such that the level dimension is positive "up" (consistent with other GCHP diagnostics except emissions). This PR is now good to go.

Issue #1321, opened 2022-07-27
[BUG/ISSUE] Using "EF" CRE for Offline Emissions, QFED, etc. causes emissions to repeat each day


chfite:
### What institution are you from?
Florida State University

### Description of the problem
I have offline emissions turned on for BIOVOC, SEASALT, DUST, and SOILNOX (and their associated extensions turned off), but for each new day in my simulation the logged output does not mention that it is opening new Offline emission files for that day. I do not think those emissions are being released, and I think it may have to do with the "EF" CRE they now use in HEMCO_Config in newer model versions. However, I am not sure what is actually causing the bug. This bug causes the model to repeat those emissions from the first day of the simulation.

I'll note that I was also seeing this same issue with the QFED emissions, that also now use the "EF" CRE. All Offline and QFED emissions DID open on the first day of the simulation, but on following days they did not. Offline Lightning does open at the beginning of each new model date, but that is likely because it uses RFY3 instead of EF for its CRE.

### Description of troubleshooting performed
I had done a 12 month simulation before realizing this issue, and when I looked at the monthly mean HEMCO Diagnostic output the global mean offline emissions magnitudes were the same for each month.

I then did additional tests for only a 3 day period (20190901-00UTC to 20190904-00UTC), saving out daily mean HEMCO diagnostics. Again, each day of those simulations reported the same mean rates for offline emissions for the globe. I cross-checked these with the actual offline emission files in /data/ExtData/HEMCO/ and they DO vary in magnitude each day, so it wasn't an issue with the netcdf file inputs.

Finally, I did a test reverting back to using the "C" CRE like was used for QFED and Offline emissions in the past, and it fixed the issue, the files were found and opened at the beginning of the new dates, and the HEMCO diagnostic outputs varied daily as they should. I am not sure if this is an issue with the timestamps of the input files or if it's the code that handles an EF CRE that is the issue.

This sounds somewhat similar to https://github.com/geoschem/geos-chem/commit/b474612aa65d046aef22457955294bf509b620b2 but that issue was with boundary conditions using CRE's of EY (not EF), with the fix involving changing EY to EFY. Would a similar fix be needed here?

### GEOS-Chem version
I am using v13.3.4 complexSOA_SVPOA and GEOS-FP 2x2.5, however I did cherry-pick a few bug fixes/updates relevant to me that were added in 13.4.0. I have attached a patch file that should describe any changes I made.

### Description of modifications
Here is the attached patch files with my changes to v13.3.4:
[...]

PR #1595, opened 2023-01-11
Fix timecycle flag for QFED and OFFLINE emissions in HEMCO_Config.rc

msulprizio:
As discussed in https://github.com/geoschem/geos-chem/issues/1321, the current use of 'EF' as the timecycle (CRE) flag for QFED and the OFFLINE emissions inventory causes those fields to only be read/updated once. Changing those to EFY ensures those fields are updated continuously.


yantosca (approved):
This PR simply replaces EF where it is used with EFY.  Good to merge!

PR #1598, opened 2023-01-12
Update CH4 simulation to use loss rates from GEOS-Chem 14.0.0 10-year benchmark

msulprizio:
The CH4_LOSS field was previously obtained from output of a GMI simulation. Here we update those fields to use CH4 loss rates from the 10-year benchmark simulation for GEOS-Chem 14.0.0. The code in global_ch4_mod.F90 has also been modified to reflect a change in units for this field from s-1 to molec/cm3/s.


msulprizio:
Todd Morring has generated loss frequencies from the 10-year benchmarks instead of the loss rates implemented here. This is for consistency with the existing GMI loss frequencies. Testing is ongoing and a separate PR will be submitted for those updates. I will close out this PR.

PR #1638, opened 2023-02-03
Add reminders to compile with CMake options to createRunDir.sh scripts

yantosca:
This PR prints out reminders to compile with certain CMake options at the end of the GCClassic and GCHP run directory creation process.  For example, if you were creating a GCClassic or GCHP run directory for a TOMAS15 simulation, you would see this text:

```console
... etc. not shown ...
-----------------------------------------------------------
Do you want to track run directory changes with git? (y/n)
-----------------------------------------------------------
n

>>>> REMINDER: You must compile with options: -DTOMAS=y -DTOMAS_BINS=15 <<<<

Created /path/to/tomas/run/dir
```

The reminders are printed just before the `createRunDir.sh` script stops executing so that the options remain prominent and visible to the user.

I have labeled this as a bug fix because reminders had been printed for some simulations but not for all relevant simulations, as they are now.

### DIFF
--- a/run/GCClassic/createRunDir.sh
+++ b/run/GCClassic/createRunDir.sh
@@ -881,10 +881,6 @@ fi
 ln -s ${wrapperdir} ${rundir}/CodeDir
 ln -s ${wrapperdir}/run/GCHP/runScriptSamples ${rundir}/runScriptSamples
 
-# Create build directory
-mkdir ${rundir}/build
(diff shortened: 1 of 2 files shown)


lizziel (approved):
Looks good to merge.

msulprizio:
@yantosca @lizziel  Question before you merge: Do we also print this information to the CMake instructions in the build/README file? If not, it might be useful to add a note there too.

yantosca:
@msulprizio @lizziel: I modified the `createRunDir.sh` scripts so it also writes the extra CMake commands to the `build/README` file.  See commit c48acc380

msulprizio (approved):
Looks great. Thanks @yantosca!

PR #1644, opened 2023-02-07
Fix bug where writing species metadata yaml file always attempted

lizziel:
This PR fixes a bug in which the model crashes if `species_metadata_output_file` is excluded from the `simulation` settings in `geoschem_config.yml`. It does this by changing the criteria to skip file write. Previously the criteria was that the output file path was equal to 'none'. This did not work since the output file path is initialized to an empty string. The new criteria is that the string length is greater than zero. If the setting is excluded from the config file it will no longer attempt to print. 

I also edited the comments and removed the CESM ifdefs from the section of code in which the yaml file write is called. WRF and CESM should not write the yaml file and the comments indicate that. Rather than prevent it via pre-processor switch the file path can be excluded from the `geoschem_config.yml` file as it is now.


yantosca (approved):
Thanks @lizziel!  OK to merge.

PR #1653, opened 2023-02-13
GEOS-only updates from Christoph Keller

lizziel:
Updates in this pull request impact GEOS-Chem only when run within GEOS. All updates were initially made by Christoph Keller. Changes include:

- Add kg/kg total <-> molec/cm3 conversion option to Convert_Spc_Units
- Add carbon-related State_Chm fields (GEOS only)
- Add new GEOS-only diagnostics for boundary layer column density of species (PblCol), relative change of CO due to CO2 photolysis (COincCO2phot), and CO2 photolysis rate (CO2photRrate)
- Change run config files in Interfaces/GEOS add new file geos_CarbonInterface.F90
- Change GEOSCHEMchem_GridCompMod for replay and carbon (GEOS-only)

This will be a zero diff update in the GEOS-Chem benchmark and can therefore go into 14.1.1.


yantosca (approved):
These are zero-diff changes that will only get activated when running GEOS-Chem in GEOS.  Good to merge.

PR #1666, opened 2023-02-23
Modify HEMCO_Config.rc.carbon templates for single-species execution

yantosca:
This PR (currently a work in progress) contains modifications to the carbon simulation code and/or configuration files in order to allow the simulation to run with only a single advected species (i.e. advect CH4, and consider CO, CO2, OCS as non-advected chemistry-only species).

I have placed the target as 14.1.1 but depending on extra work/validation needed, this might have to go into 14.2.0.


yantosca:
@msulprizio I think we can try to get this into 14.2.0.  I wouldn't want to hold up the 14.1.1 release.

PR #1677, opened 2023-02-28
Move GC-Classic run scripts to operational_examples directory

lizziel:
The PR moves existing example scripts to run GEOS-Chem Classic to a new directory in the `operational_examples` folder. Previously these scripts were in the `runScriptSamples` directory but not in a folder indicating they are custom for the Harvard Cannon cluster. These updates include changing the run directory creation script to copy from the new location if creating a benchmark simulation run directory.


msulprizio (approved):
This is a straightforward update. Looks good to merge.

PR #1681, opened 2023-03-02
Remove references to Compute Canada from data download scripts

msulprizio:
The ComputeCanada site hosting GEOS-Chem data has been retired. References to it have been removed from the download data scripts and yaml files.


yantosca (approved):
This PR removes an obsolete option from YAML config files and updates comments in the `download_data.py` script.  All of the existing server metadata is still present.  Good to merge.

PR #1686, opened 2023-03-06
Fix computation of planeCount for Harvard Cannon GCHP run scripts and integration test scripts

### LINKED ISSUE #1685: Possible error in Cannon run script
In the Harvard Cannon GCHP run script, there's a line which I think is meant to try and modify the core distribution: https://github.com/geoschem/geos-chem/blob/4722f288e90291ba904222f4bbe4fc216d17c34a/run/GCHP/runScriptSamples/operational_examples/harvard_cannon/gchp.run#L34

However, this line actually ends up being a comparison, as the left side is formatted to return (not set) the variable. This can be easily fixed as 

`planeCount=$(( ${planeCount} + 1 ))`

This may well have been a bug I introduced in the first place, so hopefully it hasn't caused others too many issues!

yantosca:

Name: Bob Yantosca
Institution: Harvard / GCST

### Describe the update

This fixes an error in the computation of the plane count.  This is passed as an argument to the `srun` command in order to run GCHP on the Harvard Cannon cluster with OpenMPI.

### Expected changes

This will increment the plane count by 1 in the rare event that the core count is not evenly divisible by the number of nodes. 

### Reference(s)

N/A

### Related Github Issue(s)

This issue was raised by @sdeastham in #1685.  This PR closes #1685.

### DIFF
--- a/run/GCHP/runScriptSamples/operational_examples/harvard_cannon/gchp.run
+++ b/run/GCHP/runScriptSamples/operational_examples/harvard_cannon/gchp.run
@@ -28,9 +28,11 @@ export OMPI_MCL_btl=openib
 # Run GCHP and evenly distribute tasks across nodes
 NX=$( grep NX GCHP.rc | awk '{print $2}' )
 NY=$( grep NY GCHP.rc | awk '{print $2}' )
-let coreCount=$(( ${NX} * ${NY} ))
-let planeCount=$(( ${coreCount} / ${SLURM_NNODES} ))
-[[ $(( ${coreCount} % ${SLURM_NNODES} )) > 0 ]] && let planeCount+=1
+coreCount=$(( ${NX} * ${NY} ))
+planeCount=$(( ${coreCount} / ${SLURM_NNODES} ))
+if [[ $(( ${coreCount} % ${SLURM_NNODES} )) > 0 ]]; then
+    planeCount=$(( ${planeCount} + 1 ))
+fi
 time srun -n ${coreCount} -N ${SLURM_NNODES} -m plane=${planeCount} --mpi=pmix ./gchp >> ${log}
 
 # Rename and move restart file and update restart symlink if new start time ok
--- a/run/GCHP/runScriptSamples/operational_examples/harvard_cannon/gchp.run
+++ b/run/GCHP/runScriptSamples/operational_examples/harvard_cannon/gchp.run
@@ -28,11 +28,9 @@ export OMPI_MCL_btl=openib
 # Run GCHP and evenly distribute tasks across nodes
 NX=$( grep NX GCHP.rc | awk '{print $2}' )
 NY=$( grep NY GCHP.rc | awk '{print $2}' )
-coreCount=$(( ${NX} * ${NY} ))
-planeCount=$(( ${coreCount} / ${SLURM_NNODES} ))
-if [[ $(( ${coreCount} % ${SLURM_NNODES} )) > 0 ]]; then
-	${planeCount}=$(( ${planeCount} + 1 ))
-fi
+let coreCount=$(( ${NX} * ${NY} ))
+let planeCount=$(( ${coreCount} / ${SLURM_NNODES} ))
+[[ $(( ${coreCount} % ${SLURM_NNODES} )) > 0 ]] && let planeCount+=1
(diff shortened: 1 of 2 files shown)


run/GCHP/runScriptSamples/operational_examples/harvard_cannon/gchp.run
```diff
@@ -28,11 +28,9 @@ export OMPI_MCL_btl=openib
 # Run GCHP and evenly distribute tasks across nodes
 NX=$( grep NX GCHP.rc | awk '{print $2}' )
 NY=$( grep NY GCHP.rc | awk '{print $2}' )
-coreCount=$(( ${NX} * ${NY} ))
-planeCount=$(( ${coreCount} / ${SLURM_NNODES} ))
-if [[ $(( ${coreCount} % ${SLURM_NNODES} )) > 0 ]]; then
-	${planeCount}=$(( ${planeCount} + 1 ))
-fi
+let coreCount=$(( ${NX} * ${NY} ))
```
lizziel, line 31: Is adding let necessary? It looks like otherwise the line is the same, and I don't think this was the line causing the issue. Same comment for the next line (planeCount).

run/GCHP/runScriptSamples/operational_examples/harvard_cannon/gchp.run
```diff
@@ -28,11 +28,9 @@ export OMPI_MCL_btl=openib
...
+let planeCount=$(( ${coreCount} / ${SLURM_NNODES} ))
+[[ $(( ${coreCount} % ${SLURM_NNODES} )) > 0 ]] && let planeCount+=1
```
lizziel, line 33: This is a bit hard to parse and understand. Could we stick with old school style and do planeCount=$(( ${planeCount} + 1 ))?

lizziel (changes requested):
My comments are just on the style. The updates use a style not used in any of our other run scripts and I wonder if it will make it harder for people to understand/adapt.

I'm putting a bunch of zero diff GCHP fixes into a new branch. We should merge this in at the same time as that.

yantosca:
Hi @lizziel.  So the `let` command in bash specifically refers to an arithmetic operation.  I thought it was a bit cleaner and less ambiguous than the other commands.

TL;DR: https://phoenixnap.com/kb/bash-let

lizziel:
From the tutorial it looks like the usage would be something like this:
`let "coreCount = NX * NY"`

As is, it looks like the old lines are still there but prefixed with let. Am I missing something?

yantosca:
@lizziel, maybe not.  We can definitely use let "coreCount = NX * NY" if that's easier.

lizziel:
Yeah, I'm thinking if we switch to using let then best use it to reduce number of characters. But I also wonder if doing that should be in a separate PR when we have time to go through all scripts for consistency, e.g. in setCommonRunSettings.sh where there is a log of unix arithmetic. Maybe this PR should just be the one line fix for planeCount?

msulprizio:
@yantosca @lizziel Is this PR ready for merge? Were the comments above resolved?

yantosca:
@msulprizo, I wasn't able to resolve the comments yet.  I can do that now.

yantosca:
The rerun of GCHP integration tests passed except for tagO3.  TagO3 will be fixed in #1724.
```console
GCHP: Execution Test Results

GCClassic #4734a35 MAPL and GEOS-Chem submodule updates: GCHP bug fixes for 14.2.0
GEOS-Chem #5d12fbdbe Revert to not using "let" in gchp.run and integrationTestExecute.sh
HEMCO     #a4f8517 Merge PR #189 (branch 'feature/add-lumped-furans') into dev/3.7.0

Number of execution tests: 5

Submitted as SLURM job: 48552102

Execution tests:
------------------------------------------------------------------------------
gchp_merra2_fullchem................................Execute Simulation....PASS
gchp_merra2_fullchem_benchmark......................Execute Simulation....PASS
gchp_merra2_fullchem_RRTMG..........................Execute Simulation....PASS
gchp_merra2_tagO3...................................Execute Simulation....FAIL
gchp_merra2_TransportTracers........................Execute Simulation....PASS
 
Summary of test results:
------------------------------------------------------------------------------
Execution tests passed: 4
Execution tests failed: 1
Execution tests not yet completed: 0
```

lizziel (commented):
Looks good!

Issue #1412, opened 2022-10-05
[FEATURE REQUEST] Automatically rename GCHP checkpoint files to GEOSChem.Restart.YYYYMMDD_hhmmz.nc4


msulprizio:
When enabling`Periodic_Checkpoint` in setCommonRunSettings.sh for GCHP 14.0.0, these files are automatically saved to the top-level run directory as gcchem_internal_checkpoint.YYYYMMDD_hhmmz.nc4. However, when restarting a simulation, GCHP expects restart files to be in Restarts/ with filename GEOSChem.Restart. YYYYMMDD_hhmmz.cXX.nc4. Users therefore need to manually move and rename those files or create a symbolic link within Restarts/ pointing to the checkpoint files. To remove this burden, we should update the sample run scripts to automatically do this at the end of each run. That won't however resolve this issue if a run crashes unexpectedly.

lizziel:
We should be careful here for the case of creating a checkpoint at the start of the run. If moving it to Restarts then the original would be over-written. Also, if there is a bug in the periodic checkpoints (it uses different code in MAPL to create them) then this may cause problems in the restarts.

lizziel:
Thinking more on this issue, I wonder if the complication and risk of renaming and moving the mid-run checkpoints to Restarts is worth the usual use case, which is to use one of the files to restart mid-run upon crashing. For that use case care has to be taken to figure out what went wrong and decide which date to start over from. It might be worth forcing users to manually copy or link the checkpoint file in Restarts. This would also help keep the run scripts short, which makes them easier to understand.

For 1-year benchmarking, which will always require using multiple checkpoints to make mass tables, we could create a special benchmark run script that does copy and rename. Another option would be to create an instantaneous mass diagnostic collection for the 1-year benchmarks, but that would cause complications in GCPy since we don't do that for GC-Classic (although we could). But if a user asked how to get instantaneous species mass at a fixed frequency I would probably recommend creating a collection, not use the restart file which has an atypical format, is vertically flipped, and has a bunch of other stuff in it.

@sdeastham, do you have opinion on this?

lizziel:
Actually, it just occurred to me we do using a collection to instantaneous mass for GC-Classic. Maybe we should do this for GCHP too. Currently monthly frequency is only possible for time-averaged collections, but I would think extending that to instantaneous collections would be doable.

sdeastham:
[...]

PR #1717, opened 2023-03-22
Use GCARR_ac rate law function for full-chemistry reactions where the Arrhenius B parameter is zero

yantosca:
### Name and Institution (Required)

Name: Bob Yantosca
Institution: Harvard + GCST

### Confirm you have reviewed the following documentation

- [x] [Contributing guidelines](https://geos-chem.readthedocs.io/en/stable/reference/CONTRIBUTING.html)

### Describe the update
This PR addresses the issue raised in #1641.  Recent chemistry updates had brought in several reactions that used the rate-law function `GCARR_abc`, even though the Arrhenius 'B` parameter is zero for these reactions.  This results in raising 0 to a power, which will always evaluate to 1, causing useless computational overhead for the same result.

We have replaced these calls to `GCARR_abc` with calls to `GCARR_ac`.  This will produce the same result but skip the useless computations.  This should result in faster execution.

Also, we now point users to `KPP/fullchem/CHANGELOG_fullchem.md` for a list of updates tot the fullchem mechanism.

### Expected changes
This is expected to be a zero-diff update.

### Reference(s)
N/A

### Related Github Issue(s)
Closes #1641

PR #1724, opened 2023-03-24
GCHP bug fixes and run directory updates

lizziel:

Name: Lizzie Lundgren

### Describe the update

This PR includes the following GCHP updates:

1. Example grid labels are now commented out in all HISTORY.rc files by default. This avoids a decomposition problem if running with over 3000 cores since grids are created for all uncommented grid labels regardless of whether they are used in a collection.

2. The interactive run script now prints error messages to screen in addition to the log, and also stops if an error is encountered. Previously some error messages were only written to log and the script kept going if an error was encountered, such as a non-existent restart file.

3. Simulations that require all species in the GC-Classic restart file, e.g. standard fullchem, now also require all species in the GCHP restart file. A new toggle in `setCommonRunSettings.sh` allows changing this behavior. Like GC-Classic, all simulations allow missing species during integration testing.

4. Overwriting existing GCHP diagnostic files is allowed again after not working since 14.0 when MAPL was updated. The behavior is now controlled from a switch at the top of `HISTORY.rc`.

5. All GCHP checkpoint files are now written directly to the `Restarts` folder in the run directory. This avoids writing large files locally if necessary since the `Restarts` folder can be symbolically linked to a different drive.

6. All GCHP checkpoint files are now renamed to the standard GEOS-Chem restart filename format, with the exception of the checkpoint file for the start time. That file is deleted so as not to overwrite the original restart file of the run.

7. GCHP tagged ozone simulation is functional again after fixing a bug in the prod/loss input filename in `ExtData.rc`.

8. Change GCHP carbon simulation inputs for global Cl and P(CO) to data generated from the 14.0.0 10-year benchmark.

### Expected changes

This update is zero diff for for all simulation diagnostics and restart files with the exception of the carbon simulation which now uses updated global Cl and P(CO data.

All other changes are related to run directory behavior. It is now possible to:
1. Allow overwriting diagnostics files
2. Restart GCHP directly from an output checkpoint file generated in the middle of a run
3. Run GCHP with >3000 cores
4. Better detect run directory problems such as missing restart file
5. Force the model to crash if a species is missing from the restart file
6. Write all restart files directly to the Restarts subdirectory

### Reference(s)

None

### Related Github Issue(s)

Closes https://github.com/geoschem/GCHP/issues/303
Closes https://github.com/geoschem/GCHP/issues/300
Closes https://github.com/geoschem/GCHP/issues/296
Closes https://github.com/geoschem/GCHP/issues/295
Closes https://github.com/geoschem/geos-chem/issues/1412
https://github.com/geoschem/geos-chem/pull/1717

### DIFF
--- a/run/GCHP/ExtData.rc.templates/ExtData.rc.carbon
+++ b/run/GCHP/ExtData.rc.templates/ExtData.rc.carbon
@@ -455,22 +455,21 @@ GFED_FRAC_DAY 1 N Y %y4-%m2-%d2T00:00:00 none none GFED_FRACDAY ./HcoDir/GFED4/v
 TIMEZONES count N V - none none UTC_OFFSET ./HcoDir/TIMEZONES/v2015-02/timezones_voronoi_1x1.nc
 
 # --- OH from the latest 10-year benchmark ---
-# ** TODO: Update to 14.0.0 **
-#GLOBAL_OH kg/m3 N N %y4-%m2-01T00:00:00       none none SpeciesConc_OH ./HcoDir/GCClassic_Output/13.0.0/%y4/GEOSChem.SpeciesConc.%y4%m2%d2_0000z.nc4
+#GLOBAL_OH kg/m3 N N %y4-%m2-01T00:00:00       none none SpeciesConc_OH ./HcoDir/GCClassic_Output/14.0.0/%y4/GEOSChem.SpeciesConc.%y4%m2%d2_0000z.nc4
 
 # --- OH from GEOS-Chem v5-07-08 ---
 # --- NOTE: Use this for CH4/IMI ---
 #GLOBAL_OH kg/m3 1985 Y F%y4-%m2-01T00:00:00 none none OH ./HcoDir/OH/v2022-11/v5-07-08/OH_3Dglobal.geos5.47L.4x5.nc
 
 # --- Global Cl concentrations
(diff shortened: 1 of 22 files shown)


lizziel:
Integration tests pass. Difference tests in progress.

yantosca (approved):
These updates all look good.  OK to merge

run/GCHP/ExtData.rc.templates/ExtData.rc.carbon
```diff
@@ -456,21 +456,21 @@ TIMEZONES count N V - none none UTC_OFFSET ./HcoDir/TIMEZONES/v2015-02/timezones
 
 # --- OH from the latest 10-year benchmark ---
 # ** TODO: Update to 14.0.0 **
```
msulprizio, line 458: Can we remove these TODO lines since this is now done?

msulprizio (changes requested):
I've requested one small change in the comments for `ExtData.rc.templates.carbon`.

msulprizio (approved):

yantosca (approved):

lizziel:
I just realized this update will require an update to the gc-cloud-infrastructure in in this [file](https://github.com/geoschem/gc-cloud-infrastructure/blob/main/benchmarks/scripts/stages/runGCHP.sh) since the restart file is now written directly to Restarts rather than being moved there. I will make a separate PR for that. Is this going to cause problems with benchmarking?

msulprizio:
> Is this going to cause problems with benchmarking?

@lizziel I don't believe this will cause an issue with benchmarking. I will try to merge both concurrently to avoid issues.

lizziel:
This PR should be merged at the same time as https://github.com/geoschem/gc-cloud-infrastructure/pull/44 to avoid cloud benchmarking issues.

PR #1728, opened 2023-03-28
Add RxnConst diagnostic for archiving reaction rate constants

jimmielin:

Name: Haipeng Lin
Institution: Harvard Univ

### Describe the update

This pull request implements the RxnConst_?RXN? diagnostic, allowing for archival of reaction rate constants (units vary) from KPP.

It is not enabled by default but can be enabled in a similar manner to the existing RxnRate_ diagnostics, which output the reaction rates (in molec cm-3 s-1) which are the rate constants with the species concentrations already applied.

Minor cosmetic fixes to the comments and metadata are added to describe the correct units of the RxnRate diagnostic (molec cm-3 s-1) as well.

### Expected changes

No changes to output are expected. These changes add a new diagnostic that is not enabled by default.

### Reference(s)

N/A

### DIFF
--- a/GeosCore/fullchem_mod.F90
+++ b/GeosCore/fullchem_mod.F90
@@ -267,6 +267,7 @@ CONTAINS
     IF (State_Diag%Archive_ProdCOfromNMVOC) State_Diag%ProdCOfromNMVOC= 0.0_f4
     IF (State_Diag%Archive_OHreactivity   ) State_Diag%OHreactivity   = 0.0_f4
     IF (State_Diag%Archive_RxnRate        ) State_Diag%RxnRate        = 0.0_f4
+    IF (State_Diag%Archive_RxnConst       ) State_Diag%RxnConst       = 0.0_f4
     IF (State_Diag%Archive_SatDiagnRxnRate) State_Diag%SatDiagnRxnRate= 0.0_f4
     IF (State_Diag%Archive_KppDiags) THEN
        IF (State_Diag%Archive_KppIntCounts) State_Diag%KppIntCounts   = 0.0_f4
@@ -977,7 +978,7 @@ CONTAINS
        !=====================================================================
        ! HISTORY (aka netCDF diagnostics)
        !
-       ! Archive KPP reaction rates [s-1]
+       ! Archive KPP reaction rates [molec cm-3 s-1]
        ! See gckpp_Monitor.F90 for a list of chemical reactions
        !
        ! NOTE: In KPP 2.5.0+, VAR and FIX are now private to the integrator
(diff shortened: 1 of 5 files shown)


lizziel:
Thanks @jimmielin. Since the units are "various", how are users supposed to know the units for the diagnostics? I don't think we have any other diagnostics that have unknown units. Maybe there's a way to make this work.

Also, we should add a new collection for this diagnostic since we have collections for the other reaction rates. We like to always add new State_Diag arrays to template HISTORY.rc files so that we can show users all of their diagnostic options and how to use them.

jimmielin:
Hi @lizziel, good question about the units! There's two potential hurdles here. 

The first is that I'm not sure if the netCDF diagnostics support adding a different unit for a different wildcard entry - e.g., EQ001 might have units of molecules-1 cm-3 s-1, but EQ002 might have molecules-2 cm-6 s-1. 

The second is that I don't think we have the capability to figure out the unit of the rate constant programmatically - it can be figured out from reading `fullchem.eqn` (or `gckpp_Function.F90`) and seeing how many reactants are there for each reaction, but this is not readily accessible in code. This is why I kept the units as "various" which is not very good but it wasn't readily fixable.

jimmielin:
Thanks @yantosca. I just looked at how CAM-chem implements this. It appears that they count the number of reactants for each equation to correctly assign units:

```fortran
          select case(num_rnts(ii))
          case(1)
             unitstr='/s'
          case(2)
             unitstr='cm3/molecules/s'
          case(3)
             unitstr='cm6/molecules2/s'
          case default
             call endrun('gas_phase_chemdr_inti: invalid value in num_rnts used to set units in reaction rate constant')
          end select
```

I wonder if we could do something similar with KPP 3.0.0's new feature, `gckpp_Function.F90` includes this subroutine called `CalcStoichNum()` which returns `StoichNum(NVAR, NREACT)`. For each tagged reaction counting the amount of >0 entries for `StoichNum(:, reaction_number)` we could determine the number of reactants.

It might be too much hassle. I'm happy to keep it as `s-1` with a caveat in the documentation that some reactions may have different units.

jimmielin:
Actually I could name the units `(cm3 molec-1)**nreact s-1` which would then be accurate to all variables. For the user who is looking for a specific reaction they would know `nreact` and the unit indication would have an obvious meaning for them.

yantosca:
@jimmielin: I like that unit description.

jimmielin:
Thanks @yantosca! I've updated the PR with the correct unit description and added collections to `HISTORY.rc` in GCClassic and GCHP.

jimmielin:
> @jimmielin, based on the CESM code it looks like units would be `(cm3 molec-1)**(nreactants - 1) s-1` . Am I misunderstanding?

D'oh! My bad. Fixed now, thanks for noticing this!

yantosca (approved):
Looks good!

jimmielin:
Hi all, just bumping this so stalebot doesn't close it. Thanks! Haipeng

lizziel:
@msulprizio, would we be able to get this into 14.2?

msulprizio:
@lizziel Yes, I'll merge it in now

Issue #1646, opened 2023-02-08
[BUG/ISSUE] Restore vertical allocation of CEDS emissions


msulprizio:
@1Dandan wrote:

>I noticed there was a github feature request (https://github.com/geoschem/HEMCO/issues/46) about the vertical allocation of energy, non-solvent industry, and ship emissions. When I browse the issue, it looks like it was implemented in GEOS-Chem v13.1 for CEDS_GBD-MAPS as suggested, but when I check versions of 13.2 and 13.4 using CEDSv2 as global emission inputs, the feature of vertical allocation is deprecated for all energy, industry and ship emissions by default.

msulprizio:
@1Dandan @barronh This is indeed a bug. When we updated to CEDSv2 the scale factors applying the vertical allocation of those emissions were inadvertently dropped. We can restore them in 14.2.0.

msulprizio:
The pull requests (#1712 and #1730) resolving this issue have now been merged into the 14.2.0 development branch.

barronh:
You rock!

PR #1730, opened 2023-03-31
Add fix for CEDS energy and industry emissions to distribute vertically

msulprizio:

Name: Melissa Sulprizio
Institution: Harvard/ GCST

### Describe the update

In HEMCO_Config.rc the dimensions for the CEDS ENE and IND entries needed to be changed from xy to xyL* to properly distribute the emissions vertically. From the original implementation in commit 87eb68ee:

>These fields are 2D in the CEDS files, but we now can specify dimension xyL* instead of xy to tell HEMCO to copy the field into each emissions level... Scale factors can then be applied to determine how much emissions to apply to each level.

### Expected changes

This fix should resolve the decrease in CEDS emissions observed after implementing PR #1712 by properly allocating emissions above the first level.

### Related Github Issue(s)

- https://github.com/geoschem/geos-chem/issues/1646


yantosca (approved):
This PR looks good to merge.... for CEDS ENE and IND sectors, we have to change "xy" to "xyL*" so that it will do the vertical allocation.

PR #1733, opened 2023-03-31
Restructure photolysis

lizziel:
### Name and Institution (Required)

Name: Lizzie Lundgren
Institution: Harvard University

### Confirm you have reviewed the following documentation

- [X] [Contributing guidelines](https://geos-chem.readthedocs.io/en/stable/reference/CONTRIBUTING.html)

### Describe the update

This PR contains structural updates that facilitate easily dropping in a different photolysis rate calculation scheme. Below is a summary of all changes:

1. Move much of `CMN_FJX_Mod` to new object `State_Chm%Phot` object. The only parts of `CMN_FJX_Mod` that remain are used only in core Fast-JX. `CMN_FJX_Mod` will be deleted when Fast-JX is retired and replaced by Cloud-J.
2. Abstract code in `Fast_JX_Mod` such that only core Fast-JX remains in the module. Core Fast-JX is analogous to code in Cloud-J and does not include GEOS-Chem-specific code except for a couple minor exceptions, such as diagnostics for GEOS.
3. Create new module `FJX_Interface_Mod` to serve as a wrapper around core Fast-JX
4. Create new module `Photolysis_Mod` to serve as a generic module for computation of J-values and related diagnostics.
5. Expand `geoschem_config.yml` photolysis menu to include an activate toggle to turn photolysis rate computation on/off. If off then J-values will be all zero.
6. Add new `Input_Opt` variable `Do_Photolysis`.
7.  Change `geoschem_config.yml` such that Fast-JX tables directory are specified in a new subsection for photolysis input directories. This sets up being able to add a second input directory for Cloud-J.
8. Add speed of light and Plank's constant to the `PhysConstants` module. These constants are used for computation of UV Flux diagnostics.
9. Remove "Fast-JX" from photolysis timer names so that the names are more generic.
10. Replace subroutine `Init_Chemistry` with `Init_Photolysis` since photolysis was the only initialization contained in the subroutine.
12. Retire unused module Set_Prof_O3.
13. Move UVFlux diagnostics out of `JValues` collection and into their own collection called `UVFlux`.

### Expected changes

This is a no diff update, with the exception of UVFlux diagnostics which have diffs on the order of 1e-30. Those differences are due to adding `_fp` to the values of Planck's constant and the speed of light, thereby preventing numerical noise introduced in a conversion from single to double precision.

### Related Github Issue(s)

This update is in support of developing Cloud-J in GEOS-Chem. 
https://github.com/geoschem/geos-chem/issues/953

PR #1750, opened 2023-04-12
Bugfix/photolysis off option

lizziel:

Name: Lizzie Lundgren

### Describe the update

This update fixes a bug introduced in 14.2.0 in the photolysis `activate: false` option in `geoschem_config.yml`, which was introduced as a run-time option for debugging purposes. Disabling photolysis caused skipping the calls to `Rd_AOD` and `Calc_AOD` within `Init_Photolysis`. However, these subroutines must be called regardless of whether J-values will be computed since they are used for other parts of GEOS-Chem.

### Expected changes

There will no longer be a run-time error if photolysis is disabled in `geoschem_config.yml`. Otherwise this is a zero diff update.

### Related Github Issue(s)

This corrects an issue introduced in https://github.com/geoschem/geos-chem/pull/1733.

### DIFF
--- a/GeosCore/photolysis_mod.F90
+++ b/GeosCore/photolysis_mod.F90
@@ -1022,15 +1022,15 @@ CONTAINS
 !
 ! !INPUT PARAMETERS:
 !
-    TYPE(OptInput), INTENT(IN) :: Input_Opt
+    TYPE(OptInput), INTENT(IN)    :: Input_Opt
 !
 ! !INPUT/OUTPUT PARAMETERS:
 !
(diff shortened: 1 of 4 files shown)


yantosca (approved):
This is a structural update that makes sure that RD_AER and CALC_AOD are called even if photolysis is off.  Also the RRATE arrays in UCX_Mod are not computed if photolysis is off.  Good to merge.

lizziel:
That's `RD_AOD` and not `RD_AER`. Just want to clarify since `RDAER` is a subroutine used elsewhere and don't want to have any confusion.

yantosca:
@lizziel: Duly noted.

Issue #1762, opened 2023-04-18
GCHP rundir script setCommonRunSettings.sh fails on "which python" if python is not installed on the node


yantosca:
### Name and Institution (Required)

Name:  Bob Yantosca
Institution: Harvard + GCST

### Confirm you have reviewed the following documentation

- [x] [Support guidelines](https://geos-chem.readthedocs.io/en/stable/reference/SUPPORT.html)
- [x] [User manuals](https://geos-chem.readthedocs.io/en/stable/geos-chem-shared-docs/supplemental-guides/related-docs.html)
- [x] [Debugging GEOS-Chem and HEMCO errors](https://geos-chem.readthedocs.io/en/stable/geos-chem-shared-docs/supplemental-guides/debug-guide.html)
- [x] [Current and past Github issues](https://github.com/geoschem/geos-chem/issues)

### Description of your issue or question

I have been testing GCHP on new nodes that are set up with Rocky Linux.  These nodes do not have python installed in the usual locations:

```console
$ which python
/usr/bin/which: no python in (~/bin:~/bin/netcdf-scripts/scripts:/n/home09/ryantosca/miniconda3/condabin:~/bin:~/bin/netcdf-scripts/scripts:/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/opt/puppetlabs/bin)
```

This causes the `setCommonRunSettings.sh` script to halt with an error.  The offending code is here:

https://github.com/geoschem/geos-chem/blob/4722f288e90291ba904222f4bbe4fc216d17c34a/run/GCHP/setCommonRunSettings.sh.template#L223-L237

This is because the `which python` returns failure before it even gets to the `$bc_ok` block.

I propose editing the if block as follows:
```bash
   # Use "bash calculator" if available; Python if not; fail otherwise
   which bc &> /dev/null; bc_ok=$?
   if [[ $bc_ok -eq 0 ]]; then
      # Use bash calculator
      SQRT=$(echo "sqrt (${Z})" | bc -l)
      N=$(echo $SQRT | awk '{print int($1+0.999)}')
   else
      which python &> /dev/null; py_ok=$?
      if [[ $py_ok -eq 0 ]]; then
	  # Use system Python
	  SQRT=$( python -c "import math; print(int(math.sqrt(${Z})))" )
	  N=$SQRT
      else
	  echo "Cannot auto-determine NX and NY (need either bc or python available)"
	  exit 70
      fi
   fi
```
[...]

PR #1777, opened 2023-04-25
Update GCHP check for python and bc libs to prevent error checking fail

lizziel:

Name: Lizzie Lundgren

### Describe the update

The PR prevents `setCommonRunSettings.sh` from failing if either the `bc` or `python` libraries are not found. This is necessary because of a recent update that added `set -e` to the script, causing the script to stop if any command has return status other than 0.

### Expected changes

No failure running `setCommonRunSettings.sh` if `which bc` or `which python` fail.

### Related Github Issue(s)

https://github.com/geoschem/geos-chem/issues/1762


yantosca (approved):
This is a simple bug fix that sets the value of bc_ok and py_ok to zero before it is used.  This will prevent GCHP from dying if python is not found on the system.  Good to merge.

Issue #1814, opened 2023-05-31
Slowdown in GEOS-Chem 14.2.0


msulprizio:
### Name and Institution (Required)

Name: Melissa Sulprizio
Institution: Harvard / GCST

### Confirm you have reviewed the following documentation

- [x] [Support guidelines](https://geos-chem.readthedocs.io/en/stable/reference/SUPPORT.html)
- [x] [User manuals](https://geos-chem.readthedocs.io/en/stable/geos-chem-shared-docs/supplemental-guides/related-docs.html)
- [x] [Debugging GEOS-Chem and HEMCO errors](https://geos-chem.readthedocs.io/en/stable/geos-chem-shared-docs/supplemental-guides/debug-guide.html)
- [x] [Current and past Github issues](https://github.com/geoschem/geos-chem/issues)

### Description of your issue or question

The 1-month benchmarks for 14.2.0 alpha tags indicate a large increase in run time:

<img width="1353" alt="image001" src="https://github.com/geoschem/geos-chem/assets/4805437/4f9534e5-8c5b-46da-bd5e-ede40f0c9045">

<img width="388" alt="Screenshot 2023-05-31 at 10 23 59 AM" src="https://github.com/geoschem/geos-chem/assets/4805437/948820a5-821e-4ab2-be95-52fba1e60f8c">

The largest increase is seen in the gas-phase chemistry and came in with the PR #1620 to restore Iy sink reactions. We see another smaller increase in wall time after merging in PR #1702 to restore consideration of sea salt alkalinity to heterogenerous acid-catalyzed reactions of halogens on sea salt aerosols.

PR #1620 also included modifying the RTOL value in fullchem_mod.F90 from 0.5e-2 to 0.5e-3 to avoid convergence errors. However, since other chemistry updates have gone into 14.2.0, we may be able to restore RTOL to its previous value.

Tagging @yantosca @beckyalexander @viral211 for further comments.

yantosca:
The wall time increases 01:20 in going from alpha.4 to alpha.5.  I took a look at the code but I didn't see anything that stands out that would have caused the slowdown.  In fact, I replaced several `(1.0_dp - H%f_Alk_SSA)` etc. expressions with `H%f_Acid_SSA`, where that is only computed once per grid box.  The increase in wall time might have also been due to the tolerance being so tight.

msulprizio:
As a follow up, the model performance returns to 14.1.1 wall times after reverting RTOL (#1815).

[...]

PR #1620, opened 2023-01-24
Restore Iy sink reactions and rebuild fullchem mechanism with KPP 3.0.0

yantosca:
This is the companion PR for #1585.  We have restored sink reactions for HOI, IONO, and IONO2, and removed IONO2 recycling on alkaline seasalt, according to recommendations by @beckyalexander.  For more information, please see the writeup in #1585.

Closes #1585

PR #1779, opened 2023-04-26
Add modifications to sea salt chemistry in order to avoid KPP integration errors

yantosca:
### Name and Institution (Required)

Name: Bob Yantosca
Institution: Harvard + GCST

### Confirm you have reviewed the following documentation

- [x] [Contributing guidelines](https://geos-chem.readthedocs.io/en/stable/reference/CONTRIBUTING.html)

### Describe the update

This is the companion PR to #1768.  This adds the following updates to the fullchem and custom mechanisms to avoid non-convergence errors in the KPP integration:

1. Cap `State_Het%f_Alk_SSA` and `State_Het%f_Alk_SSC` fractions at 1.  This will prevent small concentrations of SALA and SALC  species (which are in the denominators used to compute these expressions) from blowing up these fractions to unphysical values.  See commit 82ecd04a49351d4dc206861879ed7a2faf23a03e.

2. Balance IONO and IONO2 reactions forming seasalt aerosol, following the suggestion of @beckyalexander.  See commit 5c5b82a0efd3bccdd95e8d47085b03fd59d7e8fd.

3. Modify several rate-law reactions to avoid performing uptake on sea salt aerosol in the stratosphere, based on the suggestion of @beckyalexander and @viral211.  See commit: 1a7c81b6e013b58c87b2bb97228f71459fa5149f

Integration tests are in progress.

### Expected changes

These updates should produce small changes in most species, but larger changes in iodine and/or seasalt species.   Also, simulations should no longer terminate due to KPP integration errors.

### Reference(s)

N/A

### Related Github Issue(s)

Closes #1768

PR #1815, opened 2023-05-31
Change relative tolerance back to 0.5e-2 to address model slowdown

msulprizio:

Name: Melissa Sulprizio
Institution: Harvard / GCST

### Describe the update

In GEOS-Chem 14.2.0, 1-month benchmarks have shown a large increase in wall time beginning around 14.2.0-alpha.4 when PR #1620 for restoring the Iy sink reactions was merged. That pull request also included changing the relative tolerance value (RTOL in fullchem_mod.F90) from 0.5e-2 to 0.5e-3 to avoid integration errors in halogen chemistry. However, many other chemistry fixes and updates have gone into 14.2.0 since then so we may be able to restore the RTOL value in an attempt to reduce wall time.

### Expected changes

This fix will improve wall time and should have a small impact on model output.

### Related Github Issue(s)

#1814


yantosca (approved):
This is a simple change that undoes a modification that was brought in by @beckyalexander.  The tighter tolerance may no longer be needed as we now have added extra protections against convergence errors in #1779.  Good to merge.

PR #1848, opened 2023-06-26
Updates to run GEOS-Chem with GEOS-IT

lizziel:

Name: Lizzie Lundgren

### Describe the update

This PR includes updates needed to create functional run directories for GCHP and GC-Classic using GEOS-IT data, including on the NASA discover cluster. Users are asked whether they are using discover or not when creating a GCHP run directory creation using GEOS-IT because the filenames and directory structure for GEOS-IT are different on discover than elsewhere.

This PR also restructures the options for advection meteorology when using GEOS-FP native files, and how the GEOS-FP and GEOS-IT meteorology settings are stored in the run/shared/settings folder. These changes are zero diff and meant only to make run directory set up and config file storage easier to understand.

This is a zero diff update for benchmarking.

### Expected changes

New option during run directory creation to use GEOS-IT meteorology in GCHP and GC-Classic.

### Reference(s)

None

### Related Github Issue(s)

This PR should be merged at the same time as the following HEMCO PR which updates constants file year for GEOS-IT:
https://github.com/geoschem/HEMCO/pull/252


lizziel:
This is functional and ready for review. All integration tests pass. However, we should not merge until we get the go-ahead from Randall to make GEOS-IT an option in the standard model. It is already being used on discover, however, so if needed I could make it "for use on discover only" during run directory creation.

lizziel:
I rebased this on dev/14.3.0 and all integrations tests pass for GC-Classic and GCHP.

msulprizio (approved):

Issue #1768, opened 2023-04-21
GEOS-Chem 14.2.0 development branch crashing with integration errors


msulprizio:
### Name and Institution (Required)

Name: Melissa Sulprizio
Institution: Harvard/GCST

### Confirm you have reviewed the following documentation

- [x] [Support guidelines](https://geos-chem.readthedocs.io/en/stable/reference/SUPPORT.html)
- [x] [User manuals](https://geos-chem.readthedocs.io/en/stable/geos-chem-shared-docs/supplemental-guides/related-docs.html)
- [x] [Debugging GEOS-Chem and HEMCO errors](https://geos-chem.readthedocs.io/en/stable/geos-chem-shared-docs/supplemental-guides/debug-guide.html)
- [x] [Current and past Github issues](https://github.com/geoschem/geos-chem/issues)

### Description of your issue or question

When running a 4x5 GEOS-FP simulation for 2016 with GEOS-Chem 14.2.0-alpha.8, Nadia Colombi's runs crash with an integration error three months into the simulation with:

```
---> DATE: 2016/03/05  UTC: 15:20
 Forced exit from Rosenbrock due to the following error:
 --> Step size too small: T + 10*H = T or H < Roundoff
 T=   4.8000000000000068E-018 and H=   4.8000000000000068E-018
 ### INTEGRATE RETURNED ERROR AT:           54          44          53
 Forced exit from Rosenbrock due to the following error:
 --> Step size too small: T + 10*H = T or H < Roundoff
 T=   4.0000000000000041E-018 and H=   4.0000000000000041E-018
## INTEGRATE FAILED TWICE !!! 
 #############################################################################################################################################################################################################################################
 ### KPP DEBUG OUTPUT!
 ### Species concentrations at problem box           54          44          53
 #############################################################################################################################################################################################################################################
 ###    1.5330768753631170E-029 CH2I2
 ###    9.7494583413162528E-025 CH2IBr
 ###    2.3087058718844786E-023 CH2ICl
 ###    37627.245756688040      AERI
 ###    12612.597852921273      AONITA
 ...
```

The full log file may be viewed [here](https://github.com/geoschem/geos-chem/files/11295403/GC.log). 

[...]

PR #1702, opened 2023-03-16
Restore consideration of sea salt alkalinity to heterogenerous acid-catalyzed reactions of halogens on sea salt aerosols

msulprizio:
### Name and Institution (Required)

Name: Melissa Sulprizio
Institution: Harvard/ GCST

### Confirm you have reviewed the following documentation

- [x] [Contributing guidelines](https://geos-chem.readthedocs.io/en/stable/reference/CONTRIBUTING.html)

### Describe the update

This is the corresponding PR to https://github.com/geoschem/geos-chem/issues/1547. This PR fixes a bug in heterogeneous chemistry by restoring sea salt alkalinity to certain reactions of halogen chemistry.

This PR supersedes https://github.com/geoschem/geos-chem/pull/1548, which was intended for inclusion into 14.1.0 but was delayed because of convergence errors in KPP.

### Expected changes

Please see https://github.com/geoschem/geos-chem/issues/1547 for a complete description of this issue by @beckyalexander, with plots.

### Related Github Issue(s)

Closes https://github.com/geoschem/geos-chem/issues/1547.

PR #1878, opened 2023-07-17
Adjust relative tolerance after first KPP integrate failure

msulprizio:

Name: Melissa Sulprizio
Institution: Harvard / GCST

### Describe the update

Following the chemistry updates in 14.2.0, users are encountering KPP convergence errors more frequently. To fix this, RTOL was changed from 0.5e-2 to 0.5e-3, but this drastically slowed down simulations so it was changed back to 0.5e-2. Because we are still seeing these convergence errors, one workaround is to adjust the RTOL value to 0.5e-3 after the KPP integration fails the first time. If it fails twice, GEOS-Chem will still crash as usual. The RTOL value resets to 0.5e-2 for the next grid box in the loop.

### Expected changes

The differences for this fix should be on the level of numerical noise when a KPP integration error occurs.

### Related Github Issue(s)

- https://github.com/geoschem/geos-chem/issues/1768

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Change RTOL value from 0.5e-3 back to 0.5e-2 to address model slowdown
 - Allow the use of OFFLINE_SEASALT for seasalt alkalinity, Cl, and Br in GEOS-Chem within CESM
 - Renamed TransportTracer species for consistency with GMAO's TR_GridComp
+- Adjusted relative tolerance value if first KPP integration attempt fails
 
 ### Removed
 - `Warnings: 1` is now removed from `HEMCO_Config.rc.*` template files
--- a/GeosCore/fullchem_mod.F90
+++ b/GeosCore/fullchem_mod.F90
@@ -436,14 +436,6 @@ CONTAINS
     TIN       = T
     TOUT      = T + DT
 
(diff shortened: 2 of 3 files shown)


yantosca (changes requested):
We might need to move the original definition of RTOL into the parallel loop over (I,J,L).  RTOL is set to 0.5e-2 outside of the loop but upon error it is set to 0.5e-3 within the loop.  We would also need to make RTOL (and maybe ATOL as well, for consistency) `!$OMP PRIVATE` variables.

https://github.com/geoschem/geos-chem/blob/709729436dc0e5f1e2b282c01406b192b4b9a598/GeosCore/fullchem_mod.F90#L430-L445

msulprizio:
> We might need to move the original definition of RTOL into the parallel loop over (I,J,L). RTOL is set to 0.5e-2 outside of the loop but upon error it is set to 0.5e-3 within the loop. We would also need to make RTOL (and maybe ATOL as well, for consistency) `!$OMP PRIVATE` variables.
> 
> https://github.com/geoschem/geos-chem/blob/709729436dc0e5f1e2b282c01406b192b4b9a598/GeosCore/fullchem_mod.F90#L430-L445

@yantosca Do we need to set ATOL and RTOL as `$OMP+PRIVATE` in fullchem_mod.F90 or as `!$OMP THREADPRIVATE` in KPP/fullchem/gckpp_Global.F90?

msulprizio:
This update is no longer needed after reverting the changes introduced in PR #1702 via PR #1905. I will close it without merging.

PR #1887, opened 2023-07-24
Disable SpeciesConcMND output for benchmark simulations

msulprizio:

Name: Melissa Sulprizio
Institution: Harvard / GCST

### Describe the update

Both SpeciesConcVV and SpeciesConcMND diagnostics were being saved out to the GEOSChem.SpeciesConc*.nc4 files in benchmark simulations. The SpeciesConcMND diagnostics are not actually needed for benchmarking, so we can disable this output to reduce file size.

### Expected changes

This change will reduce the size of the GEOSChem.SpeciesConc*.nc4 files, but will not change model output otherwise.


yantosca (approved):
Looks good to merge.  Thanks @msulprizio!

PR #1916, opened 2023-08-28
Add fixes to reduce differences between the CH4 simulation and CH4 in carbon simulation

msulprizio:

Name: Melissa Sulprizio
Institution: Harvard / GCST

### Describe the update

This pull request eliminates many differences between the CH4 simulation and CH4 in the carbon simulation. Most of the updates are in HEMCO_Config.rc.carbon and HEMCO_Config.rc.CH4 to ensure consistency between the CH4 entries. This included:

- Fixing several typos in HEMCO_Config.rc.carbon related to masks and hierarchies
- Using v5 global mean OH in the carbon simulation instead of output from the GEOS-Chem 14.0.0 10-year benchmark.
- Removing the UPDATED_GFED4 option from the CH4 simulation and using the GFED HEMCO extension instead
- Ensuring both simulations use files from the same directories (i.e. latest version) for each inventory

In addition, we now save out the StateMet collection for both simulations by default to allow for benchmarking via gcpy.

### Expected changes

This update will not impact full-chemistry benchmark simulations, but will cause changes in the CH4 and carbon simulations.


yantosca (approved):
These changes correct several issues in the CH4, tagCH4, and carbon configuration files, such as:
- GCAP2SCENARIO -> GCAPSCENARIO
- Updating emissions that are used
- Updating scale factors that are used

These are OK to merge.

PR #1950, opened 2023-09-10
Add tolerance to analytical inversion comparison for requested state vector element

sabourbaray:

Name: Sabour Baray
Institution: Environment and Climate Change Canada

[✓] [Contributing guidelines](https://geos-chem.readthedocs.io/en/stable/reference/CONTRIBUTING.html)

### Describe the update

This bugfix adds a tolerance to GeosCore/global_ch4_mod.F90 when searching for the requested state vector element in geoschem_config.yml. This is based on the issue described in detail in the IMI repository:

https://github.com/geoschem/integrated_methane_inversion/issues/167

This update can be used as a temporary solution, since adding a tolerance may not be the preferred approach. The original issue may be in how HCO_GC_EvalFld interacts with the state vector built by the IMI, where there may be conversions between data types that introduces this small numerical error.

### DIFF
--- a/GeosCore/global_ch4_mod.F90
+++ b/GeosCore/global_ch4_mod.F90
@@ -126,6 +126,9 @@ CONTAINS
     ! Arrays of state vector elements for applying emissions perturbations
     REAL(fp)           :: STATE_VECTOR(State_Grid%NX,State_Grid%NY)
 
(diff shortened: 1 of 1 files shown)


msulprizio:
The issue mentioned above has been addressed by #1976, so this PR should no longer be needed.

PR #1989, opened 2023-10-13
Update sample restart files for fullchem and TransportTracers to 14.2.0 benchmark files

### LINKED ISSUE #1986: Some 14.2.1 simulation runs fail due to missing species in restart file

Name: Lizzie Lundgren

### Description of your issue or question

GEOS-Chem 14.2.0 and 14.2.1 transport tracer and standard full chemistry simulations fail because several advected species are not found in the restart file that comes with the run directory. The restart files included in run directories are the same as in 14.1.0 and need to be updated. 14.2.0/14.2.1 includes new tracers in these simulations. The transport tracer simulation also features several species with new names.

We will release a version 14.2.2 that uses the new restart files generated during the 14.2.0 benchmark.

**GC-Classic work-around:** Allow GC-Classic to use background values for missing species by updating `HEMCO_Config.rc` time cycling entry for restart variable species (`SPC_` in 2nd column) from `EFY0` to `EY`.

**GCHP work-around**: You can allow GCHP to use background values for missing species by setting `Require_Species_in_Restart` in run directory config file `setCommonRunSettings.sh` to `0` (false).

msulprizio:

Name: Melissa Sulprizio
Institution: Harvard / GCST

### Describe the update

Out-of-the-box fullchem and TransportTracers simulations for 14.2.0 and 14.2.1 are failing because additional species were added in 14.2.0. The time cycle flag used in HEMCO_Config.rc (EYFO) forces GEOS-Chem to crash with an error when species are not found in the restart file. To avoid this, we need to update the sample restart files for these simulations to restart files saved out from the 14.2.0 1-year benchmark simulations.

### Expected changes

This will cause changes to the benchmark simulations due to changing initial conditions.

### Related Github Issue(s)

Closes #1986

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ This file documents all notable changes to the GEOS-Chem repository starting in 
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased 14.2.2] - TBD
+### Changed
+- Updated sample restart files for fullchem and TransportTracers simulations to files saved out from the 14.2.0 1-year benchmarks
+
 ## [14.2.1] - 2023-10-10
 ### Added
 - Script `test/difference/diffTest.sh`, checks 2 different integration tests for differences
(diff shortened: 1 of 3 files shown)


yantosca (approved):
Thanks @msulprizio.  This is good to merge, as it just changes the directory from which restarts are pulled.

msulprizio:
All integration tests passed for GCClassic and GCHP (except TOMAS in GCClassic, which is a known issue).

PR #2002, opened 2023-10-19
12.9.3+bugfix budget diagnostic

cdholmes:

Name: Christopher Holmes
Institution: Florida State University

### Describe the update

With this update, the budget diagnostics include the troposphere and PBL mass tendencies associated with changes in the PBL height and tropopause height. The effect of changing tropopause height is now included in the "BudgetTransportTrop" diagnostic. The effect of changing PBL height is now included in the "BudgetMixingPBL" diagnostic.

A second update improves the way that the model level number corresponding to the PBL depth is determined. The meteorological input files provide the PBL height. The previous method computed the pressure at the PBL top using a globally constant scale height. The revised method determines the level from the local grid cell heights, which is more appropriate. As a result of this change the mixing increases up to 1 level near the poles and decreases up to 1 or 2 levels in the tropics.

### Expected changes

Figures show changes in the PBL due to the improved calculation.

Change in PBL top level (MERRA-2 2015-07-01 18z)
image

Change in PBL thickness, hPa (MERRA-2 2015-07-01 18z)
image

### Reference(s)

none 

### Related Github Issue(s)

The budget diagnostic bug was discussed by GCSC. I corresponded with Lizzie about the issue.

### DIFF
--- a/GeosCore/main.F90
+++ b/GeosCore/main.F90
@@ -1170,6 +1170,25 @@ PROGRAM GEOS_Chem
        ! Do not compute any data in dry-run mode
        IF ( notDryRun ) THEN
 
+          !------------------------------------------------------
+          ! Tropopause height budget diagnostics - Part 1 of 2
+          !------------------------------------------------------
+          IF ( State_Diag%Archive_BudgetTransport ) THEN
+             ! Get initial column masses
+             CALL Compute_Column_Mass( Input_Opt,                           &
+                                     State_Chm, State_Grid, State_Met,    &
+                                     State_Chm%Map_Advect,                &
+                                     .False., &
+                                     State_Diag%Archive_BudgetTransportTrop, &
+                                     .False.,  &
+                                     State_Diag%BudgetMass1,              &
+                                     RC )
+             IF ( RC /= GC_SUCCESS ) THEN
+                ErrMsg = 'Tropopause height budget diagnostics error 1'
+                CALL GC_Error( ErrMsg, RC, ThisLoc )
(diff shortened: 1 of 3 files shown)


lizziel:
Hi @cdholmes, I am reviewing your updates and getting them into the recent version. However, it looks like there may be part of the update missing. Do you also have updates for subroutine `Compute_Budget_Diagnostics`? As is, it is not clear about the two new budget arrays are computed.

lizziel:
Would you be able break this up into two different submitted updates? They would be (1) change how PBL top level is computed, and (2) add improvements to the transport trop and mixing PBL budgets. The first will change GEOS-Chem benchmark results and the latter will not. I am getting a lot of merge conflicts because 12.9.3 is so old. I can work through those for the diagnostics update but prefer that you make the PBL top level definition change off of the most recent version to avoid introducing bugs.

cdholmes:
@lizziel, The two fixes are in separate commits within this PR, but I can split the two PRs if that is helpful. 

My fix for the budget diagnostic introduces two new arrays (`BudgetTransportTropHeight` and `BudgetMixingPBLHeight`) that quantify mass changes due to the tropopause and PBL height. These arrays are added into `BudgetTransportTrop` and `BudgetMixingPBL` within `main.F90` at the end of each dynamics time step. There's no change to `Compute_Budget_Diagnostics`. I have not checked if the same strategy will work in 14.2

cdholmes:
@lizziel. I'll close this PR as the content is now split into two PRs: #2078 and #2079.

Issue #1094, opened 2022-01-14
[FEATURE REQUEST] Stratospheric adjustment with RRTMG in GEOS-Chem


sdeastham:
### Fixed dynamical heating with RRTMG, and other improvements
Our lab (MIT LAE) recently(ish) completed some work which uses RRTMG in GEOS-Chem v11-02 to evaluate the impacts of supersonic aviation on the environment. To facilitate this, we implemented the fixed dynamical heating assumption in RRTMG, allowing stratospherically-adjusted radiative forcing to be calculated at the tropopause (consistent with IPCC definitions) rather than the TOA and surface instantaneous radiative forcing which is standard in GEOS-Chem at the moment. This request is to signal that I'd like to merge the updated RRTMG into the GEOS-Chem main branch in the near future, pending appropriate approval and merge processes. New features:

* Evaluation of RF at the tropopause, as well as at the surface and TOA
* Fixed dynamical heating assumption (Feys at al, 1980; Maycock et al, 2011) allows evaluation of stratospheric adjustment
* Stratospheric temperature adjustment now available as an output
* Direct radiative effect of CO2, H2O, N2O, and CFCs can now be calculated and output if desired
* Removed arbitrary constraints which were left over from when UCX was optional (e.g. only including tropospheric BC in calculations)
* Added an implementation-agnostic message logging interface which allows GC to take advantage of pfLogger in MAPL [_not related to RRTMG - but was developed to assist with debugging_]
* Experimental implementation of seasonally-evolving FDH per Forster et al. (1997), with 

The new implementation is designed for use with GCHP. The GC-Classic call to RRTMG supports only the original functionality (instantaneous RF). Extension for use in GCHP should be relatively straightforward if a user is interested in doing so.

_References_

Fels S B, Mahlman J D, Schwarzkopf M D and Sinclair R W 1980 Stratospheric Sensitivity to Perturbations in Ozone and Carbon Dioxide: Radiative and Dynamical Response J. Atmos. Sci. 37 2265–97
de F. Forster P M and Shine K P 1997 Radiative forcing and temperature trends from stratospheric ozone changes J. Geophys. Res. 102 10841–55
Maycock a. C, Shine K P and Joshi M M 2011 The temperature response to stratospheric water vapour changes Quart. J. Roy. Meteor. Soc. 137 1070–82

yantosca:
This has been added to the GEOS-Chem Model Development Priorities table (not prioritized, for now): 

[...]

Issue #1881, opened 2023-07-18
[FEATURE REQUEST] Enable calculation of RF from stratospheric changes


sdeastham:
https://github.com/geoschem/geos-chem/blob/4722f288e90291ba904222f4bbe4fc216d17c34a/GeosCore/rrtmg_rad_transfer_mod.F90#L680C12-L680C49

It would be a trivial change to have RF calculations from RRTMG include changes in stratospheric water vapor (by using the H2O tracer rather than AVGW). As long as there are no objections I will submit a PR which makes this change.

Similarly, it's not clear to me why stratospheric changes in ozone and CH4 (https://github.com/geoschem/geos-chem/blob/4722f288e90291ba904222f4bbe4fc216d17c34a/GeosCore/rrtmg_rad_transfer_mod.F90#L801-L805) are not included in the calculation of ozone- and methane-related RF. I will include these in a subsequent PR.

yantosca:
We can close out this issue as PR #2010 was merged into 14.3.0.

PR #2010, opened 2023-10-23
Stratospheric adjustment and additional output fields for RRTMG in GCHP

sdeastham:

Name: Sebastian Eastham
Institution: Massachusetts Institute of Technology

### Describe the update

This update implements multiple updates to RRTMG for calculations of radiative forcing (RF):
* Changes in the stratosphere are now included in RRTMG calculations (resolving https://github.com/geoschem/geos-chem/issues/1881)
* In addition to the existing outputs, RRTMG can now output the contribution of the following to RF:
  * Tropospheric ozone (previously "ozone" reported changes in tropospheric ozone only)
  * All ozone (see note above)
  * CFCs
  * Water vapor
  * N2O
  * CO2
* RF can now be output at the tropopause, as well as at the surface and TOA. Tropopause RF is calculated as the sum of downwelling radiation (LW + SW) minus upwelling
* Net heating rates can now be output during a radiation call using the field "DynHeating" (strictly this is -1 times the heating rates, as the field is intended to be used to support stratospheric adjustment - see below)
* The CO2 concentration perceived by RRTMG can be modified in geoschem_config.yml
* Water vapor is now set to match the tracer field at all altitudes
* The RRTMG output fields are now set to `MAPL_UNDEF` on steps when RRTMG is not called. This ensures that the time-averaged radiative effect does not incorrectly average a series of zero values prior to the first call to RRTMG (as MAPL knows to exclude `MAPL_UNDEF` from averaging)
* The stratospherically-adjusted RF at the tropopause (the quantity referred to historically as "RF" by the IPCC) can now be calculated using an RK4 time marching integration with the fixed dynamical heating approximation (FDH) (this resolves https://github.com/geoschem/geos-chem/issues/1094). FDH is described by Forster et al. (1997), and this implementation is described by Eastham et al. (2022). If the following steps are taken, the radiative fluxes in each simulation will include stratospheric adjustments of temperature relative to the baseline simulation:
  * Running a reference simulation with no perturbations. The reference simulation corresponds to the baseline atmosphere, and is the reference for all subsequent calculations of radiative forcing (e.g. to calculate the RF attributable to aviation, one needs a baseline simulation in which everything is normal - that is this simulation). In the `rrtmg_rad_transfer_model` part of geoschem_config.yml the line `fixed_dyn_heating` should be set to `true`, and `read_dyn_heating` should be `false`. The user will need to run RRTMG with some time step (e.g. 3 hours) and include the `DynHeating` collection, time averaged, with the same time step. **This simulation needs to run for at least one time step longer than you need** so that there is an "end" sample for subsequent simulations.**
  * For all perturbations (e.g. removing aviation), set up the run directory in the same way (but e.g. setting aviation emissions to zero to see the aviation-attributable delta). However, the following changes are needed:
    * The line `IMPORT_DYN_HEATING: 0` in GCHP.rc should be changed from 0 to 1
    * A symbolic link should be made in the current directory called `Heating_Baseline` which points to the folder containing the `DynHeating` output from your reference simulation (i.e. `ln -s /path/to/reference/rundir/OutputDir Heating_Baseline`)
    * The line `read_dyn_heating` should also be set to `true` in `geoschem_config.yml`
    * If desired, you can include `DTRad` in your RRTMG output collection (included by default). This will show you the temperature adjustment reached by the stratospheric adjustment routine.
  * NB: It is normal to see occasional warnings like "RRTMG FDH routine failed to converge for 5 of 144 columns on CPU 0". This indicates that one column could not be brought to equilibrium, but usually reflects a small oscillatory error only. However, if you are consistently seeing large numbers (e.g. >25% of columns) then you are likely seeing a legitimate error.
* **EXPERIMENTAL**: There is also an optional to apply the seasonally-evolving fixed dynamical heating approximation (Kiehl et al., 1999). To use this approach, follow the same steps as are outlined above but when running the perturbation simulations _also_ turn on `seasonal_fdh` in geoschem_config.yml. **This option has not been extensively tested.**

The FDH approach will substantially increase RRTMG run time, as it requires multiple additional calls to the longwave radiative transfer routine. This could be reduced by modifying the convergence approach and/or criteria. The SEFDH approach does not require convergence and will therefore be faster but does not yield the same results (by design).

### Expected changes

There are new optional RRTMG outputs at the tropopause, using the same format as the TOA outputs but now using "Trop" as the identifier. Tropospheric ozone, CFC, N2O, water vapor, and CO2 radiative effects can be calculated by requesting the same data as usual but including the keywords `O3T`, `CFC`, `N2O`, `H2O`, and `CO2` respectively in HISTORY.rc (e.g. requesting `RadAllSkyTrop_H2O` will return the radiative effect of water vapor with regards to fluxes at the tropopause, including the effects of clouds). As before, the total effect of ozone (O3), methane (ME), and aerosols can still be broken out, and the total (BASE) effect is still available. All these fields are now included in the standard RRTMG history collection along with any calculated change in temperature due to stratospheric adjustment (`DTRad`).

For a reference simulation, or if the FDH assumption is not turned on, some small changes are still expected as the radiative effect of black carbon and other aerosols in the stratosphere (alongside typical "stratospheric aerosol") is now considered. These changes are small. The calculated radiative effect of several tracers including ozone is shown to be larger as stratospheric concentrations are now included in the calculation (if users wish to break out the specific contribution of tropospheric and stratospheric ozone changes, they can do so by requesting the O3T "species"). Enabling the FDH assumption results in stratospheric temperature adjustment until the net heating rate in each stratospheric grid cell falls to less than 0.001 K/day (the residual maximum defined as `HRmax` in `rrtmg_radiative_transfer_mod.F90`) or the time marching routine exceeds 150 days (`dtadj_max`).

Tests which prescribe a 100 ppmv increase in global CO<sub>2</sub> show that the use of FDH results in 
* A reduction in the net CO<sub>2</sub> radiative forcing at the tropopause (from 1.4 to 1.3 W/m<sup>2</sup>)
* The TOA RF increasing from 0.7 to 1.2 W/m<sup>2</sup>

In comparison, the established relationships for CO<sub>2</sub> from Etminan et al. (2016) estimate that (for a background NO<sub>2</sub> of 330 ppmv) the same change in CO<sub>2</sub> should result in 1.22 W/m<sup>2</sup> of forcing (stratospherically adjusted, at the tropopause). This suggests that the RRTMG calculations are causing improved agreement. The remaining discrepancy between TOA and tropopause RF after stratospheric adjustment, although small, is not expected and may be due to the fact that we apply temperature adjustment only in locations which are marked as `InStratosphere`. Better results may be found if we allow adjustment up to the top of the model, or use top-of-the-stratosphere rather than TOA for comparison. Happy to supply additional results if helpful.

### Reference(s)

The approach used here is described in the appendix of Eastham et al. (2022) (https://pubs.rsc.org/en/content/articlehtml/2022/ea/d1ea00081k). The seasonally-evolving fixed dynamical heating approximation is described in Kiehl et al. (1999) (https://agupubs.onlinelibrary.wiley.com/doi/abs/10.1029/1999JD900991). The original fixed dynamical heating approach is described by Forster et al. (1997) (https://link.springer.com/article/10.1007/s003820050182). The values used for comparison are calculated based on Etminan et al. (2016) (https://agupubs.onlinelibrary.wiley.com/doi/full/10.1002/2016GL071930).

### Related Github Issue(s)

https://github.com/geoschem/geos-chem/issues/1881
https://github.com/geoschem/geos-chem/issues/1094


lizziel (changes requested):
See https://github.com/sdeastham/geos-chem/pull/1 for requested changes. You can merge that directly into your branch. We should still be able to get this into 14.3.0 since we have extended the freeze date.

lizziel (approved):
Good to merge!

lizziel:
Actually there are still a couple bugs remaining to successfully build GCHP 14.3 with this update. @sdeastham, see https://github.com/sdeastham/geos-chem/pull/2. If you merge that in the next couple days we may be able to still get this into 14.3.0.

lizziel:
One more PR: https://github.com/sdeastham/geos-chem/pull/3. Hopefully this is the last! Integrations tests all pass now but I noticed diffs in non-RRTMG simulations due to the new internal state variable.

lizziel:
@msulprizio, this PR is now ready to merge into 14.3. GC-Classic and GCHP integration tests all pass and difference tests show diffs only in GCHP RRTMG diagnostics and restart file, as expected.

cbarker211:
@sdeastham Hi Seb, this is a great addition to RRTMG in GEOS-Chem, thanks for the update! I know this pull request is specifically for GCHP, but I wanted some clarification on using these updates in GCClassic. 

I get an error that "Fixed dynamical heating in RRTMG is currently only available in GCHP". Am I still able to use the tropopause outputs ('RadAllSkyLWTrop_?RRTMG? ',  'RadAllSkySWTrop_?RRTMG? ', 'RadClrSkyLWTrop_?RRTMG? ', 'RadClrSkySWTrop_?RRTMG? ') if FDN is switched off when using GCClassic? And does this mean that stratospheric adjustment is only available in GCHP?

sdeastham:
@cbarker211 happy to help! In theory the tropopause outputs should still be available - all they're doing is outputting the fluxes at the tropopause level, which doesn't require the stratospheric adjustment code. However, strat adjustment is indeed currently only available for GCHP. It shouldn't be too hard to port it to GC-Classic (basically you just need to add a few additional calls in the GC-Classic main routines, mirroring those in the GCHP main routines, and then set up the plumbing to output DTRad during the calibration run).

cbarker211:
Hi Seb, I'm not very familiar with GCHP, by the GCHP main routines, do you mean the files in the GEOS-Chem/Interfaces/GCHP folder?

sdeastham:
Yep! This bit in particular: https://github.com/geoschem/geos-chem/blob/b288af1563a9a49a154a93bdd0f88be1f58f3f2f/Interfaces/GCHP/gchp_chunk_mod.F90#L1402-L1603. For import of DynHeating, I think you'll need to add some calls to get information from HEMCO - possibly looking at how (e.g.) met data, or the other radiative data, are read in could help?

Issue #1418, opened 2022-08-30
[Feature Request] Move aerosol module variables to State_Chm


lizziel:
We are trying to move away from module variables and instead store data that is persistent across timesteps within state variables. There are several module variables in aerosol_mod.F90 that could be moved to State_Chm.

yantosca:
This will be a necessary step in the work that I will be doing to clean up `aerosol_mod.F90`.

msulprizio:
The corresponding pull request (#2015) has now been merged into `dev/14.3.0`.

PR #2015, opened 2023-10-27
Move aerosol_mod.F90 arrays to State_Chm%AerMass object

lizziel:

Name: Lizzie Lundgren

### Describe the update

This PR moves all gridded module arrays in `aerosol_mod.90` to new derived type object `State_Chm%AerMass`. Updates include:

1. Created new file `aermass_container_mod.F90` to hold gridded arrays previously in `aerosol_mod.F90`, e.g. OCFPOA, BCPO, and other aerosol mass arrays computed in `Aerosol_Conc` subroutine
2. Created new derived type `AerMassContainer` defined in the new file
3. Created `State_Chm%AerMass` to point to the `AerMassContainer` object
4. Code updates to access the aerosol mass arrays via the State_Chm pointer, e.g. `State_Chm%AerMass%OCFPOA`
5. Removal of `aerosol_mod.F90` subroutine `Cleanup_Aerosol`
6. Moved diagnostic submodule `Set_AerMass_Diagnostic` in `aerosol_mod.F90` to be stored instead in `diagnostics_mod.F90`
7. Remove obsolete `oasave.F90` and code that references it
8. Removed the `Transport` library from `GeosCore/CMakeLists.txt`

Regarding the last update, removal of the `Transport` library, this CMake code was put in years ago with the intention of making it easier to include GEOS-Chem in external models without GEOS-Chem Classic transport. However, the files in the library are built before other `GeosCore` files and include `diagnostics_mod.F90` and `calc_met_mod.F90`. Excluding this built library in models would therefore break GEOS-Chem. The presence of this library also prevents using non-transport `GeosCore` modules within `diagnostics_mod.F90`, a problem that came up during this  aerosols update. In practice we do not need this separately built library when porting GEOS-Chem to other models. There are simpler ways of not building the transport files for all cases thus far.

This is a zero diff update but I would like it to go into 14.3.0 so that it can be used in the interface to Cloud-J.

All integration tests passed for GC-Classic and GCHP except the TOMAS simulations which is a known issue unrelated to this update.

### Expected changes

This is a no diff update for all simulations.

### Reference(s)

None

### Related Github Issue(s)

closes https://github.com/geoschem/geos-chem/issues/1418


GeosCore/diagnostics_mod.F90
```diff
@@ -1659,4 +1660,498 @@ SUBROUTINE Do_Archive_SatDiagn( Input_Opt,  State_Chm,  State_Diag,        &
...
+    PM25        => State_Chm%AerMass%PM25
+    PM10        => State_Chm%AerMass%PM10
+    ISOAAQ      => State_Chm%AerMass%ISOAAQ
+    SOAS        => State_Chm%AerMass%SOAS
+    FRAC_SNA    => State_Chm%AerMass%FRAC_SNA
+    DAERSL      => State_Chm%AerMass%DAERSL
+    WAERSL      => State_Chm%AerMass%WAERSL
+
+    ! Check that species units are kg/kg dry air
+    IF ( TRIM( State_Chm%Spc_Units ) /= 'kg/kg dry' ) THEN
```
yantosca, line 1833: This will cause a merge conflict when merging into dev/no-diff-to-benchmark.  Just to be aware.

Headers/aermass_container_mod.F90
```diff
@@ -0,0 +1,482 @@
...
+     ! SOAIE       : SOA product of IEPOX & HMML        [kg/m3]
+     ! PM25        : Particulate matter < 2.5 um        [kg/m3]
+     ! PM10        : Particulate matter < 10 um        [kg/m3]
+     ! ISOAAQ      : Isoprene SOA (aqueous formation)   [kg/m3]
+     ! SOAS        : Simple SOA                         [kg/m3]
+     ! FRAC_SNA    :
+     ! DAERSL      : Mass of hydrophobic aerosol (Mian Chin)
+     ! WAERSL      : Mass of hydrophilic aerosol (Mian Chin)
+     !========================================================================
+     REAL(fp), ALLOCATABLE :: OCFPOA     (:,:)
```
yantosca, line 68: Shouldn't these be POINTER instead of allocatable?  That would allow you to point to them without having to declare these with the TARGET attribute.

Headers/aermass_container_mod.F90
```diff
@@ -0,0 +1,482 @@
...
+
+    ! OM/OC for POA
+    ALLOCATE( Aer%OCFPOA( NX, NY ), STAT=RC )
+    IF ( RC /= GC_SUCCESS ) THEN
+       errMsg = 'Error allocating array OCFPOA!'
+       CALL GC_Error( errMsg, RC, thisLoc )
+       RETURN
+    ENDIF
+    Aer%OCFPOA = 0.0_fp
+
```
yantosca, line 182: You can use the GC_CheckVar routine for convenience.  It'll print the error message and then set RC to failure if the array can't be allocated.
```F90
ALLOCATE( Aer%OCFPOA( NX, NY ), STAT=RC )
GC_CheckVar( 'OCFPOA', 0, RC )
IF ( RC /= GC_SUCCESS ) RETURN
Aer%OCFPOA = 0.0_fp
```

yantosca (changes requested):
Mostly looks good.  I made a couple of review comments under Files Changed.  Let me know what you think @lizziel.

Headers/aermass_container_mod.F90
lizziel: I updated the code to use pointer instead. I didn't need to declare them as TARGET to point to them, and I'm not sure why. Do you know?

GeosCore/diagnostics_mod.F90
lizziel: I rebased my branch onto the dev/no-diff-to-benchmark and made the updates needed compatibility.

Headers/aermass_container_mod.F90
lizziel: Done. Thanks for the tip!

lizziel:
Thanks for the feedback @yantosca! I made the updates and now have it on top of the no-diff branch. Please review again at your convenience.

yantosca (approved):
Thanks @lizziel!  Good to merge.

Headers/aermass_container_mod.F90
yantosca: You can always point to a POINTER but an allocatable array needs to have a TARGET attribute.

PR #2018, opened 2023-10-31
Fix bugs setting strat aerosol OD passed to Fast-JX

lizziel:

Name: Lizzie Lundgren

### Describe the update

This update fixes a bug where stratospheric aerosols were treated as aerosols undergoing hygroscopic growth when assigning optical depth columns passed to Fast-JX to compute J-values.

The bug set stratospheric sulfate optical depth only where RH<50%. This had no impact because RH is below 50% for all grid cells in the stratospheric.

The bug also set polar stratospheric cloud optical depth incorrectly, to stratospheric sulfate optical depth where grid box RH is between 50% and 60%; zero elsewhere. This effectively made all PSC OD zero because RH is never in this range in the stratosphere. Actual PSC OD was set to entries in the aerosol column OD array that were never used in Fast-JX.

This bug went under the radar because the aerosol optical depth array was sized incorrectly. It incorrectly sized for all entries in the jv_spec_mie.dat LUT. That table contains many entries not relevant, such as different types of ice, RH bins not used, and aerosols not used. The aerosol OD column array should only contain entries for aerosols, specifically mineral dust, aerosols undergoing hygroscopic growth (5 entries for each aerosol to cover 5 different humidities), and two stratospheric aerosols. The total number is reprented by existing parameter AN_.

### Expected changes

This update will have a very small impact on J-values and thus will impact benchmark results.

### Reference(s)

None

### Related Github Issue(s)

Coming soon.


lizziel:
Integration tests for GCHP and GC-Classic are in progress.

msulprizio (approved):

Issue #997, opened 2021-11-09
[FEATURE REQUEST] Update surface resistance for ozone deposition to ice/snow


r-pound:
### Request a new GEOS-Chem feature:
Barten et.al 2021 (https://doi.org/10.5194/acp-21-10229-2021) increased the surface resistance of ozone dry deposition to snow and ice to 1E4 s m−1, this results in a reduction of ozone deposition velocity to ice of ≤ 0.01 cm s−1 (~ 66% reduction) which has better agreement with observations and results in improved comparison between Ozone observations in the arctic. 

Following this reasoning we have also increased the surface resistance term in the ozone to ice dry deposition calculation in version 13.1.2, resulting in the following changes to annual average ozone dry deposition velocity

![polar_ice_change](https://user-images.githubusercontent.com/32539507/140938876-b7db27ac-894f-48fe-b74a-d998c91091fc.png)

The current method uses the IS_SNOW(I,J) array to set the resistance velocity for grid boxes which are snow/ice. This however does capture regions of desert as ice due to snow/ice coverage for this array being determined by albedo, resulting in the reduction in deposition velocity also being incorrectly applied to these regions. 

The reduction in ozone dry deposition to ice results in an increase in surface ozone
![surface_ozone_ice_change](https://user-images.githubusercontent.com/32539507/140938989-3b4c5e3a-62dc-4f77-8d90-eaa921a61884.png)

Comparing these to observations shows the small increase in surface ozone improves comparisons to observations in the Antarctic and Arctic. With minimal/no impact outside of these regions.

![obs_com_ice_change](https://user-images.githubusercontent.com/32539507/140939529-ce18a8bc-3cdb-4fa0-9eb6-cb99ed89338c.png)

With either changes to IS_SNOW or an alternative, we believe that these changes should be adopted into the GEOS-Chem model.

yantosca:
Thanks @r-pound.  I've added this to the [GEOS-Chem model development priorities](http://wiki.seas.harvard.edu/geos-chem/index.php/GEOS-Chem_model_development_priorities#Feature_integration_not_yet_prioritized) page.

yantosca:
@r-pound: do you have a corresponding PR that we can use?  Thanks.

r-pound:
@yantosca I've created a pull request (#1013) with the changes discussed above. Let me know if there's any questions/issues with the changes

yantosca:
Note: PR #1013 has been superseded by PR #1055

[...]

PR #2045, opened 2023-11-24
Dry deposition to snow and ice

r-pound:

Name: Ryan Pound
Institution: University of York

- [Y] [Contributing guidelines](https://geos-chem.readthedocs.io/en/stable/help-and-reference/CONTRIBUTING.html)

### Describe the update

With the changes made in 13.4.0 (commit e593ccb, pull request #945) the method of calculating array IsSnow and other majority land cover types for grid boxes changed to use the native GEOS meteorological fields. 

The result of switching from IsSnow being determined by surface albedo to the fraction of land covered with snow (FRSNO) has changed the behaviour of this array. Previously (due to surface albedo) this would be true for any grid box containing a majority of snow, land ice or sea ice. 

This is no longer the case and only represents grid boxes with a majority of snow cover. Several sections of drydep_mod.f90 have historically used the IsSnow array to update dry deposition calculations to account for special cases of dry deposition to snow/ice. However since the land cover arrays were updated to use the native fields, these special case determinations have not been updated to reflect the current state of these arrays, apart from `Update drydep code for assigning pH over land, ocean, and ice` (pull request #1435) which updated the code to reflect the need to use both IsSnow and IsIce to determine if the tile is a majority snow or ice cover. 

### Expected changes

For the dry deposition of ozone, this restores dry deposition velocity calculation behaviour to that expected from #997. This increases surface ozone over predominantly ice and snow-covered regions, with everywhere else experiencing negligible change. 

image

image

This also results in similar changes for other species, however, this should be reverting model behaviour back to that of pre-version 13.4.0 where IsSnow was calculated based on surface albedo. 

I also recommend the inclusion of change in dry deposition plots within the benchmarks. I can supply the Python script used to generate this analysis if that would help.

### DIFF
--- a/GeosCore/drydep_mod.F90
+++ b/GeosCore/drydep_mod.F90
@@ -479,7 +479,7 @@ CONTAINS
           !-----------------------------------------------------------
           ! Special treatment for snow vs. ice
           !-----------------------------------------------------------
-          IF ( State_Met%isSnow(I,J) ) THEN
+          IF ( (State_Met%isSnow(I,J)) .OR. (State_Met%isIce(I,J))) THEN
 
              !-------------------------------------
              ! %%% SURFACE IS SNOW OR ICE %%%
@@ -1465,7 +1465,7 @@ CONTAINS
           !** If the surface to be snow or ice;
           !** set II to 1 instead.
           !
-          IF(State_Met%isSnow(I,J)) II=1
+          IF((State_Met%isSnow(I,J)).OR.(State_Met%isIce(I,J))) II=1
 
           !* Read the internal resistance RI (minimum stomatal resistance for
           !* water vapor,per unit area of leaf) from the IRI array; a '9999'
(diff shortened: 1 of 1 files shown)


msulprizio:
Thanks @r-pound. I have added this bug fix to our model development priority page for inclusion in 14.3.0.

Also tagging Surface-Atmosphere Exchange Working Group Co-Chairs @jgeddes85 @cdholmes @amospktai

GeosCore/drydep_mod.F90
```diff
@@ -479,7 +479,7 @@ SUBROUTINE UPDATE_DRYDEPFREQ( Input_Opt, State_Chm, State_Diag, State_Grid, &
           !-----------------------------------------------------------
           ! Special treatment for snow vs. ice
```
lizziel, line 480: Given this section treats snow and ice the same I think we should update this comment to be "Special treatment for snow and ice". Having "vs" makes it look like they are treated differently.

lizziel:
Good catch that the pre-13.4 version of `IsSnow` included ice and thus there were parts of `drydep_mod.F90` that needed updating but weren't. Thanks for finding this!

GeosCore/drydep_mod.F90
r-pound: Agree. The comment felt a bit ambiguous to me so I left it as is in case anyone more knowledgeable about this specific case wanted to suggest a different approach.  Based on how `IsSnow` used to be defined it will have always treated snow and ice the same until 13.4 so my gut feeling is that it should read 

> Special treatment for snow and ice 

as you suggested

msulprizio (approved):

GeosCore/drydep_mod.F90
msulprizio: This comment is now fixed in c23b1efbe.

PR #2063, opened 2023-12-06
Fix unit conversion code for TOMAS following merge with 14.2.3

msulprizio:

Name: Melissa Sulprizio
Institution: Harvard / GCST

### Describe the update

Compiling with TOMAS=on was failing because the TOMAS code in fullchem_mod.F90 still used character strings for units instead of the integer values that were implemented in 14.3.2. This is now updated.

### Expected changes

This is a zero-difference update that allows GEOS-Chem to compile with the TOMAS option on.


yantosca (approved):
Good to merge!  Thanks @msulprizio.

PR #2079, opened 2023-12-19
Capture Trop and PBL height changes in budgets

cdholmes:
@lizziel, This PR contains only the fix for the budget diagnostics, as described in #2002.

Budget diagnostics now include mass tendency due to changes in
* Tropopause height (TransportTrop budget)
* PBL height (MixingPBL budget)

Name: Chris Holmes
Institution: FSU

### Expected changes

No change expected to benchmarks. See #2002 for more info.

### DIFF
--- a/GeosCore/main.F90
+++ b/GeosCore/main.F90
@@ -1170,6 +1170,25 @@ PROGRAM GEOS_Chem
        ! Do not compute any data in dry-run mode
        IF ( notDryRun ) THEN
 
+          !------------------------------------------------------
+          ! Tropopause height budget diagnostics - Part 1 of 2
+          !------------------------------------------------------
+          IF ( State_Diag%Archive_BudgetTransport ) THEN
+             ! Get initial column masses
+             CALL Compute_Column_Mass( Input_Opt,                           &
+                                     State_Chm, State_Grid, State_Met,    &
+                                     State_Chm%Map_Advect,                &
+                                     .False., &
+                                     State_Diag%Archive_BudgetTransportTrop, &
+                                     .False.,  &
+                                     State_Diag%BudgetMass1,              &
+                                     RC )
+             IF ( RC /= GC_SUCCESS ) THEN
+                ErrMsg = 'Tropopause height budget diagnostics error 1'
+                CALL GC_Error( ErrMsg, RC, ThisLoc )
+             ENDIF
+          ENDIF
+
           ! Interpolate I-3 fields to current dynamic timestep,
           ! based on their values at NSEC and NSEC+N_DYN
           CALL Interp( NSECb,     ELAPSED_TODAY, N_DYN, &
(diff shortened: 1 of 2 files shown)


lizziel:
Thanks for submitting this @cdholmes. Can you provide a detailed summary of the intended computation in main.F? The budget diagnostic code is now very different from 12.9.3 and I will need to rewrite what you have.

cdholmes:
The tropopause height can change in the subroutines `Interp` (for I3 fields), `AirQnt` and `Set_H2O_Trac`, so bracket these subroutine calls in `main.F90` with calls to the budget diagnostic. I saved this mass change in a new  array called `BudgetTransportTropHeight`.

The PBL height can change in`Compute_PBL_Height` in `main.F90`, so bracket that subroutine call with calls to budget diagnostics. (Note that the NLPBL mixing scheme also calls `Compute_PBL_Height`, so the budget calls need to go in `main.F90`, not in `Compute_PBL_Height` itself, unless the redundant calls are eliminated). I saved this mass change in a new array called `BudgetMixingPBLHeight`.

Before `Set_Diagnostics_EndofTimestep`, add `BudgetTransportTropHeight` to `BudgetTransportTrop` and add `BudgetMixingPBLHeight` to `BudgetMixingPBL`.

From my read of new `main.F90` versions, it looks like the same logic should work in 14.2, but I haven't dug deeper into the code to see if any other changes have introduced new places where the tropopause or PBL heights could change.

lizziel:
Thanks, I'll see how I can incorporate this. The main issue is compute_column_mass no longer exists.

cdholmes:
I think the only significant change for recent versions is that `Compute_Budget_Diagnostics` needs to be called with `before=.True.` instead of the first call to `compute_column_mass`.

lizziel:
Yes, the new method is cleaner with fewer lines of code. A few other things:
1. `Compute_Pbl_Height` is now only called in `main.F90` for GC-Classic and `gchp_chunk_mod.F90` for GCHP. The computation for mass change in the PBL due to changing PBL can therefore be within `Compute_Pbl_Height`. I will move it there so that there does not need to be duplicate code for GC-Classic and GCHP.
2. There are other calls to `AirQnt` and `Set_H2O_Trac`. Is it okay that the mass change due to change in tropopause level is only computed once?
3. The two new state_diag arrays are registered, meaning they can be output as their own diagnostics. Is that something you are interested in and would find useful? If not, I can change it so they are not registered.

cdholmes:
Responding to your items:
1. Sounds like a good plan. 
2. In version 12.9, the calls to `AirQnt` and `Set_H2O_Trac` (except those in `main.F90`) were inside other subroutines that were already captured in the budget diagnostics (e.g. chemistry, transport). If that's still the case, then, as you said, we only need to compute the mass change due to tropopause height change once.
3. I don't see any need to output these arrays as diagnostics.

I also noticed that the computational effort for the budget diagnostics could be reduced by half, because we are computing the column masses twice as often as needed. In the current diagnostic scheme, the column mass is computed before and after every process. However, the mass *after* process 1 is the same as the mass *before* process 2, so we could just save the "after" column mass as the "before" for the next process.

lizziel:
> I also noticed that the computational effort for the budget diagnostics could be reduced by half, because we are computing the column masses twice as often as needed. In the current diagnostic scheme, the column mass is computed before and after every process. However, the mass after process 1 is the same as the mass before process 2, so we could just save the "after" column mass as the "before" for the next process.

This goes back to what the budget diagnostics are supposed to be. They are designed to have the capability of bracketing any section of the model, with each bracket independent of the others. If the budgets should instead be computed as checkpoints, e.g. mass differences between set points in the model, then we should take the budget computations out of subroutines and put all calls into the main.F90. This would enable removing the "before" computation of mass without sacrificing clarity of where the previous mass computation was performed.

lizziel:
> In version 12.9, the calls to AirQnt and Set_H2O_Trac (except those in main.F90) were inside other subroutines that were already captured in the budget diagnostics (e.g. chemistry, transport). If that's still the case, then, as you said, we only need to compute the mass change due to tropopause height change once.

There is an additional call to Set_H2O_Trac in the main calling program, right before chemistry: https://github.com/geoschem/geos-chem/blob/3cd802eb4a9b7a226254fc3938778c8db8c07cf2/Interfaces/GCClassic/main.F90#L1599-L1610

lizziel:
See new draft PR at https://github.com/geoschem/geos-chem/pull/2103 which supercedes this.

lizziel:
I am closing this PR since it is superceded by https://github.com/geoschem/geos-chem/pull/2103.

Issue #2110, opened 2024-01-09
GCHP integration tests failing for 14.3.0-alpha.6


msulprizio:
### Name and Institution (Required)

Name: Melissa Sulprizio
Institution: Harvard/GCST

### Description of your issue or question

Integration tests for 14.3.0-alpha.6 and 14.3.0-alpha.7 failed for gchp_merra2_fullchem with the following error:

```
 Getting CH4 boundary conditions in GEOS-Chem from :NOAA_GMD_CH4
  --- Do chemistry now

Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.

Backtrace for this error:
#1  0x986108 in __cldj_interface_mod_MOD_run_cloudj
        at /n/home05/msulprizio/IT/GCHP_14.3.0-alpha.8/CodeDir/src/GCHP_GridComp/GEOSChem_GridComp/geos-chem/GeosCore/cldj_interface_mod.F90:531
#2  0x6d6618 in __photolysis_mod_MOD_do_photolysis
        at /n/home05/msulprizio/IT/GCHP_14.3.0-alpha.8/CodeDir/src/GCHP_GridComp/GEOSChem_GridComp/geos-chem/GeosCore/photolysis_mod.F90:602
...
```

@lizziel wrote:

>Looks like the issue is setting the relative humidity column passed to cloud-J. It isn't initialized for all levels.
>The fix is to find this:

```
       ! Set cloud fraction from input meteorology field                                                                  
       CLDF(1:State_Grid%NZ) = State_Met%CLDF(I,J,1:State_Grid%NZ)
       CLDF(State_Grid%NZ+1) = CLDF(State_Grid%NZ)
```

>and add (removing other RRR settings later): 

```
       ! Set humidity from input meteorology field
       ! Convert relative humidity [unitless] from percent to fraction
       RRR(1:State_Grid%NZ) = State_Met%RH(I,J,1:State_Grid%NZ) / 100.d0
       RRR(State_Grid%NZ+1) = RRR(State_Grid%NZ) * 1.d-1
```

msulprizio:
This is resolved by #2111 which has been merged into the 14.3.0 development branch.

PR #2111, opened 2024-01-09
Add fix to cldj_interface_mod.F90 to avoid floating-point exception in GCHP integration tests

msulprizio:

Name: Melissa Sulprizio
Institution: Harvard/GCST

### Describe the update

The following line in GeosCore/cldj_interface_mod.F90 was causing a floating- point exception error in GCHP integration tests:

```
    ! Set top of atmosphere relative humidity to 10% of layer below
    RRR(State_Grid%NZ+1) = RRR(State_Grid%NZ) * 1.d-1
```

This is resolved by moving the lines for setting RRR to before the loop over the number of layers with clouds.

Please provide a clear and concise overview of the update.

### Expected changes

@lizziel wrote:
>This fix may cause small differences in the results due to a change in relative humidity above level 34 used for retrieving aerosol optical properties. RRR is the relative humidity, which is used only for aerosols in Cloud-J. The way it was before, relative humidity was only assigned up to level 34 which is defined as where there are clouds. In this fix the relative humidity is set for the full column.

### Related Github Issue(s)

- https://github.com/geoschem/geos-chem/issues/2110

### DIFF
--- a/GeosCore/cldj_interface_mod.F90
+++ b/GeosCore/cldj_interface_mod.F90
@@ -483,9 +483,11 @@ CONTAINS
        CLDF(1:State_Grid%NZ) = State_Met%CLDF(I,J,1:State_Grid%NZ)
        CLDF(State_Grid%NZ+1) = CLDF(State_Grid%NZ)
 
-       ! Set humidity from input meteorology field
-       ! Convert relative humidity [unitless] from percent to fraction
+       ! Set relative humidity from input meteorology field and convert
+       ! from percent to fraction
        RRR(1:State_Grid%NZ) = State_Met%RH(I,J,1:State_Grid%NZ) / 100.d0
+
+       ! Set top of atmosphere relative humidity to 10% of layer below
        RRR(State_Grid%NZ+1) = RRR(State_Grid%NZ) * 1.d-1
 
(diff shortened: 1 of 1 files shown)


GeosCore/cldj_interface_mod.F90
```diff
@@ -483,6 +483,11 @@ SUBROUTINE Run_CloudJ( Input_Opt, State_Chm, State_Diag, &
        CLDF(1:State_Grid%NZ) = State_Met%CLDF(I,J,1:State_Grid%NZ)
        CLDF(State_Grid%NZ+1) = CLDF(State_Grid%NZ)
 
+       ! Set humidity from input meteorology field
+       ! Convert relative humidity [unitless] from percent to fraction
```
lizziel, line 487: Add one more comment: "! Set top of atmosphere relative humidity to 10% of layer below"

lizziel (changes requested):

GeosCore/cldj_interface_mod.F90
msulprizio: Thanks for catching that. It's now added in 326c63c.

lizziel (approved):

PR #2078, opened 2023-12-19
Fix PBL level calculations

cdholmes:
@lizziel, This is a PR for the PBL level fix only, as you requested.

Use local scale height and level thickness to determine PBL top level and pressure thickness. The code previously used a fixed global value for scale height, which is too large near poles and too small in tropics. This update lowers the PBL top level in the tropics and raises it near the poles. The change is typically 1 model level or less.

### Name and Institution (Required)

Name: Chris Holmes
Institution: FSU

### Related Github Issue(s)

See #2002 for the expected impact of this change on PBL heights: typically no more than 1 model level up or down.

PR #2211, opened 2024-03-22
Initialize PBL variables

cdholmes:
Fix for runtime error in #2078

Selected PBL variables must be reset to zero each time PBL height is computed.


yantosca (approved):
Thanks @cdholmes!  I thought this might have been an initialization problem.  I'll apply the fix right away.

yantosca:
Integration tests are running

PR #2103, opened 2024-01-04
Capture all tropopause and  PBL height changes in budget diagnostics

lizziel:
### Name and Institution (Required)

Name: Lizzie Lundgren
Institution: Harvard University

### Describe the update

This PR is based on updates for version 12.9.3 provided by Chris Holmes (@cdholmes) in https://github.com/geoschem/geos-chem/pull/2079.  Budget diagnostics now include mass tendency due to changes in:

- Tropopause height (TransportTrop budget)
- PBL height (MixingPBL budget)

I expanded on the updates by implementing an accumulation option for budget diagnostics and applied it to the tropopause height changes incorporated into the transport budget diagnostic for GC-Classic. Species mass change in the troposphere due to changes in the tropopause heights are now accumulated over the timestep in GC-Classic when AirQnt is called outside of an existing budget diagnostic block. This specifically includes one call in main.F (when meteorology is updated due to new timestep) and all calls from Set_H2O_Trac (when meteorology is updated due to a change in moisture). By default calling AirQnt will not result in updating the transport trop ht budget array; it must be explicitly specific in the arguments list that this should be done.

### Expected changes

The PBL mixing budget diagnostic will change for both GC-Classic and GCHP. The transport trop budget diagnostic will change for GC-Classic only (we do not use it for GCHP since tpcore is not used).

Species mass changes in the PBL and troposphere due to mixing and transport respectively will now include fluctuations in species mass at the top due to changing definition of the vertical region top level. For example, if the PBL top level changes from 26 to 25 in timestep A in the call to Compute_PBL_Height, then the mixing budget for timestep A will be the change in species mass due to mixing for levels 1 to 25 minus the species mass in level 26. This is because level 26 is no longer included in the PBL, representing a mass loss for the mixing diagnostic. If PBL top level increases rather than decreases then the species mass in the top grid cell is added.

### Reference(s)

See also https://github.com/geoschem/geos-chem/pull/2002 for initial discussion.

### Related Github Issue(s)

None

PR #2215, opened 2024-03-23
Capture all trop pbl ht changes in budget diags

cdholmes:
This PR simplifies the updated budget diagnostics developed in PR #2103, removing the distinction that some accumulated while others didn't. 

With this update, all of the budget diagnostics accumulate over the dynamic time step, giving more consistency in how the code looks and works. The extra arrays for tropopause height and PBL height budgets were then eliminated. 

I verified that the diagnostics are capturing all changes in tracer mass, with the exception of H2O, CLOCK and perhaps some other specialty simulation tracers. d

I added a budget consistency check. This prints a message global total tracer mass changes in code that is not between the "before" and "after" calls to Compute_Budget_Diagnostic. This could indicate a bug or, perhaps, that a some future feature requires updating the budget diagnostic. I think that this could be useful for unit testing. However, simulations that include H2O or CLOCK will trigger these messages, even when they run correctly. This could be fixed by moving the CLOCK updates inside chemistry or excluding H2O from the consistency check. For that reason, it could make sense to run the budget consistency check only in DEBUG simulations.

### DIFF
--- a/GeosCore/diagnostics_mod.F90
+++ b/GeosCore/diagnostics_mod.F90
@@ -226,24 +226,6 @@ CONTAINS
     ENDIF
 
     !------------------------------------------------------------------------
-    ! Budget diagnostic adjustment for changes in PBL heigh†
-    ! (mixing budget in PBL only)
-    !------------------------------------------------------------------------
-    IF ( State_Diag%Archive_BudgetMixingPBL ) THEN
-       State_Diag%BudgetMixingPBL = State_Diag%BudgetMixingPBL              &
-            + State_Diag%BudgetMixingPBLHeight
-    ENDIF
-
-    !------------------------------------------------------------------------
-    ! Budget diagnostic adjustment for changes in tropopause height
-    ! (transport budget in trop only)
-    !------------------------------------------------------------------------
-    IF ( State_Diag%Archive_BudgetTransportTrop ) THEN
-       State_Diag%BudgetTransportTrop = State_Diag%BudgetTransportTrop      &
-            + State_Diag%BudgetTransportTropHeight
-    ENDIF
-
-    !------------------------------------------------------------------------
     ! Diagnostics for the mercury and tagged mercury simulations
     !------------------------------------------------------------------------
     IF ( Input_Opt%ITS_A_MERCURY_SIM ) THEN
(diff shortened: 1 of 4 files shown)


GeosCore/diagnostics_mod.F90
```diff
@@ -405,11 +387,48 @@ SUBROUTINE Zero_Diagnostics_StartofTimestep( Input_Opt, State_Diag, RC )
...
+         State_Diag%BudgetConvectionTrop = 0.0_f8
+    IF ( State_Diag%Archive_BudgetConvectionPBL ) &
+         State_Diag%BudgetConvectionPBL = 0.0_f8
+
+    IF ( State_Diag%Archive_BudgetChemistryFull ) &
+         State_Diag%BudgetChemistryFull = 0.0_f8
+    IF ( State_Diag%Archive_BudgetChemistryTrop ) &
+         State_Diag%BudgetChemistryTrop = 0.0_f8
+    IF ( State_Diag%Archive_BudgetChemistryPBL ) &
+         State_Diag%BudgetChemistryPBL = 0.0_f8
```
lizziel, line 417: Chemistry is not called every timestep but here the array is set to zero every timestep. This is going to have implications for time-averages since the history package includes values for every timestep in its computation.

yantosca:
Thanks @cdholmes.  Does this PR totally supersede #2103 or do we need to merge #2103 first, and then this PR?

GeosCore/diagnostics_mod.F90
```diff
@@ -1301,6 +1311,37 @@ SUBROUTINE Compute_Budget_Diagnostics( Input_Opt,   State_Chm, State_Grid, &
...
+       ! Relative error, change since prior "after" call
+       rerr(1) = abs( sum( colMass(:,:,:,1) ) / PriorGlobalMass(1) - 1 )
+       rerr(2) = abs( sum( colMass(:,:,:,2) ) / PriorGlobalMass(2) - 1 )
+       rerr(3) = abs( sum( colMass(:,:,:,3) ) / PriorGlobalMass(3) - 1 )
+       
+       ! Relative error tolarance, errors larger than this will halt the model
+       rtol = 0.001
+
+       IF ( rerr(1) > rtol ) THEN
+         print*,'Budget Diagnostic Error: FULL-column tracer mass changed unexpectedly, rerr=',rerr(1)
```
lizziel: Should this error checking be on by default? We could wrap it in the debug verbose setting in geoschem_config.yml, or would you rather it always be executed?

GeosCore/diagnostics_mod.F90
```diff
@@ -1301,6 +1311,37 @@ SUBROUTINE Compute_Budget_Diagnostics( Input_Opt,   State_Chm, State_Grid, &
...
+       ENDIF
+       IF ( rerr(2) > rtol ) THEN
+         print*,'Budget Diagnostic Error: TROP-column tracer mass changed unexpectedly, rerr=', rerr(2)
```
lizziel: All of these prints need to be wrapper in Input_Opt%amIRoot logical to avoid prints from all cores when using MPI.

cdholmes:
You'll need to merge both PR #2103 and this PR.

GeosCore/diagnostics_mod.F90
lizziel: This can be fixed by separating out the components that only get executed during chemistry timesteps. However, we cannot use the time_mod.F90 functions to determine if it is a chemistry timestep or not since those functions do not work in non-classic models (including GCHP which uses ESMF for time). Passing a logical for isChemTime would work.

lizziel (changes requested):
Hi @cdholmes, I like these changes. I left several comments.

cdholmes:
@lizziel, All of your suggestions sound good to me, but I can't easily do them for a few weeks, as I haven't written much MPI code or dealt with debug compiler flags. Are you able to implement what you suggested?

lizziel:
Yes, I will merge in your updates and make the changes to my branch.

lizziel (approved):
I will incorporate my suggestions post-merge.

lizziel:
Changes now incorporated into https://github.com/geoschem/geos-chem/pull/2103

PR #2254, opened 2024-04-18
Fix bug in usage of FRSNO met-field

lizziel:

Name: Lizzie Lundgren

### Describe the update

This PR fixes a bug where `State_Met%FRSNO` is inconsistently used. In some places it is used as fraction of grid box with snow covered land, while in other places it is used as fraction of land with snow cover. The GMAO meteorology field input to the model is defined as fraction of land with snow cover. To fix the issue we now define `State_Met%FRSNO` (renamed to `State_Met%FRSNOW`) as fraction grid box with snow cover on land, and compute it upon file read by multiplying the `FRSNO` import with fraction land import (`FRLAND`). Usage of the met-field throughout the model is now updated for consistency. 

Also in this update is renaming `State_Met%LANDIC` to `State_Met%LANDICE` for clarity, and adding diagnostics for `State_Met%IsWater`, `State_Met%IsLand`, `State_Met%IsIce`, and `State_Met%IsSnow` for sanity checking. The surface type variables are type logical but the diagnostics are stored in `State_Diag` as `REAL4`.

### Expected changes

This bug fix has the most impact for GCHP runs that use raw GMAO meteorology files. The raw files have missing values for all non-land grid boxes and `FRSNO` was set to 1e15 in those locations, resulting in incorrect logical masks for water, snow, and ice. The new multiplication by `FRLAND` removes that issue.

Mercury simulations are impacted because `State_Met%FRSNO` was used as if it were fraction grid box rather than fraction land.

Very small differences may be seen in GC-Classic full chemistry runs and GCHP runs that use processed meteorology files due to small changes in time-varying `State_Met%FRSNOW` when computing surface type masks (water, snow, ice, land). These changes would only be where snow is present.

### Reference(s)

None

### Related Github Issue(s)

- closes https://github.com/geoschem/GCHP/issues/406


GeosCore/calc_met_mod.F90
```diff
@@ -293,24 +293,23 @@ SUBROUTINE AIRQNT( Input_Opt, State_Chm, State_Grid, State_Met, &
        IsLocNoon(I,J) = ( LocTimeSec(I,J)          <= 43200  .and. &
                           LocTimeSec(I,J) + Dt_Sec >= 43200 )
 
+       ! Snow
+       FRSNOW = State_Met%FRSNOW(I,J)
+
        ! Land without snow or ice
-       FRLAND_NOSNO_NOICE = State_Met%FRLAND(I,J) -                &
-               (State_Met%FRLAND(I,J)*State_Met%FRSNO(I,J))
+       FRLAND_NOSNOW_NOICE = State_Met%FRLAND(I,J) - State_Met%FRSNOW(I,J)
```
msulprizio, line 300: If this is land without snow or ice, why are we only subtracting FRSNOW? Should we also subtract FRLANDICE?

GeosCore/flexgrid_read_mod.F90
```diff
@@ -292,7 +293,18 @@ SUBROUTINE FlexGrid_Read_A1( YYYYMMDD, HHMMSS, Input_Opt, State_Grid, &
     ! Read FRSNO
     v_name = "FRSNO"
     CALL Get_Met_2D( Input_Opt, State_Grid, Q, TRIM(v_name), t_index=t_index )
-    State_Met%FRSNO = Q
+
+    ! The import FRSNO is fraction of land with snow cover. Convert to
```
msulprizio, line 297: Change FRSNO to FRSNOW

Headers/state_met_mod.F90
```diff
@@ -4566,32 +4569,32 @@ SUBROUTINE Get_Metadata_State_Met( am_I_Root, metadataID, Found,   RC,     &
...
-          IF ( isDesc  ) Desc  = 'Fraction of land ice'
+       CASE ( 'FRLANDICE' )
+          IF ( isDesc  ) Desc  = 'Fraction of land ice type in grid box'
           IF ( isUnits ) Units = '1'
           IF ( isRank  ) Rank  = 2
 
        CASE ( 'FROCEAN' )
           IF ( isUnits ) Units = '1'
-          IF ( isDesc  ) Desc  = 'Fraction of ocean'
+          IF ( isDesc  ) Desc  = 'Fraction of ocean type in grid box'
```
msulprizio, line 4588: I suggest removing "type" from the descriptions here. It makes it seem like different categories of lakes, land, ice, and ocean are identified.

Interfaces/GCHP/Includes_Before_Run.H
```diff
@@ -63,6 +62,14 @@
   State_Met%V10M                = V10M                       ! m/s
   State_Met%Z0                  = Z0                         ! m
 
+  ! The import FRSNO is fraction of land with snow cover. Convert to
```
msulprizio, line 65: Change FRSNO to FRSNOW

GeosCore/calc_met_mod.F90
```diff
@@ -293,24 +293,23 @@ SUBROUTINE AIRQNT( Input_Opt, State_Chm, State_Grid, State_Met, &
...
-       FRSNO = State_Met%FRSNO(I,J)
-      
+       FRICE = State_Met%FRLANDICE(I,J) + State_Met%FRSEAICE(I,J)
+
        ! Set IsLand, IsWater, IsIce, IsSnow based on max fractional area
-       State_Met%IsLand(I,J)  = (FRLAND_NOSNO_NOICE > MAX(FRWATER, FRICE, FRSNO))
-       State_Met%IsWater(I,J) = (FRWATER > MAX(FRLAND_NOSNO_NOICE, FRICE, FRSNO))
-       State_Met%IsIce(I,J)   = (FRICE > MAX(FRLAND_NOSNO_NOICE, FRWATER, FRSNO))
-       State_Met%IsSnow(I,J)  = (FRSNO > MAX(FRLAND_NOSNO_NOICE, FRWATER, FRICE))
+       State_Met%IsLand(I,J)  = (FRLAND_NOSNOW_NOICE > MAX(FRWATER, FRICE, FRSNOW))
```
msulprizio, line 309: Shouldn't the IsLand logical include snow-covered land? Therefore should this line instead be:

```
State_Met%IsLand(I,J) = (FRLAND > MAX(FRWATER, FRICE))
```

Interfaces/GCHP/Includes_Before_Run.H
lizziel: FRSNO is the name of the import from the file so needs to stay FRSNO in the comment.

GeosCore/flexgrid_read_mod.F90
lizziel: FRSNO is still used in the file and is only renamed for State_Met.

Headers/state_met_mod.F90
lizziel: This is the definition used in the GEOS-FP specification document but it makes sense to remove.

GeosCore/calc_met_mod.F90
lizziel: The intention is for land, ice, water, and snow to be separate categories. This met-field only appears in the code in land_mercury_mod.F90, in a commented out subroutine. So perhaps we can delete it to avoid confusion.

GeosCore/calc_met_mod.F90
lizziel: Good point. I'll look into the history of this and see what it needs to be based on usage.

GeosCore/calc_met_mod.F90
msulprizio: I think that might be best -- otherwise we'd need to rename it to something like IsLand_NoSnow to be clear

Headers/state_met_mod.F90
msulprizio: Thanks for clarifying

GeosCore/calc_met_mod.F90
lizziel: I figured out the history for this. `FRLAND_NOSNOW_NOICE` used to subtract `FRLANDIC` but a while back I discovered that was a bug and fixed it in 14.1.0 (https://github.com/geoschem/geos-chem/pull/1425). `FRLAND` in the constants file does not include ice over land, hence there is a separate entry `FRLANDIC` in the constants file. `FRLANDIC` is specifically glaciers since the source is constant. `FRSNOW` is time-varying and so we subtract it from `FRLAND`. I will update the code to be more clear about this.

GeosCore/calc_met_mod.F90
lizziel: Actually, I just found it used in drydep_mod. Since these four logical masks are supposed to mutually exclusive (only one true for each grid box), I could create an index variable similar to how we used to use LWI.

GeosCore/calc_met_mod.F90
yantosca: @msulprizio @lizziel: I think this goes back to about GEOS-5 when the LWI fields were superseded by the FR* fields.  Also it should be FRLANDICE but at the time we could only use 8 chars for the container name in the bpch file, so we lopped off the ending "E".

msulprizio (approved):

yantosca:
Integration tests are running

yantosca:
All GEOS-Chem Classic execution tests passed:
```console

GCClassic #2fa1736 Update RTD docs for PR #2266 (Remove obsolete carbon sim options)
GEOS-Chem #203a480b6 Merge PR #2254 (Fix bug in usage of FRSNO met field)
HEMCO     #fddcae5 Update version to 3.8.1 for release

Using 24 OpenMP threads
Number of execution tests: 29

Submitted as SLURM job: 29651290

```
All GCHP integration tests passed:
```console
GCHP: Execution Test Results

GCHP      #e86ba15 GEOS-Chem submod update: Merge PR #2266 (Remove unused carbon sim options)
GEOS-Chem #203a480b6 Merge PR #2254 (Fix bug in usage of FRSNO met field)
HEMCO     #fddcae5 Update version to 3.8.1 for release

Number of execution tests: 11

Submitted as SLURM job: 29652416

```

PR #2366, opened 2024-07-09
Changes for coupling GEOS-Chem to CESM in Spectral-element Dynamical Core (CAM-SE)

jimmielin:

Name: Haipeng Lin
Institution: Harvard Univ.

### Describe the update

These updates are structural and affect GEOS-Chem within CESM model only. They are intended to allow the GEOS-Chem chemistry to work using the spectral element dynamical core in the Community Atmosphere Model, the atmospheric component in CESM.

### Expected changes

Zero-diff.

### Reference(s)

N/A

### Related Github Issue

Fixes issues:
* https://github.com/geoschem/CAM/issues/36

Companion PR in geoschem/CAM: https://github.com/geoschem/CAM/pull/37

### DIFF
--- a/Headers/state_grid_mod.F90
+++ b/Headers/state_grid_mod.F90
@@ -37,8 +37,9 @@ MODULE State_Grid_Mod
      !----------------------------------------
      ! User-defined grid fields
      !----------------------------------------
-#if defined( MODEL_WRF )
+#if defined( MODEL_WRF ) || defined( MODEL_CESM )
      INTEGER            :: ID          ! Grid identifier number
+     INTEGER            :: ROOT_ID     ! Root grid identifier number (used in multi-domain-per-PET)
 #endif
      CHARACTER(LEN=255) :: GridRes     ! Grid resolution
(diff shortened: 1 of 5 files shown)


Headers/state_grid_mod.F90
```diff
@@ -37,8 +37,9 @@ MODULE State_Grid_Mod
      !----------------------------------------
      ! User-defined grid fields
      !----------------------------------------
-#if defined( MODEL_WRF )
+#if defined( MODEL_WRF ) || defined( MODEL_CESM )
```
lizziel, line 40: Could you add a clarifying comment in this block with more about what these are? Does each CPU have a different ID while root ID is the same for all? Or does each domain have a fixed ID and is handled by multiple CPUs such that multiple CPUs have the same ID?

lizziel (changes requested):

jimmielin:
Thank you, changelog updated and the trailing slash change reverted.

lizziel (approved):
Looks good to merge.

PR #2379, opened 2024-07-15
Disable support for FAST-JX except for the Hg simulation

yantosca:
### Name and Institution (Required)

Name: Bob Yantosca
Institution: Harvard + GCST

### Describe the update
This is a corresponding PR to https://github.com/geoschem/GCClassic/pull/62 and https://github.com/geoschem/GCHP/pull/426, which are intended to disable support for FAST-JX except for the Hg simulation.

We have added an error trap in `GeosCore/gc_environment_mod.F90`, right before the call to `Init_CMN_FJX`, which will stop the run if FAST-JX is enabled for simulations using the `fullchem`, `custom`, or `carbon` KPP mechanisms.

### Expected changes
This will be a no-diff-to-benchmark update.  An error message will halt simulations (other than Hg) that use FAST-JX photolysis, with error output such as:
```console
*************   S T A R T I N G   G E O S - C H E M   *************

===> Mode of operation         : GEOS-Chem "Classic"
===> GEOS-Chem version         : 14.4.1
===> Compiler                  : GNU Fortran compiler (aka gfortran)
===> Flexible precision set to : 8-byte real (aka REAL*8)
===> Parallelization w/ OpenMP : ON
===> netCDF diagnostics        : ON
===> netCDF file compression   : SUPPORTED
===> Luo et al (2019) wetdep?  : OFF

===> SIMULATION START TIME: 2024/07/15 14:54 <===

===============================================================================
G E O S - C H E M   U S E R   I N P U T

READ_INPUT_FILE: Opening ./geoschem_config.yml

SIMULATION SETTINGS
-------------------
Simulation name             : fullchem
CHEM_INPUTS directory       : /n/holyscratch01/external_repos/GEOS-CHEM/gcgrid/data/ExtData/CHEM_INPUTS/
Species database file       : ./species_database.yml
Turn on verbose output      :     F
Verbose output printed on   : root core only
Start time of run           : 20190701 000000
End time of run             : 20190801 000000
Data Directory              : /n/holyscratch01/external_repos/GEOS-CHEM/gcgrid/data/ExtData/
Meteorology field           : MERRA2
Turn on GEOS-Chem timers    :     T

GRID SETTINGS
------------
Grid resolution             : 4.0x5.0
Min/max longitude           :  -180.0000   180.0000
Min/max latitude            :   -90.0000    90.0000
X grid dimension            :    72
Y grid dimension            :    46
Z grid dimension            :    72
Use half-sized polar boxes? :     T
[...]

PR #2380, opened 2024-07-15
Disallow legacy FAST-JX unless using mercury simulation

lizziel:

Name: Lizzie Lundgren

### Describe the update

This is a companion PR to https://github.com/geoschem/GCClassic/pull/62 and https://github.com/geoschem/GCHP/pull/426. Those PRs update GC-Classic and GCHP CMake files to throw an error if the FASTJX compile option is selected for any mechanism except Hg. For other models that use GEOS-Chem we also need to have a run-time error if using legacy FAST-JX with any simulation other than mercury. This prevents having to update build settings for all external models that use GEOS-Chem.

### Expected changes

This is a no diff update.

### Reference(s)

None

### Related Github Issues/PRs

https://github.com/geoschem/GCHP/pull/426
https://github.com/geoschem/GCClassic/pull/62


lizziel:
Closing since duplicate with https://github.com/geoschem/geos-chem/pull/2379

PR #2467, opened 2024-09-19
Fix dry mass fraction of seasalt in aerosol thermodynamics

lizziel:
### Name and Institution (Required)

Name: Lizzie Lundgren (Harvard/GCST), on behalf of Yuk Chun Chan (UW) and Becky Alexander (UW)

### Describe the update

This PR applies two corrections to the dry mass ratio of seasalt that is hard-coded in aerosol thermodynamics. The changes are updates submitted by Yuk Chun Chan (UW) and Becky Alexander (UW):
    
From Yuk Chun Chan:

> To account for the missing Ca, Mg, and K, the Na-to-seasalt mass ratio was artificially increased from 0.3061 (i.e., seawater average) to 0.397 in previous versions. Since we are re-introducing Ca, Mg, and K to the model now, we should adjust the Na-to-SS mass ratio from 0.397 back to 0.3061. Otherwise, we would get excess amount of cations in the model.

    
From Becky Alexander

> The TNA equation needs "to be changed, which deals with coarse-mode sea salt aerosol. The current value is 0.378. I don't know why these values are different for fine and coarse mode. They should both be 0.3061, which is the dry mass fraction of sea salt.

These changes will go into 14.5.0 on top of earlier updates in this version to add back in cations Ca, Mg, and K for aerosol theremodynamics.

### Expected changes

Tagging @yc-chan and @beckyalexander to comment on expected changes.

### Reference(s)

None

### Related Github Issue

closes https://github.com/geoschem/geos-chem/issues/2391

Issue #2391, opened 2024-07-18
Comment back in Ca, Mg, and K cations in aerosol thermodynamics


msulprizio:
### Your name

Melissa Sulprizio

### Your affiliation

Harvard

### Provide a clear and concise overview of the new feature requested.

These species are commented out in ISORROPIA and HETP and should be uncommented.

Tagging @beckyalexander for more details and a better estimate of delivery timeframe.

### Will you be implementing this feature yourself?

No

lizziel:
@msulprizio you beat me to creating this. I discussed this issue with @beckyalexander and I volunteered to make the update. I can put it into 14.5.0.

yantosca:
@msulprizio @lizziel @beckyalexander: The benchmark plots are available at https://gc-dashboard.org/difference?primary_key=diff-gcc-4x5-1Mon-14.4.2-gcc-4x5-1Mon-14.5.0-alpha.2

msulprizio:
This has been addressed by #2398

yc-chan:
@beckyalexander and I found a potential mass-balance issue in this implementation.

To account for the missing Ca, Mg, and K, the Na-to-seasalt mass ratio was artificially increased from 0.3061 (i.e. seawater average) to 0.397 in previous versions. 

Since we are re-introducing Ca, Mg, and K to the model now, we should adjust the Na-to-SS mass ratio from 0.397 back to 0.3061. Otherwise, we would get excess amount of cations in the model.

lizziel:
Thanks for reporting this. Can you confirm this is the code that is affected? https://github.com/geoschem/geos-chem/blob/bef56c605e018eecbd91646a51ce82c7cd77f56a/GeosCore/aerosol_thermodynamics_mod.F90#L639-L646

Also please confirm the fix is as follows:
1. Delete the currently uncommented code which uses 0.397
2. Uncomment the currently commented out code that uses 0.3061

Note that the equations are different in other ways than the scale factor. Please let us know what the correct equation should be.

lizziel:
@yc-chan, @beckyalexander - We plan to freeze 14.5.0 in the next few days and begin benchmarking. Please let us know exactly what the fix is as soon as possible to ensure it gets into this version. Otherwise it will need to wait for 14.6.0.

beckyalexander:
[...]

PR #2489, opened 2024-09-30
Fix a bug in coarse-mode seasalt Ca, Mg, and K

yc-chan:

Yuk Chun Chan (UW) and Becky Alexander (UW)

### Describe the update

In the past, GC only called ISORROPIA for doing fine-mode-aerosol calculations, so it assumed all the Ca, Mg, and K in aerosols are from fine-mode seasalt. However, DO_ATE is now used for coarse-mode-aerosol calculations as well so we have to update the code for that.

After this bug fix, GC will use SALC instead of SALA to calculate the contribution of seasalt to coarse-mode Ca, Mg, and K. Note that the alkalinity factor AlkR is now included in the formulas to make sure these cations are treated the same way as Na.

### Expected changes

Without this fix, I expect that GC will likely underestimate cation abundance in coarse-mode aerosols in marine boundary layer, especially after [the previous correction to Na-to-seasalt mass ratio](https://github.com/geoschem/geos-chem/pull/2467).

### Related Github Issue

[Comment back in Ca, Mg, and K cations in aerosol thermodynamics #2391
](https://github.com/geoschem/geos-chem/issues/2391)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Changed
 - Copy values from `State_Chm%KPP_AbsTol` to `ATOL` and `State_Chm%KPP_RelTol` to `RTOL` for fullchem and Hg simulations
-- Changed previously zero Ca2, K, and Mg cation values passed to HETP to scaled SALA species concentrations and updated Na:seasalt mass fraction accordingly
+- Introduced seasalt Ca, K, Mg back to aerosol thermodynamics via HETP.
 - Updated `HEMCO_Config.rc.fullchem` (GCClassic + GCHP) and `ExtData.rc` to add emissons of new species from Travis et al 2023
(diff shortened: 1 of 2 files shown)


lizziel (commented):
These changes look fine to me but I wonder if we also want to edit the existing description in the Changelog. We currently have one entry for adding back in the cations: "Changed previously zero Ca2, K, and Mg cation values passed to HETP to scaled SALA species concentrations and updated Na:seasalt mass fraction accordingly".   @yc-chan, what do you recommend to accurately describe all of the updates related to the cations?

yc-chan:
What about "Introduced seasalt Ca, K, Mg back to aerosol thermodynamics via HETP." ? 

I am not familiar with the convention for Changelog. How detailed do we need usually?

lizziel:
That update sounds good. Could you push the change to your branch?

yc-chan:
Done!

lizziel (approved):

msulprizio:
This pull request passes all GCClassic and GCHP integration tests

Issue #2014, opened 2023-10-26
Tracer mass in restart files differs from the running model


cdholmes:
### Name and Institution (Required)

Name: Christopher Holmes
Institution: FSU

### Description of your issue or question

The global total tracer mass contained in restart files is not the same as the mass in the model. I have documented this problem in GCClassic version 14.2 and 12.9. A consequence of this is that tracer mass is not conserved when the model integration is stopped and restarted. 

### Steps to reproduce

The following directions apply to GCClassic 14.2. 

1. I added the following code to print the global total tracer mass at the end of subroutine Compute_Budget_Diagnostics in diagnostics_mod.F90. (version 14.2)
```fortran
    if ( after ) then
    ! Pick a tracer of interest (N=2 is ACTA in fullchem)
    N=2
    ! Print global total tracer mass before and after budget diagnostic, and the change between the two
    print'(A10,3(F20.6))','**BUDGET',
                      sum(colMass(:,:,N,1)), &
                      sum( State_Chm%Species(N)%Conc(:,:,:) * &
                           State_Met%AD(:,:,:) ), &
                      sum( State_Chm%Species(N)%Conc(:,:,:) * &
                           State_Met%AD(:,:,:) ) - &
                      sum(colMass(:,:,N,1))
    endif
``` 
The global tracer mass in the final restart file should be as it was after the last call to `Compute_Budget_Diagnostics` (after wet deposition). 

2. I added `Met_AD` to the Restart collection in HISTORY.rc, so that the restart file contains the dry air mass.
3. The Python code below reads restart files and prints the global total dry air mass and the global total tracer mass.
```Python
# Print global total tracer mass
import xarray as xr
from glob import glob

# Open species database, for molar weights
import yaml
with open('species_database.yml', 'r') as file:
  speciesdata = yaml.safe_load(file)
# Molar weight in grams is speciesdata['speciesname']['MW_g']

# Concentrations at beginning and end of simulation
restart0 = xr.open_dataset('Restarts/GEOSChem.Restart.20190701_0000z.nc4').isel(time=0)
restart1 = xr.open_dataset('Restarts/GEOSChem.Restart.20190701_0100z.nc4').isel(time=0)

# Dry airmass computed from DELPDRY (delpdry units are hPa), kg
restart0['airmass_dry'] = restart0.Met_DELPDRY * restart0.AREA / 9.81 * 100
[...]

Issue #111, opened 2019-10-24
[FEATURE REQUEST] Investigate saving REAL*8 fields to the restart file


yantosca:
## Overview
We are looking into removing differences between GEOS-Chem runs that execute over a long period, and runs that are restarted.  One of the reasons for these differences might be insufficient precision in the GEOS-Chem restart files.  We can experiment with saving REAL*8 data to the SpeciesRst collection and see if this helps to improve the situation.

yantosca:
While the History diagnostics had derived types that would allow for carrying diagnostics as REAL*8, the I/O code (in `History/history_netcdf_mod.F90` would cast everything to REAL*4 and then write it to netCDF.  This situation has been corrected in commit 652bb239abed2b12e4331ca0de18d0a3db2de7f7.  Several modifications were made to the following modules:

- `Headers/registry_mod.F90`
- `Headers/state_diag_mod.F90`
- `History/history_mod.F90`
- `History/history_netcdf_mod.F90`

Specifically, we now have included a variable within derived-type objects that indicates the KIND value (i..e. REAL*8 or REAL*4) for the netCDF output.  Prior to this, it was always assumed that netCDF variables would be written to disk with the same KIND value with which they were defined.

NOTE: Similar modifications need to be made in HEMCO in order to read in REAL*8 data.  Right now HEMCO can only read in REAL*4 data from netCDF files.

msulprizio:
This will likely go into an early GEOS-Chem 13 version

yantosca:
The GCC History can now save output as REAL*4 or REAL*8 (cf. https://github.com/geoschem/geos-chem/commit/6e004b7a9cb0d1dbd09e4142e79a5b984abf0f2e merge).

However, at present, HEMCO can still only read and carry REAL*4 data in its data structures.  This needs to be resolved first before this issue is closed.

yantosca:
The remaining work needs to be done on the HEMCO side, as the GEOS-Chem "Classic" History component can now write REAL\*8 fields to disk.  

Therefore, I am closing out this issue, and have opened a new feature request (https://github.com/geoschem/HEMCO/issues/69) on the HEMCO repository issue tracer,

PR #2521, opened 2024-10-15
Fixes towards achieving bit-for-bit reproducibility in GC-Classic upon restart

lizziel:

Name: Lizzie Lundgren

### Describe the update

This PR applies changes in GC-Classic needed to achieve bit-for-bit reproducibility when breaking up a run in time. Please note that with these updates out-of-the-box GC-Classic will still not have reproducibility upon restart due to a limitation of HEMCO. HEMCO reads inputs, including GEOS-Chem restart file variables, as single precision when most of these fields in GEOS-Chem are double precision.

Separately I implemented an update that reads the GEOS-Chem restart file locally, preserving native precision found in the file. With this PR and that update on top of it I am able to achieve bit-for-bit reproducibility in GC-Classic full chemistry benchmark simulation when breaking up the run in time. We are not going to bring in the update to read the restart file locally because we need to use HEMCO for its flexgrid capability. However, I will make that update available as a dev branch should anyone want to use it (see [here](https://github.com/geoschem/geos-chem/tree/feature/bypass_hemco_to_read_gcclassic_restart_as_real8)). An update to HEMCO to read native precision will come later.

Updates within GEOS-Chem that change benchmark output:
1. Fix a parallelization issue in GEOS-Chem's computation of `State_Chm%DryDepNitrogen` used in the HEMCO soil NOx extension and saved out to the GEOS-Chem restart file. I believe this was previously undetected because the differences are at the level of double precision numerical noise and thus were previously obscured by outputting to file as single precision. This update includes writing to the restart and diagnostic files using native precision which exposed the issue.
2. Fix bug in `Chem_ORVCsesq` restart file entry in GC-Classic fullchem `HEMCO_Config.rc`.
3. Register `State_Met` and `State_Chm` arrays in GC-Classic History using the native precision used in the model. Previously the precision was unspecified and outputs thus used the default of single precision.
4. Specify `REAL*8` as the output precision of `State_Diag%SpeciesRst` which is the species concentration array written to the restart file GC-Classic. Previously it was set to `REAL*4`.
5. Convert species concentration units of mol/mol <-> kg/kg in main.F90. This update ensures error is not introduced due to unit conversions used for saving species to the restart file in mol/mol. Conversion to mol/mol must occur every timestep as part of updating diagnostics used for restart variables so that GEOS-Chem Classic can achieve bit-for-bit reproducibility upon breaking up runs. We still use kg/kg dry air throughout main.F90. Conversion from mol/mol to kg/kg occurs at the start of every timestep. This also means that mol/mol is not converted to kg/kg in the subroutine that retrieves restart variables from HEMCO during initialization since it will occur later.
6.   Move setting prescribed CH4 to just after emissions in GC-Classic. It is necessary to move where prescribed CH4 is set because it is dependent on PBL height and previously PBL height was not updated until after Set_CH4 was called. This introduced small differences in GC-Classic when breaking up runs. CH4 is now set using updated PBL height and is grouped with emissions where other prescribed surface VMRs are applied. It is still applied outside of emissions.

This PR also contains several no-diff updates:
- Fix minor error in checking variables after `State_Chm` array allocation
- Remove unused local variables in `ucx_mod`
- Move getting restart var `DELPDRY` to subroutine `GC_Get_Restart`. This is a no diff update but now allows print of min, max, sum of restart variable `Met_DELPDRY` to be included with all of the other restart variables retrieved from HEMCO into GEOS-Chem for use upon startup.
- Add `GeosUtil/print_mod.F90` for specialized model print options. This module currently has one subroutine only, to print min, max, and mean of species concentrations to log. It currently only prints if root thread and does not convert units. Optional arguments allow printing one species or a consecutive sequence of species indexes. The current units are printed to the log.
- Clean up prints in `Get_GC_Restart` subroutine so that all species initial concentrations are printed to log, including if background values are used (with note about that), as well as initial concentrations of other arrays in the restart file that are read into the model.
- No diff fixes to `SOIL_WETDEP` and `SOIL_DRYDEP` subroutines used to update `State_Chm%WetDepNitrogen` and `State_Chm%DryDepNitrogen` used in the Soil NOx extension in HEMCO. Changes include removing the unused level argument and uncommenting a commented out Soil NOx extension logical used to determine whether to call `SOIL_WETDEP` in `Do_Washout_as_Sfc`.

### Expected changes

This update will cause small differences in all GEOS-Chem Classic simulations.

### Reference(s)

None

### Related Github Issue

https://github.com/geoschem/HEMCO/issues/18
https://github.com/geoschem/geos-chem/issues/111
https://github.com/geoschem/geos-chem/issues/2014

### DIFF
--- a/GeosCore/hco_interface_gc_mod.F90
+++ b/GeosCore/hco_interface_gc_mod.F90
@@ -4385,22 +4385,6 @@ CONTAINS
       D = GET_FIRST_I3_TIME()
       CALL FlexGrid_Read_I3_1( D(1), D(2), Input_Opt, State_Grid, State_Met )
 
-      ! Get delta pressure per grid box stored in restart file to allow
-      ! mass conservation across consecutive runs.
-      v_name = 'DELPDRY'
-      CALL HCO_GC_GetPtr( Input_Opt, State_Grid, TRIM(v_name), Ptr3D, RC, FOUND=FOUND )
-      IF ( FOUND ) THEN
-         State_Met%DELP_DRY = Ptr3D
-         IF ( Input_Opt%amIRoot ) THEN
-            WRITE(6,*) 'Initialize DELP_DRY from restart file'
-         ENDIF
-      ELSE
-         IF ( Input_Opt%amIRoot ) THEN
-            WRITE(6,*) 'DELP_DRY not found in restart, set to zero'
-         ENDIF
-      ENDIF
-      Ptr3D => NULL()
-
       ! Set dry surface pressure (PS1_DRY) from State_Met%PS1_WET
       ! and compute avg dry pressure near polar caps
       CALL Set_Dry_Surface_Pressure( State_Grid, State_Met, 1 )
--- a/GeosCore/hco_interface_gc_mod.F90
+++ b/GeosCore/hco_interface_gc_mod.F90
@@ -4991,7 +4991,8 @@ CONTAINS
              tmpFlx = 0.0_fp
              !$OMP PARALLEL DO            &
              !$OMP DEFAULT( SHARED       )&
-             !$OMP PRIVATE( I, J, tmpFlx )
+             !$OMP PRIVATE( I, J, tmpFlx )&
+             !$OMP COLLAPSE( 2 )
              DO J = 1, State_Grid%NY
              DO I = 1, State_Grid%NX
                 tmpFlx = dflx(I,J,N) / MW_kg * AVO * 1.e-4_fp                &
--- a/GeosCore/hco_interface_gc_mod.F90
+++ b/GeosCore/hco_interface_gc_mod.F90
@@ -4943,9 +4943,6 @@ CONTAINS
 
        ! Loop over only the drydep species
        ! If drydep is turned off, nDryDep=0 and the loop won't execute
-       !$OMP PARALLEL DO                                                     &
-       !$OMP DEFAULT( SHARED                                                )&
-       !$OMP PRIVATE( ND, N, ThisSpc, MW_kg, tmpFlx, S                      )
        DO ND = 1, State_Chm%nDryDep
 
           ! Get the species ID from the drydep ID
@@ -4992,22 +4989,24 @@ CONTAINS
           !-----------------------------------------------------------------
           IF ( Input_Opt%LSOILNOX ) THEN
              tmpFlx = 0.0_fp
+             !$OMP PARALLEL DO            &
+             !$OMP DEFAULT( SHARED       )&
+             !$OMP PRIVATE( I, J, tmpFlx )
              DO J = 1, State_Grid%NY
              DO I = 1, State_Grid%NX
-                tmpFlx = dflx(I,J,N)                                         &
-                       / MW_kg                                               &
-                       * AVO           * 1.e-4_fp                            &
+                tmpFlx = dflx(I,J,N) / MW_kg * AVO * 1.e-4_fp                &
                        * GET_TS_CONV() / GET_TS_EMIS()
-
                 CALL Soil_DryDep( I, J, 1, N, tmpFlx, State_Chm )
              ENDDO
              ENDDO
+             !$OMP END PARALLEL DO
           ENDIF
 
           ! Free species database pointer
           ThisSpc => NULL()
+
        ENDDO
-       !$OMP END PARALLEL DO
+
     ENDIF
 
     !=======================================================================
--- a/GeosCore/hco_interface_gc_mod.F90
+++ b/GeosCore/hco_interface_gc_mod.F90
@@ -4996,7 +4996,7 @@ CONTAINS
              DO I = 1, State_Grid%NX
                 tmpFlx = dflx(I,J,N) / MW_kg * AVO * 1.e-4_fp                &
                        * GET_TS_CONV() / GET_TS_EMIS()
-                CALL Soil_DryDep( I, J, 1, N, tmpFlx, State_Chm )
+                CALL Soil_DryDep( I, J, N, tmpFlx, State_Chm )
              ENDDO
              ENDDO
              !$OMP END PARALLEL DO
--- a/GeosCore/hco_utilities_gc_mod.F90
+++ b/GeosCore/hco_utilities_gc_mod.F90
@@ -1559,8 +1559,7 @@ CONTAINS
 ! !IROUTINE: get_gc_restart
 !
 ! !DESCRIPTION: Subroutine GET\_GC\_RESTART reads species concentrations
-!  [mol/mol] from the GEOS-Chem restart file and uses them to initialize
-!  species concentrations in [kg/kg dry]. If species data are missing from
+!  [mol/mol] from the GEOS-Chem restart file. If species data are missing from
 !  the restart file, pre-configured background values are used. If using the
 !  mercury simulation, additional restart data are read from file.
 !\\
@@ -1584,7 +1583,7 @@ CONTAINS
    USE State_Met_Mod,     ONLY : MetState
    USE Time_Mod,          ONLY : Expand_Date
    USE Timers_Mod,        ONLY : Timer_End, Timer_Start
-   USE UnitConv_Mod
+   USE UnitConv_Mod,      ONLY : MOLES_SPECIES_PER_MOLES_DRY_AIR
 #ifdef APM
    USE APM_Init_Mod,      ONLY : APMIDS
 #endif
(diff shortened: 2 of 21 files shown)


yantosca:
@lizziel, I think I know what this may be.  The N deposition code (I think `GET_N_DEP`) is trying to sum data into a 2-D array within a 3-D parallel loop.  This might lead to numerical differences if the loop over levels (L) is parallelized.   

You could get around that by declaring the variable to hold the level sum with `!$OMP REDUCTION(+: varname)`.   But it might just be more elegant to store the N deposition in a 3-D field of State_Met which could be summed in the vertical outside of a parallel loop.  I think the N deposition code pre-dated the State objects in GC v11-01.

GeosCore/hco_interface_gc_mod.F90
```diff
@@ -5008,22 +4989,24 @@ SUBROUTINE Compute_Sflx_for_Vdiff( Input_Opt,  State_Chm, State_Diag,      &
           !-----------------------------------------------------------------
           IF ( Input_Opt%LSOILNOX ) THEN
              tmpFlx = 0.0_fp
+             !$OMP PARALLEL DO            &
+             !$OMP DEFAULT( SHARED       )&
+             !$OMP PRIVATE( I, J, tmpFlx )
```
yantosca, line 4994: Add a `!$OMP COLLAPSE( 2 )` statement here, which will vectorize the loop.

GeosCore/hco_utilities_gc_mod.F90
```diff
@@ -1706,39 +1712,35 @@ SUBROUTINE Get_GC_Restart( Input_Opt, State_Chm, State_Grid, &
...
+      ! in the chemistry grid) or a small number (non-advected species
+      ! levels above the chemistry grid).
       IF ( SpcInfo%Is_InRestart ) THEN
 
-         ! Print the min & max of each species as it is read from
+         ! Print the min, max, and sum of each species as it is read from
          ! the restart file in mol/mol
          IF ( Input_Opt%amIRoot ) THEN
-            WRITE( 6, 120 ) N, TRIM( SpcInfo%Name ), &
+            WRITE( 6, 530 ) N, TRIM( SpcInfo%Name ), &
```
yantosca, line 1725: I think we can simply replace `SUM ( Ptr3D(:,:,1:State_Grid%NZ) )` with `SUM( Ptr3D )` since the `Temp3D` array is defined with NZ levels.

GeosCore/hco_utilities_gc_mod.F90
```diff
@@ -1706,39 +1712,35 @@ SUBROUTINE Get_GC_Restart( Input_Opt, State_Chm, State_Grid, &
...
 
-         ! Convert file value [mol/mol] to [kg/kg dry] for storage
          !$OMP PARALLEL DO       &
          !$OMP DEFAULT( SHARED ) &
          !$OMP PRIVATE( I, J, L )
          DO L = 1, State_Grid%NZ
          DO J = 1, State_Grid%NY
          DO I = 1, State_Grid%NX
-            Spc(N)%Conc(I,J,L) = Ptr3D(I,J,L) * MW_g / AIRMW
+            Spc(N)%Conc(I,J,L) = Ptr3D(I,J,L)
```
yantosca, line 1735: Can add a `!$OMP COLLAPSE( 3 )` statement to vectorize this loop.

GeosCore/hco_utilities_gc_mod.F90
```diff
@@ -1897,85 +1896,40 @@ SUBROUTINE Get_GC_Restart( Input_Opt, State_Chm, State_Grid, &
...
+   ELSE
+      State_Met%DELP_DRY = 0.0_fp
+      IF ( Input_Opt%amIRoot ) WRITE( 6, 520 ) ADJUSTL( v_name_in_file )
    ENDIF
 
+   ! Nullify pointer
+   Ptr3D => NULL()
+
    !=========================================================================
    ! Get variables for KPP mechanisms (right now just fullchem and Hg)
```
yantosca, line 1934: The comment should read `(right now fullchem, Hg, and carbon)'

Interfaces/GCClassic/main.F90
```diff
@@ -1706,6 +1724,35 @@ PROGRAM GEOS_Chem
...
+          ! Set species concentration back to v/v dry air (mol/mol dry air)
+          ! prior to setting restart file arrays. Doing this conversion every
+          ! timestep is required for bit-for-bit reproducibility when breaking up
+          ! runs in time.
+          !------------------------------------------------------------------------
+          ! Turn off diagnostics timer
+          IF ( Input_Opt%useTimers ) THEN
+             CALL Timer_End( "Diagnostics", RC )
+          ENDIF
+
```
yantosca, line 1737: @lizziel: Should we also start the unit conversion timer here?

Interfaces/GCClassic/main.F90
```diff
@@ -1706,6 +1724,35 @@ PROGRAM GEOS_Chem
...
+          ! Convert units
+          CALL Convert_Spc_Units(                                                  &
+               Input_Opt  = Input_Opt,                                             &
+               State_Chm  = State_Chm,                                             &
+               State_Grid = State_Grid,                                            &
+               State_Met  = State_Met,                                             &
+               new_units  = previous_units,                                        &
+               RC         = RC                                                    )
+
```
yantosca, line 1746: @lizziel: Should we end the unit conversion timer here?

Interfaces/GCClassic/main.F90
```diff
@@ -883,6 +886,20 @@ PROGRAM GEOS_Chem
           ENDIF
        ENDIF
 
+       !---------------------------------------------------------------------
```
yantosca, line 889: Should we start the unit conversion timer here?

Interfaces/GCClassic/main.F90
```diff
@@ -883,6 +886,20 @@ PROGRAM GEOS_Chem
...
+       CALL Convert_Spc_Units(                                                  &
+            Input_Opt      = Input_Opt,                                         &
+            State_Chm      = State_Chm,                                         &
+            State_Grid     = State_Grid,                                        &
+            State_Met      = State_Met,                                         &
+            new_units      = KG_SPECIES_PER_KG_DRY_AIR,                         &
+            previous_units = previous_units,                                    &
+            RC             = RC                                                )
+
```
yantosca, line 902: Should we end the unit conversion timer here?

yantosca (changes requested):
Thanks @lizziel, this looks mostly good.  I just had a few minor comments.

Interfaces/GCClassic/main.F90
lizziel: I don't think so since the unit conversion timer start and end are within Convert_Spc_Units.

GeosCore/hco_utilities_gc_mod.F90
lizziel: This is unrelated to the PR changes. Could that update go into docs/dev?

lizziel: Actually, looking at this code, are you sure it is supposed to include carbon? The section is to retrieve `KPP_HVALUE` from the restart file. If it should include carbon then there is a bug that would need to be fixed.

GeosCore/hco_utilities_gc_mod.F90
lizziel: Done!

GeosCore/hco_utilities_gc_mod.F90
lizziel: Done!

yantosca (approved):
Thanks @lizziel.  This is good to merge now.

yantosca:
After merging locally atop 14.6.0-alpha.8, all GCClassic integration tests passed:
```console

CodeDir   : 797f2ca GEOS-Chem update: Merge PR #2523 (Fix PDOWN def & reevap req)
GEOS-Chem : b1676fe42 Merge PR #2521 (Bit-for-bit reproducibility in GCClassic)
HEMCO     : 2423647 Merge PR #279 (Use USTAR from met in DustDEAD extension)
Cloud-J   : f8a2b7f Update version number for 8.0.1 release
HETP      : 2a99b24 Merge pull request #2 from geoschem/bugfix/initialize_local_variables

Using 24 OpenMP threads
Number of execution tests: 30

Submitted as SLURM job: 4282271

```
as well as all GCHP integration tests:
```console
GCHP: Execution Test Results

CodeDir       : 5da2a88 GEOS-Chem update: Merge PR #2523 (Fix PDOWN def & reevap req)
MAPL          : 6207fd0 Merge pull request #38 from geoschem/bugfix/intel_build_error_in_AbstractComponent
GMAO_Shared   : 4ddb3ec Merge pull request #2 from geoschem/feature/mapl-upgrade
ESMA_cmake    : ad5deba Added ecbuild as a submodule of ESMA_cmake
gFTL-shared   : 4b82492 Merge branch 'upstream_v1.5.0' into feature/v1.5.0
FMS           : 259759d Merge pull request #3 from geoschem/feature/update_gmao_libs
FVdycoreCubed : af42462 Merge PR #8 (Add PLEadv diagnostic for offline advection in GCHP)
geos-chem     : b1676fe42 Merge PR #2521 (Bit-for-bit reproducibility in GCClassic)
HEMCO         : 2423647 Merge PR #279 (Use USTAR from met in DustDEAD extension)
yaFyaml       : 19afe50 Merge branch 'upstream_v1.0.4' into feature/v1.0.4
pFlogger      : 2c4b724 Merge branch 'upstream_v1.9.1' into feature/v1.9.1
Cloud-J       : f8a2b7f Update version number for 8.0.1 release
HETP          : 2a99b24 Merge pull request #2 from geoschem/bugfix/initialize_local_variables

Number of execution tests: 12

Submitted as SLURM job: 4282450

```

Issue #2469, opened 2024-09-20
Input convective precipitation flux is zero in GEOS-FP (6/2020+) and GEOS-IT


lizziel:
### Your name

Lizzie Lundgren

### Your affiliation

Harvard University

### What happened? What did you expect to happen?

Viral Shah reports:

> There was an error in archiving the convective precipitation fields (PFLCU and PFICU) in our GEOS-IT and GEOS-FP data after the switch to Grell-Freitas (the fields are just set to zero). These fields are used in GEOS-Chem to calculate convective washout and re-evaporation

This bug impacts all GEOS-FP fields starting in Jun 2020 and the entire GEOS-IT inventory.

yuanjianz:
Glad to find this after I have just reported this to Randall (What a coincidence! :anguished: ). So it is confirmed to be a bug instead of a feature. It seems GEOS-IT has a much large convective rainfall re-evaporation than MERRA-2 here. Does it mean that we might be missing a lot of downward transport through convective precipitation and re-evaporation?
<img width="795" alt="image" src="https://github.com/user-attachments/assets/39753c2b-e748-4f8e-bcdb-bf8a682ffa34">

yuanjianz:
According to #1454, MERRA-2 precipitation flux reconstructed from DQR* - REEVAP* is lower than PFI* + PFL* by about 30-40%. In GEOS-IT, my tests shows that at least LSAN is lower when reconstructing from DQRLSAN and REEVAPLS. Maybe fixed GEOS-FP runs can give us more detailed hints, @viral211.

![output](https://github.com/user-attachments/assets/1f0f1e46-bb28-44b0-be82-c051ba44f4ed)
*from 1 year GEOS-IT transport tracer simulation for 2019

I also notice that PRECCON is preserved in GEOS-IT archive. I wonder if it is possible to utlize this information. For example, scale grid-wise reconstructed precipitation flux with PRECCON(surface flux)/reconstructed surface flux.

viral211:
In GF, DQRCU is the net precipitation source, and we can calculate the precipitation fluxes and the re-evaporated fraction from it alone - with just unit conversion. DQR is in units of kg(precip) / kg (air) / s, and precip fluxes in kg(precip) / m2 / s. REEVAPCN is the cumulative (over layers) re-evaporated flux, which is different from the previous definition, but we don’t need to use that field now anyway.
 
[...]

PR #2523, opened 2024-10-16
Fix convection issues when meteorology is Grell Freitas scheme

yuanjianz:

Yuanjian Zhang (WashU), Lizzie Lundgren (Harvard)

### Describe the update

See discussion https://github.com/geoschem/geos-chem/issues/1409

GEOS-IT and GEOS-FP after Jun 2020 switch to Grell Freitas convection from RAS. Some meteorology field definitions are changed. This update accommodates these fields to geos-chem convection.

Reported by @viral211 first:
>There was an error in archiving the convective precipitation fields (PFLCU and PFICU) in our GEOS-IT and GEOS-FP data after the switch to Grell-Freitas (the fields are just set to zero). These fields are used in GEOS-Chem to calculate convective washout and re-evaporation

@yuanjianz:
>Since DQRCU by definition is changed to net precipitation (MERRA-2 definition DQRCU minus layer-wise re-evaporation), do you think we should add layer-wise REEVAP back to match the DQRCU for determining cloud base height? It may lower the cloud base height somehow, given it seems some gas tracer is substantially below observation constraint?

Changes include:
1. Add surface precipitation flux fields as inputs to GCHP.
2. Add meteorology source name to GCHP run directory `geoschem_config.yml`
3. Create new `Input_Opt` logical for whether to reconstruct convective precipitation flux. This switch is set automatically during run-time to true for GEOS-IT, true for GEOS-FP if start date is June 1 2020 or later, and false for all other cases.
4. Use the new `Input_Opt` logical to use an alternative method in convection which bypasses use of PFLCU and PFICU when necessary. This fixes the zero convective precipitation and high cloud base in runs using GEOS-FP (>=01Jun2020) or GEOS-IT.
5. Add a warning during run directory creation about the convection scheme change and the precipitation flux bug (only if selecting GEOS-FP meteorology). This warning stops run directory creation and requires acknowledgement from the user to continue.

Related convection fixes are also applicable to simulations using met-fields generated with RAS. These are submitted via separate PR https://github.com/geoschem/geos-chem/pull/2573. Those updates will impact the GEOS-Chem benchmark.

### Expected changes

Lower cloud base and restored washout in geos-chem convection driven by GEOS-IT and GEOS-FP after Jun 2020. This will not affect the GEOS-Chem benchmark which uses MERRA2.

### Related Github Issue

https://github.com/geoschem/geos-chem/issues/2469


GeosCore/convection_mod.F90
```diff
@@ -616,24 +652,29 @@ SUBROUTINE DO_CLOUD_CONVECTION( Input_Opt,  &
...
+    ! PFLCU and PFICU are converted from kg/m2/s to m3/m2/s
+    ! using water and ice densities, respectively.
+    ! m3/m2/s becomes cm3/cm2/s using a factor of 100.
+    IF ( Is_GF_Conv == .FALSE. ) THEN
+       PDOWN(:) = ( ( PFLCU(:) / 1000e+0_fp ) &
+                +   ( PFICU(:) /  917e+0_fp ) ) * 100e+0_fp
+    ELSE
+       PDOWN(NLAY) = 0
+       DO K = NLAY-1, 1, -1
+          PDOWN(K) = PDOWN(K+1) + DQRCU_MET(K+1) &
```
viral211, line 667: Shouldn't this be: PDOWN(K) = PDOWN(K+1) + DQRCU_MET(**K**) * DELP(**K**) * G0_100 * 100e+0_fp

yuanjianz: I did this to align with the pre-existing configuration:
https://github.com/geoschem/geos-chem/blob/bef56c605e018eecbd91646a51ce82c7cd77f56a/GeosCore/convection_mod.F90#L569-L575
But if we look at where PDOWN is used, it seems to imply that PDOWN(K) to be the flux at the bottom of level K. So I think you are right, and we need to fix both RAS and GF's PDOWN. 

Also, I have a question about:
https://github.com/geoschem/geos-chem/blob/bef56c605e018eecbd91646a51ce82c7cd77f56a/GeosCore/convection_mod.F90#L1101-L1107
Why is essential to have re-evaporation to have wash-out? Couldn't it be washout event without re-evaporation?

viral211: You are right. Re-evaporation should not be a condition for washout, but I guess there will always be re-evaporation when the precip falls through drier air below the cloud base. Could you remove the re-evaporation condition and see how much difference it makes? 
Let's fix PDOWN in RAS too. Thanks!

yuanjianz: Hi Viral, please check the latest commit.

viral211: Hi @yuanjianz, this looks great! One thing I could not tell for certain is whether we go through the entire routine even if there is no convection taking place in a column at the time step, or if there is an early exit somewhere if, say, the convective mass flux is zero throughout the column.

lizziel:
I updated the PR description to include my contributions submitted via https://github.com/yuanjianz/geos-chem/pull/1.

run/GCClassic/createRunDir.sh
```diff
@@ -353,6 +353,26 @@ while [ "${valid_met}" -eq 0 ]; do
...
+
+	# Print warning about GEOS-FP and require user to acknowledge it.
+	fp_msg="WARNING: The convection scheme used to generate archived GEOS-FP meteorology \nfiles changed from RAS to Grell-Freidas starting June 1 2020 with impact on \nvertical transport. Discussion and analysis of the impact is available at \ngithub.com/geoschem/geos-chem/issues/1409. In addition, there is a bug in \nconvective precipitation flux following the switch where all values are zero \nin the input files. This bug is addressed by computing fluxes online for runs \nstarting on or after June 1 2020. The fix does not extend to the case of running \nacross the time boundary. Due to these issues we recommend splitting up GEOS-FP \nruns in time such that a single simulation does not span the switch. Configure \none run to end on June 1 2020 and then use its output restart to start another \nrun on June 1. Alternatively consider using MERRA2. If you wish to use a \nGEOS-FP meteorology year different from your simulation year please create a \nGEOS-Chem GitHub issue for assistance to avoid accidentally using zero \nconvective precipitation flux.\n"
```
yantosca, line 358: Fix typo: "Grell-Freidas" -> "Grell-Freitas"

run/GCHP/createRunDir.sh
```diff
@@ -317,6 +317,25 @@ while [ "${valid_met}" -eq 0 ]; do
 
        	met="geosfp"
 
+	# Print warning about GEOS-FP and require user to acknowledge it.
+	fp_msg="WARNING: The convection scheme used to generate archived GEOS-FP meteorology \nfiles changed from RAS to Grell-Freidas starting June 1 2020 with impact on \nvertical transport. Discussion and analysis of the impact is available at \ngithub.com/geoschem/geos-chem/issues/1409. In addition, there is a bug in \nconvective precipitation flux following the switch where all values are zero \nin the input files. This bug is addressed by computing fluxes online for runs \nstarting on or after June 1 2020. The fix does not extend to the case of running \nacross the time boundary. Due to these issues we recommend splitting up GEOS-FP \nruns in time such that a single simulation does not span the switch. Configure \none run to end on June 1 2020 and then use its output restart to start another \nrun on June 1. Alternatively consider using MERRA2. If you wish to use a \nGEOS-FP meteorology year different from your simulation year please create a \nGEOS-Chem GitHub issue for assistance to avoid accidentally using zero \nconvective precipitation flux.\n"
```
yantosca, line 321: Fix typo: "Grell-Freidas" -> "Grell-Freitas"

yantosca (changes requested):
Hi @lizziel, thanks for the PR.  I just caught a couple of typos in the warning messages in the rundir creation scripts.  Otherwise looks good to go!

yantosca (approved):
Thanks @lizziel and @yuanjianz!  Good to merge.

lizziel:
Great, I'll bring this in along with the GCHPctmEnv fixes PR.

PR #2608, opened 2024-12-03
Fix tagged CO simulation in dev branch

lizziel:

Name: Lizzie Lundgren

### Describe the update

This PR fixes a bug introduced to the dev/no-diff-to-benchmark branch when 14.5.0 was merged in. That merge reverted an
update to change the start date and restart file for tagged CO simulations. With this updates the tagged CO integration tests will no longer fail.

Note that there is no need for a changelog update because the issue is only in the dev/no-diff-to-benchmark branch and is not in any official releases.

### Expected changes

Tagged CO simulations will use the carbon restart file and start on Jan 1 by default. This returns functionality introduced in the CO2 updates in the dev/no-diff-to-benchmark branch.

### Reference(s)

None

### Related Github Issue

None needed since bug is not in any releases.

### DIFF
--- a/run/shared/download_data.yml
+++ b/run/shared/download_data.yml
@@ -79,8 +79,8 @@ restarts:
     remote: v2020-02/GEOSChem.Restart.POPs_BaP.20190701_0000z.nc4
     local:  GEOSChem.Restart.20190701_0000z.nc4
   tagco:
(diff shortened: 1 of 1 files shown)


yantosca (changes requested):
Looks OK but we just need a changelog update

yantosca (approved):
Now approved.  I didn't see the comment about the changelog at first.  D'oh!

PR #2617, opened 2024-12-06
KPP-Standalone submodule update: Now generate box-model output with 3-digit exponents

yantosca:

Name: Bob Yantosca
Institution: Harvard + GCST

### Describe the update

This PR updates the KPP-Standalone submodule commit hash to commit https://github.com/geoschem/KPP-Standalone/commit/eeee8951f03cc538a3f7a20d4a2a98470db5a021.

This fixes an error in the KPP-Standalone submodule where very small concentrations (1e-100 or less) would be written to the log with garbled output.  We now use 3 digits in the exponent to avoid this from happening.

### Expected changes
This is a zero-diff update.

### Reference(s)
N/A

### Related Github Issue
- See PR #2588


yantosca:
Integration tests are running.

msulprizio (approved):

PR #2675, opened 2025-01-09
Fixed the problem that GCHP cannot read meteorology field from yaml

yuanjianz:

Name: Yuanjian Zhang
Institution: WashU

### Describe the update

Fixed the problem that GCHP cannot read meteorology field from yaml

### DIFF
--- a/GeosCore/input_mod.F90
+++ b/GeosCore/input_mod.F90
@@ -609,10 +609,28 @@ CONTAINS
     ENDIF
     Input_Opt%CHEM_INPUTS_DIR = TRIM( v_str )
 
+#if defined( MODEL_GCHP )
+    !------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


yantosca (changes requested):
Looks good @yuanjianz, we would just need a CHANGELOG.md update.

lizziel:
I believe this fixes an issue brought in with the updates for GEOS-IT in dev/no-diff-to-benchmark (see https://github.com/geoschem/geos-chem/pull/2523). It would only show up for GEOS-FP and GEOS-IT runs which we do not test in integration/AWS testing. It therefore does not need an update to the changelog.

yantosca (approved):
Now good to merge.  Please disregard my previous comment about the changelog.

Issue #2738, opened 2025-02-19
Bugs in fixed level and PBL budget diagnostics


lizziel:
### Your name

Lizzie Lundgren

### Your affiliation

Harvard University

### What happened? What did you expect to happen?

While developing GEOS-Chem 14.6.0 I noticed an issue with the column mass array used to compute budget diagnostics for fixed level and PBL. The array's 4th dimension is incorrectly allocated with size 3 and both the PBL and fixed level diagnostics use index 3 of the 4th dimension to track column tracer mass. The 4th dimension should be allocated with size 4 and the fixed level budget diagnostic should index 4.

### What are the steps to reproduce the bug?

Run a GEOS-Chem fullchem benchmark with budget diagnostics on.

### What GEOS-Chem version were you using?

14.5

### What environment were you running GEOS-Chem on?

Local cluster

### What compiler and version were you using?

GCC 12

### Will you be addressing this bug yourself?

Yes

### In what configuration were you running GEOS-Chem?

GCClassic

### What simulation were you running?

Full chemistry

### As what resolution were you running GEOS-Chem?

4x5

### What meterology fields did you use?

MERRA-2

yantosca:
We can now close this issue, as PR #2739 has been merged into the GEOS-Chem 14.6.0 development stream.

PR #2739, opened 2025-02-19
Fix bugs in fixed level and PBL budget diagnostics

lizziel:

Name: Lizzie Lundgren

### Describe the update

This PR fixes an issue with the column mass array used to compute budget diagnostics for fixed level and PBL. The array's 4th dimension was incorrectly allocated with size 3 and both the PBL and fixed level diagnostics used index 3 of the 4th dimension to track column tracer mass. The 4th dimension is now allocated with size 4. The PBL budget diagnostics continue to use index 3 but the fixed level budget diagnostic now uses index 4.

### Expected changes

This update only affected budget diagnostics. PBL and fixed level budget diagnostics will now be correct. Full and Trop budget diagnostics are not affected.

### Reference(s)

None

### Related Github Issue

close https://github.com/geoschem/geos-chem/issues/2738

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Fixed
 - Fixed PDOWN definition to lower rather than upper edge
+- Fixed bugs in column mass array affecting budget diagnostics for fixed level and PBL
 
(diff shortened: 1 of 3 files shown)


yantosca (approved):
Thanks @lizziel!  This is a simple fix to bump up the 4th dimension of the colmass array and is OK to merge.

PR #2763, opened 2025-03-10
Fix dry-run simulation issues in 14.6.0 development version

yantosca:

Name: Bob Yantosca
Institution: Harvard + GCST

### Describe the update
This PR fixes a few issues with GEOS-Chem Classic dry-run simulations that were discovered in the 14.6.0-alpha.10 tag.

- Placed the first call to `Convert_Spc_Units` in the `IF ( notDryRun )` block that precedes it.  This will make sure that the unit conversions are not done during dry-run simulations (which will cause an out-of-bounds error).  This came in with the PR #2521.
- Added `--no-sign-request` to `aws s3 cp` commands in `run/shared/download_data.yml`.  This will allow for S3 data download via anonymous login (for users without AWS accounts).
- Added cosmetic changes (comments, updated column alignments, etc).

### Expected changes
The updates in this PR will allow dry-run simulations to proceed to completion (avoiding an out-of-bounds error).

### Related Github Issue

- See #2521


lizziel (approved):
Looks good. Maybe we should add a GC-Classic dry-run simulation to integration testing to catch issues with dry-run in new features.

yantosca:
@lizziel: Agreed, we should definitely have a dry-run integration test.  Maybe even one for each general simulation type.

PR #252, opened 2020-03-19
Reimplement a regressed feature for WRF-GC in DRYDEP_MOD.

jimmielin:
For some reason, a duplicate allocation check in DRYDEP_MOD implemented for WRF-GC was commented out by @ganluoasrc. I have reimplemented it under `MODEL_WRF`.

Signed-off-by: Haipeng Lin <hplin@seas.harvard.edu>

PR #2806, opened 2025-04-10
Fix bugs in GCHP species diagnostics and metric.py

lizziel:

Name: Lizzie Lundgren

### Describe the update

This PR fixes a bug introduced earlier in 14.6.0 development that omitted a new call to set species diagnostics in GCHP. It also update the metrics.py script in run directories to fix the incorrect start and end dates printed for GCHP. Simulation duration and end are printed instead.

### Expected changes

This update will restore GCHP species diagnostics that have been zero since 14.6.0-alpha.9. When these updates are applied to 14.6.0-alpha.9 then GCHP is zero-diff with respect to 14.6.0-alpha.8.

The metrics.py will have different dates shown in the printed header for GCHP.

### Reference(s)

None

### Related Github Issue

https://github.com/geoschem/geos-chem/pull/252


yantosca (approved):
Thanks @lizziel.  Good to merge. I'll start bringing this in.

PR #2870, opened 2025-05-23
Adjoint updates

kaysuselj:
### Name and Institution (Required)

Name: Kay Suselj
Institution: JPL

### Describe the update

Modifications to adjoint code.

### Expected changes

Changes only affect adjoint calculation.

### Related Github Issue
Should be merged concurrently with
- https://github.com/geoschem/hemco/pull/323
- https://github.com/geoschem/GCHP/pull/489

PR #2888, opened 2025-06-10
Adjoint mixing

kaysuselj:

Name:
Institution:

### Describe the update

Please provide a clear and concise overview of the update.

### Expected changes

Please provide details on how this update will impact model output and include plots or tables as needed.

### Reference(s)

If this is a science update, please provide a literature citation.

### Related Github Issue

Please link to the corresponding Github issue(s) here. If fixing a bug, there should be an issue describing it with steps to reproduce.

### DIFF
--- a/Interfaces/GCHP/Chem_GridCompMod.F90
+++ b/Interfaces/GCHP/Chem_GridCompMod.F90
@@ -1636,8 +1636,7 @@ CONTAINS
        DO I = 1, nFlds
 
           ! Get info about this species from the species database
-          N = State_Chm%Map_Advect(I)
-          ThisSpc => State_Chm%SpcData(N)%Info
+          ThisSpc => State_Chm%SpcData(I)%Info
 
           ! Pass tracer name
           Int2Adj(I)%Name = TRIM(ThisSpc%Name)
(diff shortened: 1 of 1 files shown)


Interfaces/GCHP/Chem_GridCompMod.F90
```diff
@@ -3464,13 +3461,14 @@ SUBROUTINE Finalize_( GC, Import, Export, Clock, RC )
...
+
        CFN = 0d0
        DO L = 1, State_Grid%NZ
        DO J = 1, State_Grid%NY
        DO I = 1, State_Grid%NX
           if (State_Chm%CostFuncMask(I,J,L) > 0d0) THEN
-             WRITE (*, 1047) I, J, L, State_Chm%Species(NFD)%conc(I,J,L)
-             CFN = CFN + State_Chm%Species(NFD)%Conc(I,J,L)
+!             WRITE (*, 1047) I, J, L, State_Chm%Species(NFD)%conc(I,J,L)
+!             CFN = CFN + State_Chm%Species(NFD)%Conc(I,J,L)
```
kaysuselj, line 3471: My code crashes when trying to print out. I don't really see that something would be wrong with these two lines, but I had a few other occurrences of crashes when writing output to the screen.

yantosca:
Thanks @kaysuselj for this.  Does this PR replace #2870?

Interfaces/GCHP/Chem_GridCompMod.F90
yantosca: @kaysuselj it could be a formatting error or an out-of-bounds error.

kaysuselj:
Yes basically this is an update to https://github.com/geoschem/geos-chem/pull/2870, with two additional changes:
1. Commented the lines where the code crashes while writing output -- I am not sure this is a problem with my setup only, or it is problem with the code. I have similar issue at other places in the code.
2. One more change in Chem_GridCompMod (line ~1639). Could you please see if that makes sense.

kaysuselj:
@lizziel : Thanks, I will try compiling with the debug on and see if I get any useful information.

PR #2893, opened 2025-06-11
Fix for emission configurations for carbon simulations

1Dandan:

Name: Dandan Zhang

### Describe the update

I notices several time range or unit error in the `HEMCO_Config.rc.CH4` and `HEMCO_Config.rc.carbon` by checking the corresponding ExtData path. 

The changes of IMI-specfic config requires the changes for HEMCO at https://github.com/geoschem/HEMCO/pull/330

- For the time range, for example, `JPL_WETCHARTS` at the path of `~/ExtData/HEMCO/CH4/v2024-01/JPL_WetCharts/HEensemble/JPL_WetCharts_2010-2019.Ensemble_Mean.0.5x0.5.nc` only has time range of 2010 although it is from the ensemble of 2009-2017, while the time range of 2009-2017 was supplied. Similar justifications apply to other modifications.
- The units in `HEMCO_Config.rc` are sometimes not consistent with the units of netCDF file in the corresponding path as well, and thus these errors are also fixed.
- `ExtData.rc.carbon` was modified accordingly to be consistent with `HEMCO_Config.rc.carbon` to read correct time range and unit.
- The problem with daily GFED emissions is that they are calculated from monthly GFED emission dataset multiplied by the daily scaling factor dataset, which would be first regridded and then multiplied and thus would not ensure correct and consistent emission fluxes. Detailed explanation can be found at https://github.com/geoschem/HEMCO/issues/181.
- The entry of `OH_PosteriorSF` and corresponding scaling factor of `OH_SF` were never used. Instead, hemispheric scaling factor constant for OH was used. Thus, `OH_PosteriorSF` and `OH_SF` were deleted
- [Update scaling factor ID at 3000 to avoid conflicts with CEDS_01x01 scaling factor](https://github.com/geoschem/geos-chem/pull/2893/commits/393f5267def4cb91e7feadfc117b144c83fc3ec7) enabled in carbon simulation.

Specific changes for IMI would only affect IMI runs with the options of `AnalyticalInversion`, `UseTotalPriorEmis`, and `Emis_PosteriorSF`. Thus, these updates, especially the fixes for time range and unit should be added for both methane-only (GCC) and carbon simulations (GCC and GCHP).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Added operational run scripts for WashU Compute2
 - Added the option for LPJ_MERRA2 wetland CH4 emissions in CH4 and carbon simulations
 - Added GC-Classic config file option to read restart file as `REAL*8` via GEOS-Chem rather than HEMCO
+- Added chemistry budget diagnostics to GCHP carbon HISTORY.rc
 
 ### Changed
 - Replaced comments in template HEMCO configuration files directing users to obsolete wiki documentation with comments directing users to `hemco.readthedocs.io`
@@ -33,6 +34,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Updated GFEIv3 files to correct issue in original version
 - Updated `download_data.py` for compatibility with 0.125 x 0.15625 grids plus all pre-defined nested-grids
 - Restructured `download_data.py` to avoid several instances of repeated code
+- Commented out met-fields PEDGEDRY, PFICU, PFILSAN, PFLCU, and PFLLSAN by default in GC-Classic and GCHP carbon HISTORY.rc, and GC-Classic CH4 HISTORY.rc
 
 ### Fixed
 - Restored entries for TMB emissions in `HEMCO_Config.rc.fullchem` template files for GCClassic and GCHP
@@ -45,13 +47,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Fixed error where `//` were not being changed to `/` in `download_data.py`
 - Change precision of area import from GCHP advection from `REAL*4` to native `REAL*8`
 - Fixed time-range and units for CH4 emission inventories to be consistent with the corresponding netCDF files in ExtData directory for `HEMCO_Config.rc` and `ExtData.rc`
-- Switched to daily GFED4 instead of monthly GFED4 to avoid inconsistent fire emissions across resolutions
 - Updated scaling factor ID at 3000 to avoid conflicts with CEDS_01x01 scaling factor enabled in carbon simulation for IMI analytical inversion
 
 ### Removed
 - Removed entries for FINN v1.5 biomass burning emissions from template HEMCO configuration files
 - Removed `Is_Advected` tags from `run/shared/species_database*.yml` template files
-- Removed `OH_PosteriorSF` entry in  is never used in IMI and thus deleted
+- Removed `OH_PosteriorSF` entry in carbon and CH4 HEMCO_Config.rc since never used
 
 ## [14.6.3] - 2025-07-28
 ### Added


msulprizio:
@1Dandan Could you please modify the description of this PR when ready to include more detail, including explaining why you made each change?

1Dandan:
Yes. Things are still evolving. I will make it more comprehensive when it is ready for review.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.carbon
```diff
@@ -525,8 +524,8 @@ Mask fractions:              false
 # Use updated files (v2024-01); these are COARDS-compliant.
 #==============================================================================
 (((RESERVOIRS
-0 CH4_RES_DAM $ROOT/CH4/v2024-01/ResME/ResME_Dam_Emissions.0.1x0.1.nc     CH4emis 2022/1-12/1/0 C xy kg/m2/s CH4 1500 15 1
-0 CH4_RES_SFC $ROOT/CH4/v2024-01/ResME/ResME_Surface_Emissions.0.1x0.1.nc CH4emis 2022/1-12/1/0 C xy kg/m2/s CH4 1500 15 1
+0 CH4_RES_DAM $ROOT/CH4/v2024-01/ResME/ResME_Dam_Emissions.0.1x0.1.nc     CH4emis 2022/1/1/0 C xy kg/m2/s CH4 1500 15 1
+0 CH4_RES_SFC $ROOT/CH4/v2024-01/ResME/ResME_Surface_Emissions.0.1x0.1.nc CH4emis 2022/1/1/0 C xy kg/m2/s CH4 1500 15 1
```
yantosca: @1Dandan @msulprizio: Shouldn't we be using the monthly reservoir emissions here?

run/GCHP/HEMCO_Config.rc.templates/HEMCO_Config.rc.carbon
```diff
@@ -525,8 +524,8 @@ Mask fractions:              false
 # Use updated files (v2024-01); these are COARDS-compliant.
 #==============================================================================
 (((RESERVOIRS
-0 CH4_RES_DAM $ROOT/CH4/v2024-01/ResME/ResME_Dam_Emissions.0.1x0.1.nc     CH4emis 2022/1-12/1/0 C xy kg/m2/s CH4 1500 15 1
-0 CH4_RES_SFC $ROOT/CH4/v2024-01/ResME/ResME_Surface_Emissions.0.1x0.1.nc CH4emis 2022/1-12/1/0 C xy kg/m2/s CH4 1500 15 1
+0 CH4_RES_DAM $ROOT/CH4/v2024-01/ResME/ResME_Dam_Emissions.0.1x0.1.nc     CH4emis 2022/1/1/0 C xy kg/m2/s CH4 1500 15 1
+0 CH4_RES_SFC $ROOT/CH4/v2024-01/ResME/ResME_Surface_Emissions.0.1x0.1.nc CH4emis 2022/1/1/0 C xy kg/m2/s CH4 1500 15 1
```
yantosca: @1Dandan @msulprizio @lizziel: Also wondering if we should be using the monthly emissions here as well.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.carbon
1Dandan: There is only one time dimension for these files, and thus I changed it accordingly.

1Dandan:
Added changes for CHANGELOG.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.CH4
```diff
@@ -464,7 +463,7 @@ VerboseOnCores:              root       # Accepted values: root all
 # Use updated files (v2024-01); these are COARDS-compliant.
 #==============================================================================
 (((JPL_WETCHARTS
-0 JPLW_CH4  $ROOT/CH4/v2024-01/JPL_WetCharts/HEensemble/JPL_WetCharts_2010-2019.Ensemble_Mean.0.5x0.5.nc emi_ch4 2010-2019/1-12/1/0 C xy molec/cm2/s CH4 - 10 1
+0 JPLW_CH4  $ROOT/CH4/v2024-01/JPL_WetCharts/HEensemble/JPL_WetCharts_2010-2019.Ensemble_Mean.0.5x0.5.nc emi_ch4 2010/1-12/1/0 C xy molec/cm2/s CH4 - 10 1
```
msulprizio, line 468: The year range here should be 2010-2019. The JPL WetCharts file is a model ensemble mean for 2010-2019. If you check the netCDF file, it does indeed have monthly values for 10 years (i.e. time dimension of 120). I think the actual issue here and with GCHP is the file itself. The time dimension is specified incorrectly. When I do `ncdump -v time -t WetCHARTs_Highest_performance_Ensemble_v1.3.1_2010_2019.nc`, it returns:

```
netcdf WetCHARTs_Highest_performance_Ensemble_v1.3.1_2010_2019 {
dimensions:
        lon = 720 ;
        lat = 360 ;
        time = 120 ;
        model = 9 ;
variables:
       double time(time) ;
                time:units = "months since Jan 1st 2001 00:00:00" ;
                
...

 time = "0000-02-01", "0000-03-01", "0000-04-01", "0000-05-01", "0000-06-01", 
    "0000-07-01", "0000-08-01", "0000-09-01", "0000-10-01", "0000-11-01", 
    "0000-12-01", "0001-01-01", "0001-02-01", "0001-03-01", "0001-04-01", 
    "0001-05-01", "0001-06-01", "0001-07-01", "0001-08-01", "0001-09-01", 
    "0001-10-01", "0001-11-01", "0001-12-01", "0002-01-01", "0002-02-01", 
    "0002-03-01", "0002-04-01", "0002-05-01", "0002-06-01", "0002-07-01", 
    "0002-08-01", "0002-09-01", "0002-10-01", "0002-11-01", "0002-12-01", 
    "0003-01-01", "0003-02-01", "0003-03-01", "0003-04-01", "0003-05-01", 
    "0003-06-01", "0003-07-01", "0003-08-01", "0003-09-01", "0003-10-01", 
    "0003-11-01", "0003-12-01", "0004-01-01", "0004-02-01", "0004-03-01", 
    "0004-04-01", "0004-05-01", "0004-06-01", "0004-07-01", "0004-08-01", 
    "0004-09-01", "0004-10-01", "0004-11-01", "0004-12-01", "0005-01-01", ... etc...
```

and an error 

```
CDMS error: Error on character time conversion, string = Jan 1st
```

So I think the issue is with the attributes/values of the time dimension in the file itself, not with the HEMCO configuration file.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.CH4
```diff
@@ -93,11 +93,10 @@ VerboseOnCores:              root       # Accepted values: root all
     --> AnalyticalInversion    :       false
     --> UseTotalPriorEmis      :       false    # Skips global/regional inventories
     --> Emis_PosteriorSF       :       false    # Apply posterior scale factors to total emis?
-    --> OH_PosteriorSF         :       false    # Apply posterior scale factor  to global OH?
 # -----------------------------------------------------------------------------
 111     GFED                   : on    CH4
     --> GFED4                  :       true
-    --> GFED_daily             :       true
+    --> GFED_daily             :       false
```
msulprizio: This should remain set to true, especially for GCClassic simulations. This is the recommended setting for CH4 emissions and has been used in CH4/IMI publications. 

I understand the issue as reported in https://github.com/geoschem/HEMCO/issues/181 but that applies to GCHP only. You could set it to false in the GCHP HEMCO_Config.rc file only, but that would lead to differences between GCClassic and GCHP carbon simulations. So I would recommend either setting it to false in your personal GCHP runs directories or accepting the differences that come with changes in resolution.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.carbon
msulprizio: While it is correct that the ResME_*_Emissions.0.1x0.1.nc files have one time dimension, we apply monthly scale factors to them via:

```
(((RESERVOIRS
1500 SEASONAL_RES_MASK $ROOT/CH4/v2024-01/ResME/reservoir_mask.01x01.nc Mask 2022/1-12/1/0 C xy 1 1 -180/-90/180/90
)))RESERVOIRS
```

Therefore to tell HEMCO to update this field monthly, we need to keep the monthly time stamp as 1-12.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.CH4
```diff
@@ -878,18 +863,15 @@ ${RUNDIR_GLOBAL_Cl}
...
 
-(((OH_PosteriorSF
-4 OH_SF      Post_SF_OH.nc         SF_OH       2010-2017/1/1/0    C xy 1 1
-)))OH_PosteriorSF
-
 #==============================================================================
 # --- Scale factors for analytical inversions ---
 #==============================================================================
 (((AnalyticalInversion
+4 SCALE_PERT      ./gridded_pert_scale_1.nc          scale    2000/1/1/0 C xy 1 1
```
msulprizio, line 881: Is this a custom file for your runs? If so, please remove.

run/GCHP/ExtData.rc.templates/ExtData.rc.carbon
```diff
@@ -220,8 +223,8 @@ EDGAR8_CH4_CHE                 kg/m2/s N Y F%y4-%m2-01T00:00:00  none none emi_c
 EDGAR8_CH4_IRO                 kg/m2/s N Y F%y4-%m2-01T00:00:00  none none emi_ch4 ./HcoDir/CH4/v2024-02/EDGARv8/%y4/v8.0_FT2022_GHG_CH4_%y4_IRO_flx.nc             
 
 # --- EDGAR v7 livestock ---
-EDGAR7_CH4_LIVESTOCK__4A      kg/m2/s N Y F%y4-%m2-01T00:00:00  none none  emi_ch4  ./HcoDir/CH4/v2023-04/EDGARv7/%y4/v7.0_FT2021_CH4_%y4_ENF.0.1x0.1.nc
-EDGAR7_CH4_LIVESTOCK__4B      kg/m2/s N Y F%y4-%m2-01T00:00:00  none none  emi_ch4  ./HcoDir/CH4/v2023-04/EDGARv7/%y4/v7.0_FT2021_CH4_%y4_MNM.0.1x0.1.nc
+EDGAR7_CH4_LIVESTOCK__4A      kg/m2/s N Y F%y4-01-01T00:00:00  none none  emi_ch4  ./HcoDir/CH4/v2023-04/EDGARv7/%y4/v7.0_FT2021_CH4_%y4_ENF.0.1x0.1.nc
+EDGAR7_CH4_LIVESTOCK__4B      kg/m2/s N Y F%y4-01-01T00:00:00  none none  emi_ch4  ./HcoDir/CH4/v2023-04/EDGARv7/%y4/v7.0_FT2021_CH4_%y4_MNM.0.1x0.1.nc
```
msulprizio, line 227: This change is fine. I think we originally had `%m2` here because of the 1-12 timestamp used in HEMCO_Config.rc, but the EDGARv7 livestock files do have one timestamp. We need the monthly timestamp in HEMCO_Config.rc because we apply monthly scale factors (`EDGAR_SEASONAL_SF_ENF` and `EDGAR_SEASONAL_SF_MNM`) to the livestock emissions. The  monthly livestock scale factors are specified further down in ExtData.rc.carbon for GCHP.

run/GCHP/ExtData.rc.templates/ExtData.rc.carbon
```diff
@@ -243,7 +246,7 @@ EDGAR8_CH4_SWD_INC             kg/m2/s N Y F%y4-%m2-01T00:00:00  none none emi_c
 
 # --- JPL WetCHARTs v1.0 (Bloom et al., https://doi.org/10.3334/ORNLDAAC/1502) ---
 # Include scaling to convert molecules/cm2/s to kg/m2/s
-JPLW_CH4  kg/m2/s N Y F%y4-%m2-01T00:00:00 none 2.66350462E-22 emi_ch4  ./HcoDir/CH4/v2024-01/JPL_WetCharts/HEensemble/JPL_WetCharts_2010-2019.Ensemble_Mean.0.5x0.5.nc
+JPLW_CH4  kg/m2/s Y Y F2010-%m2-01T00:00:00 none 2.66350462E-22 emi_ch4  ./HcoDir/CH4/v2024-01/JPL_WetCharts/HEensemble/JPL_WetCharts_2010-2019.Ensemble_Mean.0.5x0.5.nc
```
msulprizio, line 249: This should be restored following the comment above for JPL WetCharts. The file does indeed include data for 2010-2019 and we don't want it hardcoded to a single year.

run/GCHP/ExtData.rc.templates/ExtData.rc.carbon
```diff
@@ -425,6 +428,29 @@ CO2_WEEKLY  1 D Y F2006-01-%d2T00:00:00 none none weekly_scale_factors  ./HcoDir
 # --- Domestic aviation surface correction factor ---
 AVIATION_SURF_CORR      1       Y  Y F2004-01-01T00:00:00  none none CO2 ./HcoDir/CO2/v2022-11/FOSSIL/Aviation_SurfCorr_SclFac.1x1.nc
 
+#========================================================
+# --- Scale factors for OH ---
+#========================================================
+#OH_pert_factor 1 N Y - none none oh_scale ./gridded_perturbation_oh_scale.nc
```
msulprizio, line 437: Remove these lines? We don't read a netCDF file for perturbing OH scale factors.

run/GCHP/ExtData.rc.templates/ExtData.rc.carbon
```diff
@@ -425,6 +428,29 @@ CO2_WEEKLY  1 D Y F2006-01-%d2T00:00:00 none none weekly_scale_factors  ./HcoDir
...
+# Entries below are provided for examples only. Add your own here!
+#==============================================================================
+EMIS_SF 1 N Y -   none none ScaleFactor gridded_posterior.nc
+
+#==============================================================================
+# --- Scale factors for analytical inversions ---
+#==============================================================================
+# Add perturbations to individual state vector element (N) following this format
+SCALE_PERT       1 N Y 2000-01-01T00:00:00 none none scale       ./gridded_pert_scale_1.nc
+#SCALE_ELEM_000N  1 N Y 2000-01-01T00:00:00 none none StateVector ./StateVector.nc
```
msulprizio, line 449: Remove these lines also. It's not clear what they're needed for.

run/GCHP/HEMCO_Config.rc.templates/HEMCO_Config.rc.carbon
```diff
@@ -164,11 +163,11 @@ Mask fractions:              false
 (((UseTotalPriorEmis
 
 (((Emis_PosteriorSF
-0 CH4_Emis_Prior ../../prior_run/OutputDir/HEMCO_sa_diagnostics.$YYYY$MM$DD0000.nc EmisCH4_Total $YYYY/$MM/$DD/0 C xy kg/m2/s CH4 3 1 500
+0 CH4_Emis_Prior ../../prior_run/OutputDir/GEOSChem.Emissions.$YYYY$MM$DD_0000z.nc4 EmisCH4_Total $YYYY/$MM/$DD/0 C xy kg/m2/s CH4 3 1 500
```
msulprizio, line 166: I think this would remain `HEMCO_sa_diagnostics.$YYYY$MM$DD0000.nc`. Regardless of running the IMI with GCClassic or GCHP, the HEMCO standalone would run in the same manner. Therefore the same HEMCO SA diagnostic file would be saved out.

run/GCHP/HEMCO_Config.rc.templates/HEMCO_Config.rc.carbon
msulprizio: See my comment above. Yes, it should remain the monthly timestamp `1-12`.

run/GCHP/HEMCO_Config.rc.templates/HEMCO_Config.rc.carbon
```diff
@@ -1360,21 +1335,18 @@ ${RUNDIR_CO2_COPROD}
...
 
-(((OH_PosteriorSF
-4 OH_SF      Post_SF_OH.nc         SF_OH       2010-2017/1/1/0    C xy 1 1
-)))OH_PosteriorSF
-
 #==============================================================================
 # --- Scale factors for analytical inversions ---
 #==============================================================================
 (((AnalyticalInversion
+4 SCALE_PERT      ./gridded_pert_scale_1.nc          scale    2000/1/1/0 C xy 1 1
```
msulprizio, line 1360: Remove this line?

run/GCHP/HISTORY.rc.templates/HISTORY.rc.carbon
```diff
@@ -52,9 +52,9 @@ GRID_LABELS: #PE24x144-CF
 #==============================================================================
 # Declare collection names and toggle on/off with #
 #==============================================================================
-COLLECTIONS: 'Emissions',
-             'SpeciesConc',
-             #'Metrics',
+COLLECTIONS: 'SpeciesConc',
+             'Emissions',
+             'Metrics',
```
msulprizio: Keep the `Metrics` diagnostic commented out. It doesn't actually produce useful diagnostic output for the carbon simulation at this time.

msulprizio (changes requested):
Thanks @1Dandan. Please see my comments below.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.CH4
1Dandan: This issue would apply for GCC as well. Because when HEMCO reading the monthly GFED4 and daily scaling factor, they will also be regridded to GCC resolution and then multiplied to get daily GFED4 emissions, which would induce inconsistent fire emission fluxes.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.CH4
1Dandan: From the original `HEMCO_Config.rc.CH4`, it indicates only the file of `JPL_WetCharts_2010-2019.Ensemble_Mean.0.5x0.5.nc` is used. And that file only contains values in 2010.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.CH4
1Dandan: This is used to achieve the capability to use gridded state vector and scaling factor directly, see the PRs of https://github.com/geoschem/HEMCO/pull/330 and https://github.com/geoschem/integrated_methane_inversion/pull/356.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.carbon
1Dandan: I see. Sure. I will change it back to 1-12.
Hi @lizziel, how should we update the time range for this file in `ExtData.rc`. This gridded emission file itself has one time dimension, and the file for the gridded scaling factor has 12 time dimension.

run/GCHP/ExtData.rc.templates/ExtData.rc.carbon
1Dandan: Hi @lizziel, the same question for ExtData. The emission file is one time dimension, but the scaling factor has 12 time dimension. How should we add them in ExtData.rc?

run/GCHP/ExtData.rc.templates/ExtData.rc.carbon
1Dandan: The same comments for `HEMCO_Config.rc` for JPLW_CH4: it only uses the file of `JPL_WetCharts_2010-2019.Ensemble_Mean.0.5x0.5.nc`, which only has one time dimension.

run/GCHP/ExtData.rc.templates/ExtData.rc.carbon
1Dandan: Indeed. These should be removed.

run/GCHP/ExtData.rc.templates/ExtData.rc.carbon
1Dandan: It was for corresponding lines in HEMCO_Config.rc.carbon of:
```
3 EMIS_SF    gridded_posterior.nc            ScaleFactor 2000/1/1/0 C xy 1 1
4 SCALE_PERT      ./gridded_pert_scale_1.nc          scale    2000/1/1/0 C xy 1 1
```

run/GCHP/HEMCO_Config.rc.templates/HEMCO_Config.rc.carbon
1Dandan: This is for GCHP. Because we are running the inversion over GCHP grid, we will need Emission output on GCHP grid as well. Therefore, HEMCO standalone will not be used for IMI-GCHP. Instead, GCHP runs with other processes turned off would be used to generate prior emissions.

run/GCHP/HISTORY.rc.templates/HISTORY.rc.carbon
1Dandan: Sure. This is not intentional.

run/GCHP/ExtData.rc.templates/ExtData.rc.carbon
lizziel: @1Dandan, the scale factors should be applied in HEMCO separate from the reading of the file in MAPL. Changing these lines to read the annual file on the first day of the year should be fine, and indeed cuts down on unnecessary file read.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.carbon
lizziel: @1Dandan, the scaling factor application and data read are separate in GCHP so the frequency should not need to be updated in ExtData.rc. Emissions are updated in GCHP every dynamic timestep so if the monthly scale factor changes the computed emissions should change. Have you found that this is not the case?

run/GCClassic/HISTORY.rc.templates/HISTORY.rc.CH4
```diff
@@ -101,7 +101,7 @@ COLLECTIONS: 'Restart',
   SpeciesConc.duration:       ${RUNDIR_HIST_TIME_AVG_DUR}
   SpeciesConc.mode:           'time-averaged'
   SpeciesConc.fields:         'SpeciesConcVV_?ALL?           ',
-                              ##'SpeciesConcMND_?ALL?          ',
```
lizziel, line 104: Why is this change made here and in other HISTORY.rc files for both GC-Classic and GCHP? @1Dandan, could you add a description of the diagnostic default changes to the main description of the PR and why you made them?

run/GCHP/HISTORY.rc.templates/HISTORY.rc.carbon
```diff
@@ -206,6 +206,10 @@ COLLECTIONS: 'Emissions',
                               'BudgetConvectionTrop_CH4      ', 'GCHPchem',
                               'BudgetConvectionPBL_CH4       ', 'GCHPchem',
                               'BudgetConvectionLevs1to35_CH4 ', 'GCHPchem',
+                              'BudgetChemistryFull_CH4      ', 'GCHPchem',
```
lizziel, line 209: Thanks for catching that this was missing. I assume we didn't have chemistry budget diagnostics for carbon simulation because we did not use KPP. @msulprizio or @yantosca, could you confirm that was the case?

run/GCHP/HEMCO_Config.rc.templates/HEMCO_Config.rc.carbon
```diff
@@ -135,11 +135,10 @@ Mask fractions:              false
     --> AnalyticalInversion    :       false
     --> UseTotalPriorEmis      :       false    # Skips global/regional inventories
     --> Emis_PosteriorSF       :       false    # Apply posterior scale factors to total emis?
-    --> OH_PosteriorSF         :       false    # Apply posterior scale factor  to global OH?
 # -----------------------------------------------------------------------------
 111     GFED                   : on    CH4/CO/CO2
     --> GFED4                  :       true
-    --> GFED_daily             :       true
+    --> GFED_daily             :       false
```
lizziel: See my earlier comment about this for the analogous GC-Classic file.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.CH4
lizziel: I think any change to this default should be discussed with the relevant working groups before we change it. They may not be aware of the issue.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.CH4
lizziel: @msulprizio, we seem to be talking about different files. The config file for GC-Classic and GCHP is `JPL_WetCharts_2010-2019.Ensemble_Mean.0.5x0.5.nc` and not `WetCHARTs_Highest_performance_Ensemble_v1.3.1_2010_2019.nc`. The first one has 12 months of data, for 2010. Should the config files be using the other one instead? I can fix the times so it works for GCHP, in case that was the reason it isn't used yet.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.CH4
lizziel: I made a comment on the HEMCO PR that I think we should not use this for GC-Classic for now, but have the new feature be an extra feature to make GCHP work. If it is not needed for GC-Classic, i.e. gives identical results as the old way, then I think we should keep the old way too.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.CH4
```diff
@@ -878,18 +863,15 @@ ${RUNDIR_GLOBAL_Cl}
...
 
 # Add perturbations to individual state vector element (N) following this format
-# Start scale factor ID at 2000 to avoid conflicts with other SFs/masks
-#200N SCALE_ELEM_000N ./Perturbations.txt - - - xy count 1
+# Start scale factor ID at 3000 to avoid conflicts with other SFs/masks
```
lizziel, line 884: See comment above. Also, why did it change from 2000 to 3000? Is this needed for GC-Classic?

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.carbon
```diff
@@ -73,7 +73,7 @@ Mask fractions:              false
     --> GRPI                   :       true     # 2022
     --> EDGARv8                :       true     # 2010-2022
     --> QFED2                  :       false    # 2009-2015
-    --> JPL_WETCHARTS          :       true     # 2009-2017
+    --> JPL_WETCHARTS          :       true     # 2010
```
lizziel, line 74: See my comments on the CH4 HEMCO_Config.rc since they are same for this file.

run/GCHP/ExtData.rc.templates/ExtData.rc.carbon
lizziel: Yes, I agree with @1Dandan. @msulprizio, I wonder if we should be using the other file? Or maybe this gets swapped out for the other file when setting up the IMI?

run/GCHP/ExtData.rc.templates/ExtData.rc.carbon
lizziel: See my comment on the HEMCO PR. I wonder, however, if we should comment out all lines in ExtData.rc that are used only for IMI. They can be uncommented when setting up IMI run directories. This could avoid confusion about what is used for a regular GCHP run.

run/GCHP/HEMCO_Config.rc.templates/HEMCO_Config.rc.carbon
lizziel: I discussed this with DJJ and he agrees the GCHP forward model should be used to generate prior emissions. We need to avoid regridding error between prior and posterior, and the only way to do this is to have them on the same grid.

run/GCHP/HEMCO_Config.rc.templates/HEMCO_Config.rc.carbon
```diff
@@ -1360,21 +1335,18 @@ ${RUNDIR_CO2_COPROD}
 # Entries below are provided for examples only. Add your own here!
 #==============================================================================
 (((Emis_PosteriorSF
-3 EMIS_SF    gridded_posterior.nc  ScaleFactor 2000/1/1/0         C xy 1 1
+3 EMIS_SF    gridded_posterior.nc            ScaleFactor 2000/1/1/0 C xy 1 1
```
lizziel, line 1353: The naming here is a bit confusing for an example. Maybe change name to "scale_factors_for_prior.nc"? Assuming I am interpreting how this is used correctly...

run/GCHP/HEMCO_Config.rc.templates/HEMCO_Config.rc.carbon
```diff
@@ -1360,21 +1335,18 @@ ${RUNDIR_CO2_COPROD}
...
 # --- Scale factors for analytical inversions ---
 #==============================================================================
 (((AnalyticalInversion
+4 SCALE_PERT      ./gridded_pert_scale_1.nc          scale    2000/1/1/0 C xy 1 1
 
 # Add perturbations to individual state vector element (N) following this format
-# Start scale factor ID at 2000 to avoid conflicts with other SFs/masks
-#200N SCALE_ELEM_000N ./Perturbations.txt - - - xy count 1
+# Start scale factor ID at 3000 to avoid conflicts with other SFs/masks
+#300N SCALE_ELEM_000N ./StateVector.nc StateVector 2000/1/1/0 C xy 1 1 N
```
lizziel, line 1364: Again, I think the naming maybe could be more descriptive. If GC-Classic describes it as perturbations maybe this could change to something like that? Why is the hierarchy N?

lizziel (commented):
@1Dandan and @msulprizio, I made an initial pass-through review of this. After reviewing the other PRs and testing the HEMCO update I may have a different perspective. If you could answer the questions I asked that would be great.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.CH4
1Dandan: Yeah, although this is a general problem for applying a gridded scaling factor to emissions, open fire emissions are particularly sensitive to it as daily scaling factor varies a lot for fires. It may be just minor differences for other gridded scaling factor.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.CH4
1Dandan: It should be the same as the old way for IMI-GCC. However, I feel it would be easier to synchronize future updates between IMI-GCC and IMI-GCHP. But if finally it is preferred to have more divergent code blocks between IMI-GCC and IMI-GCHP, I can definitely change the IMI code too.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.CH4
1Dandan: It is need if using carbon simulation instead of methane simulation. Currently we use methane simulation for GCC, if carbon simulation was switched, it would cause problem using scaling factor starting 2000.

run/GCClassic/HISTORY.rc.templates/HISTORY.rc.CH4
1Dandan: Oh, nothing particularly important, I think I was just cleaning out additional "#".

run/GCHP/HEMCO_Config.rc.templates/HEMCO_Config.rc.carbon
1Dandan: This is used by default for IMI-GCC. It was kind of legacy to keep it the same from the python code which generated this file. Also, the target we are actually optimizing was the scaling factor, so I feel it is OK to keep it as it is.

run/GCHP/HEMCO_Config.rc.templates/HEMCO_Config.rc.carbon
1Dandan: Actually N is not the hierarchy, but the mask ID over which gridded scaling factor would be applied as implemented in https://github.com/geoschem/HEMCO/pull/330.

run/GCHP/HISTORY.rc.templates/HISTORY.rc.carbon
1Dandan: For methane, we would have chemistry budget from the reaction with OH and Cl radicals, and thus I added them here for CH4.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.carbon
1Dandan: I see. Thanks @lizziel for clarifications.

run/GCHP/ExtData.rc.templates/ExtData.rc.carbon
1Dandan: One more thought, to minimize the differences between IMI-GCC and IMI-GCHP, gridded scaling factor including OH scaling factor is needed, and thus I restored these lines again. See https://github.com/geoschem/integrated_methane_inversion/pull/356.

run/GCHP/HEMCO_Config.rc.templates/HEMCO_Config.rc.carbon
lizziel: Got it! Thanks for the clarification.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.CH4
lizziel: Melissa and I discussed this and think we can go with the new way, pending testing that demonstrates no differences are introduced.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.CH4
msulprizio: @1Dandan @lizziel You two are correct. I was looking at the file for the highest performance ensemble member. This change makes sense now and is good to keep.

run/GCHP/ExtData.rc.templates/ExtData.rc.carbon
msulprizio: I made a correction above. This update looks good to keep.

run/GCHP/HEMCO_Config.rc.templates/HEMCO_Config.rc.carbon
msulprizio: This makes sense. Thanks!

run/GCHP/HISTORY.rc.templates/HISTORY.rc.carbon
msulprizio: @lizziel I think it was just an omission since the carbon simulation always used KPP (the CH4 simulation does not).

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.CH4
msulprizio: I agree with @1Dandan that it makes sense to be consistent between HEMCO_Config.rc for CH4 and carbon simulations. The carbon simulation already has scale factors starting at 2000 for "Masks used for CO and CO2 simulation emissions." Hence the need to bump these ones to 3000.

run/GCHP/HEMCO_Config.rc.templates/HEMCO_Config.rc.carbon
msulprizio: I agree with @1Dandan. We should keep the filename as is for now to avoid breaking the IMI.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.CH4
msulprizio: Good point. This is an issue in general with HEMCO but no one has seen fit to modify HEMCO to apply scale factors before regridding.

We should keep `GFED_daily : true` as the default for CH4 and carbon simulations. Could you revert it here and elsewhere in the PR @1Dandan?

As mentioned above, if users are concerned about the regridding impact, they can always turn it off in their own runs.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.CH4
msulprizio: I agree this is fine to keep for now. I'll resolve this comment.

run/GCHP/ExtData.rc.templates/ExtData.rc.carbon
msulprizio: OK thanks for clarifying @1Dandan

msulprizio (changes requested):
The one remaining request that I have is to revert `GFED_daily` to true. Then I think it should be OK to merge, assuming @lizziel has no other concerns.

1Dandan:
Hi @msulprizio, I have reverted back to use GFED daily.

CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ This file documents all notable changes to the GEOS-Chem repository starting in
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unrealeased] - TBD
+### Fixed
+- Fixed time-range and units for CH4 emission inventories to be consistent with the corresponding netCDF files in ExtData directory for `HEMCO_Congig.rc` and `ExtData.rc`
+- Use monthly GFED4 intead of daily GFED4 to avoid inconsistent fire emissions across resolutions
```
msulprizio: @1Dandan I just realized this line should also be deleted now

1Dandan: Thanks @msulprizio. Indeed, I just deleted this line.

run/GCClassic/HEMCO_Config.rc.templates/HEMCO_Config.rc.CH4
lizziel: Sounds good.

run/GCClassic/HISTORY.rc.templates/HISTORY.rc.CH4
lizziel: Okay. We have some logic for removing hashes in HISTORY.rc during run directory creation but it seems only applicable to benchmark simulations.

lizziel:
I rebased this PR on dev/14.7.0 and pushed edits to the changelog. I also verified the benchmark and carbon simulations are no-diff with this update. @yantosca, @msulprizio - do you want to review again?

yantosca (approved):
Given that @lizziel and @msulprizio have already approved this PR I will approve as well.  (I tried to remove myself as a reviewer and it wouldn't let me.)

yantosca (approved):

PR #2959, opened 2025-07-21
Dynamically scale GCHP mass flux and courant number based on timestep

yuanjianz:
Diagnosed mass flux and courant number are already time-integrated. So we need to adapt it to the current model timestep by applying a scale of t_run / t_diag.

Name: Yuanjian Zhang
Institution: WashU

### Describe the update

Dynamically scale GCHP mass flux and courant number based on timestep.

Diagnosed mass flux and courant number are already time-integrated. So
we need to adapt it to the current model timestep by applying a scale of
t_run / t_diag (450s).

### Expected changes

This update can help correct mass flux for GCHP.

### Related Github Issue

https://github.com/geoschem/GCHP/issues/503 https://github.com/GEOS-ESM/MAPL/issues/3093

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ This file documents all notable changes to the GEOS-Chem repository starting in 
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## TBD
+## Fixed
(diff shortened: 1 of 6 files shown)


yantosca (approved):
Thanks @yuanjianz!  We can get this into 14.6.3.

lizziel:
This needs more work since grid resolution needs to be incorporated as well for compatibility with the version of MAPL we currently use. I'll create another PR to go into dev/no-diff-to-benchmark.

PR #2967, opened 2025-07-23
Add hidden option to read GC-Classic restart file as REAL8 not through HEMCO

lizziel:

Name: Lizzie Lundgren

### Describe the update

This PR adds a hidden option to read the GC-Classic restart file locally in the GEOS-Chem repository as REAL8. This option is only available for simulations other than nested and Hg. It is a hidden feature, meaning the entry for it is not included in the `geoschem_config.yml` file templates used for making run directories. To tap into the feature uses need to manually add simulation menu entry `read_restart_as_real8` and set it to true. This will result in reading the GC-Classic restart file locally in GEOS-Chem as REAL8 and use those values rather than the values read by HEMCO which are in REAL4.
    
This update is to facilitate testing bit-for-bit reproducibility of GC-Classic when breaking up a run in time. Ideally HEMCO would be able to read file data as REAL8 but that feature has not yet been implemented. Once it is the capability to read the GC-Classic restart file locally as REAL8 in GEOS-Chem can be removed.

### Expected changes

This is a no diff update for all simulations.

### Reference(s)

None

### Related Github Issue

This update builds upon https://github.com/geoschem/geos-chem/pull/2521


GeosCore/hco_utilities_gc_mod.F90
```diff
@@ -1639,6 +1646,28 @@ SUBROUTINE Get_GC_Restart( Input_Opt, State_Chm, State_Grid, &
...
+   INTEGER            :: fId, ierr, varid
+   CHARACTER(LEN=255) :: nc_dir              ! netCDF directory name
+   CHARACTER(LEN=255) :: nc_file             ! netCDF file name
+   CHARACTER(LEN=255) :: nc_path             ! netCDF path name
+   CHARACTER(LEN=255) :: a_name              ! netCDF attribute name
+   CHARACTER(LEN=255) :: a_val               ! netCDF attribute value
+   CHARACTER(LEN=6)   :: TimeStr
+   CHARACTER(LEN=8)   :: DateStr
+   INTEGER            :: st3d(3), ct3d(3)    ! For 3D arrays, including time
+   INTEGER            :: st4d(4), ct4d(4)    ! For 4D arrays, including time
```
yantosca: @lizziel: I think the 4D comment should have the text "species" instead of "time"

GeosCore/hco_utilities_gc_mod.F90
```diff
@@ -1675,6 +1729,49 @@ SUBROUTINE Get_GC_Restart( Input_Opt, State_Chm, State_Grid, &
...
+      nc_path = TRIM( nc_dir ) // TRIM( nc_file )
+
+      ! Test if the file exists
+      INQUIRE( FILE=TRIM( nc_path ), EXIST=FileExists )
+
+      ! Test if the file exists and define an output string
+      IF ( FileExists ) THEN
+         Msg = 'Restart file: Opening'
+      ELSE
+         Msg = 'Rstart file: REQUIRED FILE NOT FOUND'
```
yantosca: Fix typo: "Rstart" -> "Restart"

GeosCore/hco_utilities_gc_mod.F90
```diff
@@ -1700,10 +1797,34 @@ SUBROUTINE Get_GC_Restart( Input_Opt, State_Chm, State_Grid, &
...
+         IF (ierr /= NF90_NOERR) THEN
+            FOUND = .FALSE.
+         ELSE
+            ! Read from file
+            st4d   = (/  1,  1,  1,  1 /)
+            ct4d   = (/ State_Grid%NX, State_Grid%NY, State_Grid%NZ, 1 /)
+            CALL NcRd( Ptr4D_r8, fId, TRIM(v_name_in_file), st4d, ct4d )
+            FOUND = .TRUE.
+
+            ! Read the T:units attribute
```
yantosca: Fix comment "T:units" -> "units"

GeosCore/hco_utilities_gc_mod.F90
```diff
@@ -1719,26 +1841,30 @@ SUBROUTINE Get_GC_Restart( Input_Opt, State_Chm, State_Grid, &
...
          !$OMP COLLAPSE( 3 )
          DO L = 1, State_Grid%NZ
          DO J = 1, State_Grid%NY
          DO I = 1, State_Grid%NX
-            Spc(N)%Conc(I,J,L) = Ptr3D(I,J,L)
+            IF ( use_HEMCO_for_restart_read ) THEN
+               Spc(N)%Conc(I,J,L) = Ptr3D(I,J,L)
+            ELSE
+               Spc(N)%Conc(I,J,L) = Ptr4D_r8(I,J,L,1)
+            ENDIF
```
yantosca: Now that we have an IF statement in the loop, I think it'd be more efficient if we removed the parallel loop and just used array assignment:
```F90
            IF ( use_HEMCO_for_restart_read ) THEN
               Spc(N)%Conc = Ptr3D
            ELSE
               Spc(N)%Conc = Ptr4D_r8(:,:,:,1)
            ENDIF

yantosca (changes requested):
Hi @lizziel, this looks good, I just flagged a couple of minor issues.

GeosCore/hco_utilities_gc_mod.F90
lizziel: By time I meant that 3D is two dimensions of space and one of time. Otherwise people might be confused about what the 4D array is. Does that make sense? I could clarify with "3D arrays (lon,lat,time)"

GeosCore/hco_utilities_gc_mod.F90
lizziel: Now fixed

GeosCore/hco_utilities_gc_mod.F90
lizziel: Now fixed

GeosCore/hco_utilities_gc_mod.F90
lizziel: Good idea. Done!

GeosCore/hco_utilities_gc_mod.F90
lizziel: I ended up just removing the ", including time" comment. That makes the variable more versatile.

Issue #2956, opened 2025-07-21
Hg ocean <-> atmosphere fluxes: components do not sum to net flux


agriff86:
### Your name

Alan Griffiths

### Your affiliation

ANSTO

### What happened? What did you expect to happen?

I'm new to GEOS-Chem and just getting started with the Hg model, so when reading this please consider that I might be the problem :)

With the default Hg configuration (offline Hg ocean emissions), diagnostics output enabled, netCDF output contains the net flux of Hg0 ( `EmisHg0ocean` ) as well as the two components which make up the net flux ( `FluxHg0fromOceanToAir` , `FluxHg0fromAirToOcean` ).  But the sum of the component fluxes is a factor of 2 smaller than the net flux, i.e.

(`FluxHg0fromOceanToAir` - `FluxHg0fromAirToOcean`) == 1/2 `EmisHg0ocean`

The values reported for `EmisHg0ocean` look good (netCDF output matches the output I see by adding `PRINT` statements in the routine which calculates the flux), so I'd say the component fluxes are small by a factor of two.  I think that this is because the component fluxes are sent to the history module twice per timestep, half of the time with zero values.  Here's what happens, in more detail:

1. Start of timestep
2. `FluxHg0fromOceanToAir` is initialised to 0.0 in in `zero_diagnostics_startoftimestep`
3. Values are accumulated into a HistoryItem in `history_update` **with a value of zero**
4. `FluxHg0fromOceanToAir` is updated from the newly calculated value (`State_Diag%FluxHg0fromOceanToAir(I,J) = FUP(I,J)` in `GeosCore/mercury_mod.F90` )
5. Values are accumulated a second time into a HistoryItem in `history_update`, this time with the correct value

A quick fix would be to stop setting `FluxHg0fromOceanToAir` and `FluxHg0fromAirToOcean` to zero in `zero_diagnostics_startoftimestep`, but then it is probably worth thinking about what should happen with the online ocean model and also the other diagnostics (`FluxHg2HgPfromAirToOcean` may need to be treated in the same way?).  So, although disabling this bit of code works for me, for this immediate problem:

     IF ( State_Diag%Archive_FluxHg0fromAirToOcean ) THEN
        State_Diag%FluxHg0fromAirToOcean = 0.0_f4
     ENDIF

     IF ( State_Diag%Archive_FluxHg0fromOceantoAir ) THEN
        State_Diag%FluxHg0fromOceanToAir = 0.0_f4
     ENDIF

[...]

PR #2974, opened 2025-07-28
Fix ocean flux diagnostics in Hg simulation

agriff86:
Zero values of Hg flux diagnostics were being sent to the history module resulting in erroneous values (too small by 1/2) for some diagnostics `FluxHg0fromOceanToAir` `FluxHg0fromAirToOcean` in netCDF output. This change means that the flux components equal the net flux `EmisHg0ocean`.

Name: Alan Griffiths
Institution: ANSTO

### Describe the update

Previously the diagnostics were:
 - zeroed at the start of each (transport?) timestep
 - sent to the history module for averaging at each heartbeat timestep, but
 - not computed until the middle of the transport timestep, along with other surface fluxes
Therefore, zero values were being sent to the history module for half of the time. 

This is a minimal change which stops zeroing the diagnostics at the start of the timestep, so that instead the value is carried over from the previous timestep.

### Expected changes

`FluxHg0fromOceanToAir` minus `FluxHg0fromAirToOcean` will equal the net flux `EmisHg0ocean`.  No change to chemistry as the net flux (which is seen by the model) is unchanged.

### Related Github Issue

(https://github.com/geoschem/geos-chem/issues/2956)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ This file documents all notable changes to the GEOS-Chem repository starting in 
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### Fixed
+- In Hg simulation, fix directional ocean flux diagnostics so they equal net flux
+
 ## [14.6.2] - 2025-06-11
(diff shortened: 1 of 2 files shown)


yantosca (approved):
Thanks @agriff86, this looks good to merge.  We'll get into an upcoming version soon.

PR #3013, opened 2025-09-04
Load-Balancing Solutions for Chemistry module

Jordan-Sun:

Name: Daisy Wang / Jordan Sun

#### Describe
GCHP’s chemistry step exhibits **substantial per‑column workload variability** because the KPP solver takes a different number of sub‑steps depending on local atmospheric state and time. When columns are pinned to MPI ranks and ranks are sequentially mapped to cores in nodes, this variability creates **rank‑ and node‑level imbalance** (some cores and subsequenly hosts have mostly “heavy” columns, others mostly “light”), leaving resources idle while stragglers finish. We propose Hierarchical Balancing with MPI Shared Memory approch to addresses this by (i) **round‑robin rank to node mapping** to decorrelate node‑level loads, and (ii) **MPI RMA based intra‑node shared‑memory pooling** that dynamically redistributes column work across the ranks on a node during the chemistry phase. The redistribution follows a **Decouple - Compute - Remap** pipeline that packs per‑column KPP state into 1‑D arrays, executes chemistry from a node‑local work pool, then restores canonical ownership before transport, preserving ESMF/MAPL contracts and halo semantics. The approach yields large runtime reductions (**20-40%** in our experiments) without requiring a priori load information.  

- Both the modification and the test results are based on version 14.5.2.
- We also tested the modification on the latest version (14.6.3).

#### Modified Scope
- `GeosCore/fullchem_mod.F90`: `Init_FullChem`, `Do_FullChem`, `Cleanup_FullChem`
- `GeosCore/gc_environment_mod.F90`: `GC_Init_Extra` (call update)

#### Key edits
- **`Init_FullChem`**
  - **Signature change:**  
    From: `SUBROUTINE Init_FullChem(Input_Opt, State_Chm, State_Diag, RC)`  
    To:   `SUBROUTINE Init_FullChem(Input_Opt, State_Chm, State_Diag, State_Grid, RC)`  
    (adds `State_Grid`).
  - **Intra-node MPI-RMA based Shared Memory setup:**
    - Create `shm_comm` via `MPI_Comm_split_type`.
    - Gather per-rank cell counts; compute **prefix offsets** and `NCELL_total`; broadcast offsets.
    - Allocate/attach MPI **shared-memory windows** for packed chemistry arrays (see list below).
    - Allocate `Idx_to_IJL(3, NCELL_MAX)` to record the mapping \( $$N \leftrightarrow (I,J,L)$$ \).

- **`Do_FullChem`**
  - Replace the single triple loop over `(I,J,L)` with **Decouple - Compute - Remap**:
    - **Decouple (pack):** pack each rank’s per-column KPP state into its segment of node-wide 1-D arrays; 
    - **Compute:** enable **dynamic, node-local work sharing** via a shared atomic counter `next_cell_index` (`MPI_Fetch_and_op`); each rank claims the next cell and **integrates in place** in the shared arrays; optional `cell_status(N)` for tracing.
    - **Remap (unpack):** copy updated state for originally owned columns back to the rank’s grid; proceed to transport.

- **`Cleanup_FullChem`**
  - Free all `win_*` shared windows, deallocate bookkeeping arrays, and free `shm_comm`.

- **`GC_Init_Extra` (in `gc_environment_mod.F90`)**
  - Update the call to `Init_FullChem` to pass the new `State_Grid` argument.

#### Assumptions / constraints
- KPP is **cell-independent**; canonical ESMF/MAPL decomposition is restored before non-chemistry phases.
- Redistribution is **scoped to chemistry only**.
- Requires **MPI-3 shared memory / RMA** support.

### Expected changes

**Output: scientific results remain unchanged.** 
Example: C180, 2019-07-02:
```bash
Variable Xdim matches within the threshold of 1e-07.
Variable Ydim matches within the threshold of 1e-07.
Variable nf matches within the threshold of 1e-07.
Variable ncontact matches within the threshold of 1e-07.
Variable contacts matches within the threshold of 1e-07.
Variable anchor matches within the threshold of 1e-07.
Variable lons matches within the threshold of 1e-07.
Variable lats matches within the threshold of 1e-07.
Variable corner_lons matches within the threshold of 1e-07.
Variable corner_lats matches within the threshold of 1e-07.
Variable lev matches within the threshold of 1e-07.
Variable time matches within the threshold of 1e-07.
Variable Met_AIRDEN matches within the threshold of 1e-07.
Variable Met_BXHEIGHT matches within the threshold of 1e-07.
Variable Met_PBLH matches within the threshold of 1e-07.
Variable Met_PMID matches within the threshold of 1e-07.
Variable Met_PMIDDRY matches within the threshold of 1e-07.
Variable Met_PS1DRY matches within the threshold of 1e-07.
Variable Met_PS1WET matches within the threshold of 1e-07.
Variable Met_RH matches within the threshold of 1e-07.
Variable Met_T matches within the threshold of 1e-07.
Variable Met_TropHt matches within the threshold of 1e-07.
Variable Met_TropP matches within the threshold of 1e-07.
Variable PM25 matches within the threshold of 1e-07.
Variable SpeciesConcVV_CO matches within the threshold of 1e-07.
Variable SpeciesConcVV_NO matches within the threshold of 1e-07.
Variable SpeciesConcVV_NO2 matches within the threshold of 1e-07.
Variable SpeciesConcVV_O3 matches within the threshold of 1e-07.
Variable SpeciesConcVV_OH matches within the threshold of 1e-07.
All variables match within the given threshold.
```

**Performance: reduces wall time by ~24–41% across platforms and resolutions.**
> **Setup:** 576 ranks (cores) for C48–C180; 1152 ranks (cores) for C360; GCC + Intel MPI.

| Platform | Resolution | Original Time (s) | HB-SHM Time (s) | Time Saved (s) | Reduction | Speedup |
|---|---:|---:|---:|---:|---:|---:|
| Compute1 | C48  | 1825.2 | 1392.3 | 432.9  | 23.72% | 1.31× |
| Compute1 | C90  | 4840.6 | 3381.4 | 1459.2 | 30.15% | 1.43× |
| Compute1 | C180 | 16776.9 | 11540.4 | 5236.5 | 31.21% | 1.45× |
| AWS      | C48  | 1737.8 | 1275.7 | 462.0  | 26.59% | 1.36× |
| AWS      | C90  | 4925.0 | 3056.8 | 1868.2 | 37.93% | 1.61× |
| AWS      | C180 | 17352.5 | 10277.4 | 7075.2 | 40.77% | 1.69× |
| AWS      | C360 | 55499.7 | 34156.0 | 21343.7 | 38.46% | 1.62× |

**Memory overhead: additional memory is required for shared windows.**  
> **Assumptions**: REAL/INTEGER = 8 B; Nspec=356, Nreact=1058, N(R)_state=20, N(I)_ctrl=20, N(R)_ctrl=20, N(I)_stat=20. Per-node values assume the noted node counts in our layouts.

- **C48:** 995,328 cells — **Total extra ≈ 11.12 GiB**; with 4 nodes ⇒ **≈ 2.78 GiB per node**  
- **C90:** 3,499,200 cells — **Total extra ≈ 39.08 GiB**; with 16 nodes ⇒ **≈ 2.44 GiB per node**  
- **C180:** 13,996,800 cells — **Total extra ≈ 156.32 GiB**; with 16 nodes ⇒ **≈ 9.77 GiB per node**  
- **C360:** 55,987,200 cells — **Total extra ≈ 625.29 GiB**; with 16 nodes ⇒ **≈ 39.08 GiB per node**  

Note: per-cell footprint ≈ 12 KB ≈ 1,500 doubles; constant across resolutions.

### Reference(s)

Please refer to [recently submitted and still unpublished work](https://drive.google.com/file/d/1OGxPt1m9OVKSSxCQUO1YPXk-qUg5oZNs/view).

### Related Github Issue

Our approach effectively addresses the load-imbalance problem discussed in [GCHP#52](https://github.com/geoschem/GCHP/issues/52).


yantosca:
Thanks @Jordan-Sun and @Daisy0419,  We will take a look at this.

Question: Have you also tried this update with GEOS-Chem Classic?  We should make sure that it doesn't break any GEOS-Chem Classic functionality or worsen GEOS-Chem Classic performance before we bring these updates in.

Tagging @lizziel @msulprizio @msl3v

KPP/custom/custom.kpp
```diff
@@ -1,14 +1,14 @@
...
+#AUTOREDUCE   on                     { ... with autoreduce enabled but optional   }
+#LANGUAGE     Fortran90              { Generate solver code in Fortran90 ...      }
+#UPPERCASEF90 on                     { ... with .F90 suffix (instead of .f90)     }
+#DRIVER       none                   { Do not create gckpp_Main.F90               }
+#HESSIAN      off                    { Do not create the Hessian matrix           }
+#MEX          off                    { MEX is for Matlab, so skip it              }
+#STOICMAT     off                    { Do not create stoichiometric matrix        }
 
-#INCLUDE custom.eqn       { Chemical reactions for fullchem mechanism  }
+#INCLUDE custom.eqn       { Chemical reactions for custom mechanism  }
```
yantosca: @Jordan-Sun @Daisy0419 in 14.6.3 and later we now use `#MINVERSION 3.2.0`, as this version contains the functionality for the KPP-standalone model.

GeosCore/chemistry_mod.F90
```diff
@@ -3,1299 +3,3708 @@
 !------------------------------------------------------------------------------
 !BOP
 !
-! !MODULE: chemistry_mod.F90
+! !MODULE: fullchem_mod.F90
```
yantosca: @Jordan-Sun @Daisy0419: It looks like fullchem_mod has clobbered the existing chemistry_mod.  This will break the existing workflow for the non-fuillchem simulations.

Would it be possible to restore the original chemsitry_mod, and then to call fullchem_mod (and the new module that you are calling fullchem_mod) from there)?

GeosCore/chemistry_mod.F90
```diff
@@ -3,1299 +3,3708 @@
...
+   !  ENDDO
+
+   IF (shm_rank == 0) THEN
+      next_cell_index = 1
+   END IF
+
+   ! flush processor caches for every window you just wrote
+   ! Start access epochs for all shared-memory windows
+   CALL MPI_Win_lock_all(0, win_cost_1D, RC)
+   CALL MPI_Win_lock_all(0, win_C_1D, RC)
```
yantosca: I believe that adding the MPI code here will break other implementations of GEOS-Chem (such as GEOS-Chem Classic) which do not use MPI.  We would have to wrap these with e.g. `#ifdef MODEL_GCHP` statements

Also, to reduce clutter in the code, I would recommend abstracting these MPI calls into subroutines whenever possible.

KPP/custom/fullchem_SulfurChemFuncs.F90
yantosca, line 1: The `KPP/custom/fullchem_SulfurChemFuncs.F90` should be left as a symbolic link to `KPP/fullchem/fullchem_SulfurChemFuncs.F90`.

yantosca, line 1: This symbolic link should be restored.

GeosCore/chemistry_mod.F90
```diff
@@ -3,1299 +3,3708 @@
...
+   ! flat pointers returned by MPI (INTEGER addresses)
+   INTEGER(KIND=MPI_ADDRESS_KIND) :: cost_1D_flat_int
+   INTEGER(KIND=MPI_ADDRESS_KIND) :: C_1D_flat_int
+   INTEGER(KIND=MPI_ADDRESS_KIND) :: RCONST_1D_flat_int
+   INTEGER(KIND=MPI_ADDRESS_KIND) :: ICNTRL_1D_flat_int
+   INTEGER(KIND=MPI_ADDRESS_KIND) :: RCNTRL_1D_flat_int
+   INTEGER(KIND=MPI_ADDRESS_KIND) :: ISTATUS_1D_flat_int
+   INTEGER(KIND=MPI_ADDRESS_KIND) :: RSTATE_1D_flat_int
+   INTEGER(KIND=MPI_ADDRESS_KIND) :: cell_status_flat_int
+   INTEGER(KIND=MPI_ADDRESS_KIND) :: next_cell_index_int   ! scalar
```
yantosca: I think the `MPI_ADDRESS_KIND` parameter will break non-MPI implementations of GEOS-Chem.   We should consider placing these in an `#if defined( MODEL_GCHP ) or defined( MODEL_GEOS )` block, so as not to break non-MPI implementations of GEOS-Chem.

README.md
```diff
@@ -1,41 +1,65 @@
-[![Release](https://img.shields.io/github/v/release/geoschem/geos-chem?label=Latest%20Release)](http://wiki.geos-chem.org/GEOS-Chem_versions)
-[![DOI](https://zenodo.org/badge/DOI/10.5281/zenodo.1343546.svg)](https://doi.org/10.5281/zenodo.1343546)
-[![License](https://img.shields.io/badge/License-MIT-blue.svg)](https://github.com/geoschem/geos-chem/blob/master/LICENSE.txt)
+# Hierarchical Load Balancing with MPI Shared Memory for GCHP
```
lizziel: This seems to be replacing the README for the GEOS-Chem repository. We should keep the GEOS-Chem README the same and move this information elsewhere, ideally ReadTheDocs.

Headers/state_diag_mod.F90
```diff
@@ -862,6 +862,12 @@ MODULE State_Diag_Mod
      REAL(f4),           POINTER :: KppNegatives0(:,:,:)
      LOGICAL                     :: Archive_KppNegatives0
 
+     REAL(f4),           POINTER :: KppRank(:,:,:)
```
lizziel: New diagnostics should also be included in HISTORY.rc files.

GeosCore/chemistry_mod.F90
```diff
@@ -3,1299 +3,3708 @@
...
+#ifdef MODEL_GEOS
+    !$OMP PRIVATE( NOxTau,     NOxConc, NOx_weight, NOx_tau_weighted        )&
+#endif
+    !$OMP COLLAPSE( 3                                                       )&
+    !$OMP SCHEDULE( DYNAMIC, 24                                             )&
+    !$OMP REDUCTION( +:errorCount                                           )
+
+   origin_val = 1                      ! add 1 each time
+   disp       = 0_MPI_ADDRESS_KIND     ! first integer in the window
+   CALL MPI_Win_lock_all(MPI_MODE_NOCHECK, win_next_cell_index, ierr)
```
yantosca: We would need to block this call out with an #ifdef

GeosCore/chemistry_mod.F90
```diff
@@ -3,1299 +3,3708 @@
...
+      CALL MPI_Win_flush_local(0, win_next_cell_index, ierr)
+
+      ! WRITE (*,*) 'Rank', shm_rank, 'got ticket', fetched_value
+      IF (fetched_value > NCELL_total) EXIT
+      
+      I_CELL = fetched_value
+      IF (cell_status(I_CELL) > 0) CYCLE
+
+      cell_status(I_CELL) = cell_status(I_CELL) + 1
+
```
yantosca: We would need to block out these calls with an #ifdef to avoid breaking non-MPI implementations

GeosCore/chemistry_mod.F90
```diff
@@ -3,1299 +3,3708 @@
...
-                             State_Diag = State_Diag,                &
-                             State_Grid = State_Grid,                &
-                             State_Met  = State_Met,                 &
-                             RC         = RC                        )
-#endif
+    CALL MPI_Barrier(shm_comm, RC)
+    IF (RC /= MPI_SUCCESS) THEN
+      CALL GC_Error('MPI_Barrier failed', RC, ThisLoc)
+      RETURN
+    ENDIF
```
yantosca: We would need to block this call out with an #ifdef

GeosCore/rdtsc.c
```diff
@@ -0,0 +1,11 @@
+#include <stdint.h>
```
lizziel: Is the code original or borrowed? It is a utility so seems like it should be put someplace other than GeosCore. It also needs documentation explaining what it is, .e.g. "rdtsc() utilizes inline assembly to read the Time-Stamp Counter (TSC) of the CPU."

GeosCore/chemistry_mod.F90
```diff
@@ -3,1299 +3,3708 @@
...
+    ! This is needed for input to the KPP standalone box model.
+    !------------------------------------------------------------------------
+    CALL KppSa_Config( Input_Opt, RC )
+    IF ( RC /= GC_SUCCESS ) THEN
+       ErrMsg = 'Error encountered in "KPP_Standalone"!'
+       CALL GC_Error( ErrMsg, RC, ThisLoc )
+       RETURN
+    ENDIF
+
+   !setup shared mamory
```
yantosca: MPI-specific code would need to be blocked off with an #ifdef to avoid breaking non-MPI implmentations of GEOS-Chem

GeosCore/chemistry_mod.F90
lizziel: The names of these modules are different. Is this an old version of the file?

GeosCore/fullchem_mod.F90
```diff
@@ -11,13 +11,19 @@
 !\\
 ! !INTERFACE:
 !
+! Only define HIRES_TIMER or MPI_TIMER, never both.
```
lizziel: Why is this comment here?

lizziel (changes requested):
I did some very initial comments on this PR to start.

yantosca (changes requested):
Thanks much @Jordan-Sun and @Daisy0419.  I have a few concerns about your PR as written.

1. `chemistry_mod.F90` is replaced by `fullchem_mod.F90`.  This will break the non-full-chemistry simulations, as the calls to those routines are located there.
2. There are a lot of calls to MPI functions added to GEOS-Chem code.  This will break the non-MPI implementations of GEOS-Chem (such as GEOS-Chem Classic).  These should be blocked out with #ifdefs so as to make it a compile-time option.
3. It would seem that the load-balancing updates in this PR requires some detailed knowledge of how MPI works.  I am concerned that the GCST would not be able to debug or modify this going forward.  It also makes modifying the code more difficult for our users.

GeosCore/chemistry_mod.F90
yantosca: @lizziel: It looks like the existing `chemistry_mod.F90` has been replaced with `fullchem_mod.F90`, which will break the existing functionality for non-fullchem simulations.

lizziel:
Thanks @Daisy0419. When I click on "Files Changed" for this PR I get an error message. There are also lots of conflicts listed which indicate a mismatch and base version of the PR and the target branch main. I think something got messed up in the PR branch. Try doing git log and see if main is in the history.

Headers/state_diag_mod.F90
Jordan-Sun: This was left over from earlier testing and isn’t strictly necessary anymore. Would you prefer that I keep it and add it to HISTORY.rc, or should I just remove it?

msl3v:
Hey all.

First of all, this appears to be a major advancement for GCHP and likely all distributed GC projects! And possibly a generalizable approach for all of the atmospheric chem community faced with the load-balance issue this is addressing. I am sincerely impressed.

Looking at the implementation, it seems to be a sharp change in how the KPP loop is executed. Fullchem_mod.F90 is at the center of GEOS-Chem and its siblings (GCHP, GEOS-CF, WRF-GC, etc). I'm wondering if there are some strategies that can most cleanly implement these changes without either forking the code into two modules or creating "if mpi" chunks of code. A lot of development happens in this module, and spending the time upfront to keep this clean will be very helpful in the long run.

I'm thinking maybe methods that keep the MPI invisible like how ESMF and MAPL currently do. 
If we want to start a separate discussion here, or have a video chat, it might be worth it.

This PR is too important, in my opinion.

yantosca:
Closing this PR.  Please see discussion in PR #3015.

Issue #2536, opened 2024-10-25
Mass conservation upon restart


yuanjianz:
### Your name

Yuanjian Zhang

### Your affiliation

WashU

### What happened? What did you expect to happen?

Both GC-Classics and GCHP are not able to conserve mass upon restart.

@lizziel:
> Initial global mass change upon startup and then stabilizes. Mass conservation is better after a short spin-up period. Issues with unit conversion during restart handling? In theory, passive tracer mass in restart should be conserved upon startup and across a run.

Also, see https://github.com/geoschem/geos-chem/issues/2014 and @lizziel's work for bit-of-bit reproducibility in https://github.com/geoschem/geos-chem/pull/2521

This could be the result of not properly scaling mixing ratios based on DELP_DRY_RESTART and DELP_DRY_MET.

Mixing ratios should be scaled based on `scaled_MR = MR_restart * DELP_DRY_RESTART / DELP_DRY_MET` when everything is first read into the model. Then all processes including transport, convection, and chemistry, etc can be executed.

## GC-Classics
https://github.com/geoschem/geos-chem/blob/bef56c605e018eecbd91646a51ce82c7cd77f56a/GeosCore/hco_interface_gc_mod.F90#L4415-L4419
Currently, the above snippet in `Get_Met_Fields` is not updating mixing ratios since `Get_Met_Fields` is right before `Get_GC_Restart`:
https://github.com/geoschem/geos-chem/blob/bef56c605e018eecbd91646a51ce82c7cd77f56a/GeosCore/hco_interface_gc_mod.F90#L1010-L1033

It is recommended to either move `Get_Met_Fields` after `Get_GC_Restart` or move the reading of `DELP_DRY_RESTART` into `Get_GC_Restart` and complete the AirQnt-update_mixing_ratio there.

## GCHP
GCHP advection is executed before geos-chem module, so it would be necessary to read DELP_DRY_RESTART in ChemComp Initialization state and do the scaling before AdvComp Run state.

lizziel:
This issue should be investigated after we incorporate the GC-Classic reproducibility updates. Regarding GCHP, we achieve bit-for-bit reproducibility upon restart which implies that if this issue is true (no mass conservation upon restart) then it is also true timestep to timestep. Further investigation is needed.

PR #3014, opened 2025-09-05
Conserve initial restart file mass

lizziel:

Name: Lizzie Lundgren

### Describe the update

This is still draft and I will continue to edit this description.

This update changes where restart file species concentrations are scaled by pressure for mass conservation, and fixes a problem where mass was not conserved if restart file meteorology did not match run-time meteorology in first timestep.

GC-Classic changes:
Concentrations are now scaled within `Get_GC_Restart` by the ratio of field `Met_DELPDRY` (delta pressure per level) in the restart file and `State_Met%DELP_DRY`.
If Met_DELPDRY is not found in the restart file then a warning is printed that species mass could not be determined from the restart file and concentrations will be used as is.
If the field is found in the restart file then global mass is computed based on the restart file contents and printed to log along with min, max, and sum of [mol/mol].

GCHP changes:
Concentrations are now scaled in the run method within `Chem_GridCompMod` by the ratio of field `DELP_DRY` (delta pressure per level) in the restart file and `State_Met%DELP_DRY`.
All of these changes are within MODEL_GCHPCTM blocks and thus do not impact GEOS.

Changes/notes affecting both GC-Classic and GCHP:
Concentrations are only scaled for species found in the restart file.
If verbose, then global mass is computed and printed at the start of each GEOS-Chem timestep. This uses new subroutine in `print_mod` called `Print_Species_Global_Mass`.
If delta pressure is not found in the restart file then a warning is printed that species mass could not be determined from the restart file and concentrations will be used as is.

### Expected changes

Small differences. More explanation coming soon.

### Reference(s)

None

### Related Github Issue

closes https://github.com/geoschem/geos-chem/issues/2536


GeosCore/hco_utilities_gc_mod.F90
```diff
@@ -1773,14 +1781,70 @@ SUBROUTINE Get_GC_Restart( Input_Opt, State_Chm, State_Grid, &
...
+   IF ( use_HEMCO_for_restart_read ) THEN
+      CALL HCO_GC_GetPtr( Input_Opt, State_Grid, TRIM( v_name_in_hemco ),  &
+           Ptr3D,     RC,         FOUND=FOUND_DELP_DRY )
+   ELSE
+      Temp4D_r8 = 0.0_fp
+      Ptr4D_r8 => Temp4D_r8
+      st4d   = (/  1,  1,  1,  1 /)
+      ct4d   = (/ State_Grid%NX, State_Grid%NY, State_Grid%NZ, 1 /)
+      CALL NcRd( Ptr4D_r8, fId, TRIM(v_name_in_file), st4d, ct4d )
+      FOUND = .TRUE.
```
yantosca: @lizzlei: Should `FOUND` be `FOUND_DELP_DRY` here?  Otherwise the next IF block won't execute even though we've read `Met_DELPDRY` directly from the restart file.

GeosCore/hco_utilities_gc_mod.F90
```diff
@@ -1854,6 +1918,50 @@ SUBROUTINE Get_GC_Restart( Input_Opt, State_Chm, State_Grid, &
...
+         !       differs from current meteorology
+         IF ( FOUND_DELP_DRY ) THEN
+
+            ! Also print mass based on restart file mixing ratio and meteorology
+            SpcMass = 0.d0
+            SpcMassPtr => SpcMass
+            !$OMP PARALLEL DO       &
+            !$OMP DEFAULT( SHARED ) &
+            !$OMP PRIVATE( I, J, L )
+            DO L = 1, State_Grid%NZ
```
yantosca, line 1934: I would add an `!$OMP COLLAPSE( 3 ) here for efficiency

GeosCore/hco_utilities_gc_mod.F90
```diff
@@ -1854,6 +1918,50 @@ SUBROUTINE Get_GC_Restart( Input_Opt, State_Chm, State_Grid, &
...
+            SpcMassPtr => NULL()
+
+            ! Update mixing ratio to conserve mass if current pressure different
+            ! from restart file pressure
+            IF ( update_mixing_ratio ) THEN
+               ! Update concentrations
+               !$OMP PARALLEL DO       &
+               !$OMP DEFAULT( SHARED ) &
+               !$OMP PRIVATE( I, J, L )
+               DO L = 1, State_Grid%NZ
```
yantosca, line 1955: I would add an `!$OMP COLLAPSE( 3 )` here for efficiency

yantosca (changes requested):
Thanks @lizziel.  I flagged a couple of minor things but otherwise looks good.

GeosCore/hco_utilities_gc_mod.F90
lizziel: Yes. Good catch!

GeosCore/hco_utilities_gc_mod.F90
lizziel: Now added

GeosCore/hco_utilities_gc_mod.F90
lizziel: Now added

PR #3015, opened 2025-09-05
Load-Balancing Solutions for Chemistry Module (Fresh Fork)

Jordan-Sun:
Rebased to the latest geos-chem version, otherwise should be ident to #3013.

Name: Daisy Wang / Jordan Sun

#### Describe
GCHP’s chemistry step exhibits **substantial per‑column workload variability** because the KPP solver takes a different number of sub‑steps depending on local atmospheric state and time. When columns are pinned to MPI ranks and ranks are sequentially mapped to cores in nodes, this variability creates **rank‑ and node‑level imbalance** (some cores and subsequenly hosts have mostly “heavy” columns, others mostly “light”), leaving resources idle while stragglers finish. We propose Hierarchical Balancing with MPI Shared Memory approch to addresses this by (i) **round‑robin rank to node mapping** to decorrelate node‑level loads, and (ii) **MPI RMA based intra‑node shared‑memory pooling** that dynamically redistributes column work across the ranks on a node during the chemistry phase. The redistribution follows a **Decouple - Compute - Remap** pipeline that packs per‑column KPP state into 1‑D arrays, executes chemistry from a node‑local work pool, then restores canonical ownership before transport, preserving ESMF/MAPL contracts and halo semantics. The approach yields large runtime reductions (**20-40%** in our experiments) without requiring a priori load information.  

- Test results are based on version 14.5.2.
- Modification are based and *tested (todo: test on GC-Classic)* on the latest version (14.6.3).

#### Modified Scope
- `GeosCore/fullchem_mod.F90`: `Init_FullChem`, `Do_FullChem`, `Cleanup_FullChem`
- `GeosCore/gc_environment_mod.F90`: `GC_Init_Extra` (call update)

#### Key edits
- **`Init_FullChem`**
  - **Signature change:**  
    From: `SUBROUTINE Init_FullChem(Input_Opt, State_Chm, State_Diag, RC)`  
    To:   `SUBROUTINE Init_FullChem(Input_Opt, State_Chm, State_Diag, State_Grid, RC)`  
    (adds `State_Grid`).
  - **Intra-node MPI-RMA based Shared Memory setup:**
    - Create `shm_comm` via `MPI_Comm_split_type`.
    - Gather per-rank cell counts; compute **prefix offsets** and `NCELL_total`; broadcast offsets.
    - Allocate/attach MPI **shared-memory windows** for packed chemistry arrays (see list below).
    - Allocate `Idx_to_IJL(3, NCELL_MAX)` to record the mapping \( $$N \leftrightarrow (I,J,L)$$ \).

- **`Do_FullChem`**
  - Replace the single triple loop over `(I,J,L)` with **Decouple - Compute - Remap**:
    - **Decouple (pack):** pack each rank’s per-column KPP state into its segment of node-wide 1-D arrays; 
    - **Compute:** enable **dynamic, node-local work sharing** via a shared atomic counter `next_cell_index` (`MPI_Fetch_and_op`); each rank claims the next cell and **integrates in place** in the shared arrays; optional `cell_status(N)` for tracing.
    - **Remap (unpack):** copy updated state for originally owned columns back to the rank’s grid; proceed to transport.

- **`Cleanup_FullChem`**
  - Free all `win_*` shared windows, deallocate bookkeeping arrays, and free `shm_comm`.

- **`GC_Init_Extra` (in `gc_environment_mod.F90`)**
  - Update the call to `Init_FullChem` to pass the new `State_Grid` argument.

#### Assumptions / constraints
- KPP is **cell-independent**; canonical ESMF/MAPL decomposition is restored before non-chemistry phases.
- Redistribution is **scoped to chemistry only**.
- Requires **MPI-3 shared memory / RMA** support.

### Expected changes

**Output: scientific results remain unchanged.** 
Example: C180, 2019-07-02:
```bash
Variable Xdim matches within the threshold of 1e-07.
Variable Ydim matches within the threshold of 1e-07.
Variable nf matches within the threshold of 1e-07.
Variable ncontact matches within the threshold of 1e-07.
Variable contacts matches within the threshold of 1e-07.
Variable anchor matches within the threshold of 1e-07.
Variable lons matches within the threshold of 1e-07.
Variable lats matches within the threshold of 1e-07.
Variable corner_lons matches within the threshold of 1e-07.
Variable corner_lats matches within the threshold of 1e-07.
Variable lev matches within the threshold of 1e-07.
Variable time matches within the threshold of 1e-07.
Variable Met_AIRDEN matches within the threshold of 1e-07.
Variable Met_BXHEIGHT matches within the threshold of 1e-07.
Variable Met_PBLH matches within the threshold of 1e-07.
Variable Met_PMID matches within the threshold of 1e-07.
Variable Met_PMIDDRY matches within the threshold of 1e-07.
Variable Met_PS1DRY matches within the threshold of 1e-07.
Variable Met_PS1WET matches within the threshold of 1e-07.
Variable Met_RH matches within the threshold of 1e-07.
Variable Met_T matches within the threshold of 1e-07.
Variable Met_TropHt matches within the threshold of 1e-07.
Variable Met_TropP matches within the threshold of 1e-07.
Variable PM25 matches within the threshold of 1e-07.
Variable SpeciesConcVV_CO matches within the threshold of 1e-07.
Variable SpeciesConcVV_NO matches within the threshold of 1e-07.
Variable SpeciesConcVV_NO2 matches within the threshold of 1e-07.
Variable SpeciesConcVV_O3 matches within the threshold of 1e-07.
Variable SpeciesConcVV_OH matches within the threshold of 1e-07.
All variables match within the given threshold.
```

**Performance: reduces wall time by ~24–41% across platforms and resolutions.**
> **Setup:** 576 ranks (cores) for C48–C180; 1152 ranks (cores) for C360; GCC + Intel MPI.

| Platform | Resolution | Original Time (s) | HB-SHM Time (s) | Time Saved (s) | Reduction | Speedup |
|---|---:|---:|---:|---:|---:|---:|
| Compute1 | C48  | 1825.2 | 1392.3 | 432.9  | 23.72% | 1.31× |
| Compute1 | C90  | 4840.6 | 3381.4 | 1459.2 | 30.15% | 1.43× |
| Compute1 | C180 | 16776.9 | 11540.4 | 5236.5 | 31.21% | 1.45× |
| AWS      | C48  | 1737.8 | 1275.7 | 462.0  | 26.59% | 1.36× |
| AWS      | C90  | 4925.0 | 3056.8 | 1868.2 | 37.93% | 1.61× |
| AWS      | C180 | 17352.5 | 10277.4 | 7075.2 | 40.77% | 1.69× |
| AWS      | C360 | 55499.7 | 34156.0 | 21343.7 | 38.46% | 1.62× |

**Memory overhead: additional memory is required for shared windows.**  
> **Assumptions**: REAL/INTEGER = 8 B; Nspec=356, Nreact=1058, N(R)_state=20, N(I)_ctrl=20, N(R)_ctrl=20, N(I)_stat=20. Per-node values assume the noted node counts in our layouts.

- **C48:** 995,328 cells — **Total extra ≈ 11.12 GiB**; with 4 nodes ⇒ **≈ 2.78 GiB per node**  
- **C90:** 3,499,200 cells — **Total extra ≈ 39.08 GiB**; with 16 nodes ⇒ **≈ 2.44 GiB per node**  
- **C180:** 13,996,800 cells — **Total extra ≈ 156.32 GiB**; with 16 nodes ⇒ **≈ 9.77 GiB per node**  
- **C360:** 55,987,200 cells — **Total extra ≈ 625.29 GiB**; with 16 nodes ⇒ **≈ 39.08 GiB per node**  

Note: per-cell footprint ≈ 12 KB ≈ 1,500 doubles; constant across resolutions.

### Reference(s)

Please refer to [recently submitted and still unpublished work](https://drive.google.com/file/d/1OGxPt1m9OVKSSxCQUO1YPXk-qUg5oZNs/view).

### Related Github Issue

Our approach effectively addresses the load-imbalance problem discussed in [GCHP#52](https://github.com/geoschem/GCHP/issues/52).


Jordan-Sun:
Hi @yantosca,

As of commit `eff70c7`, we have verified that the update works with GEOS-Chem Classic, producing the same output in roughly the same amount of wall clock time. Please let us know if there are other models we should test this update on.

For more information, see [this pull request](https://github.com/Daisy0419/geos-chem/pull/1).

GeosCore/fullchem_mod.F90
```diff
@@ -109,7 +160,7 @@ SUBROUTINE Do_FullChem( Input_Opt,  State_Chm, State_Diag,                 &
     USE fullchem_AutoReduceFuncs, ONLY : fullchem_AR_SetKeepActive
     USE fullchem_AutoReduceFuncs, ONLY : fullchem_AR_UpdateKppDiags
     USE fullchem_AutoReduceFuncs, ONLY : fullchem_AR_SetIntegratorOptions
-#endif
+#endif
```
lizziel: Remove whitespace added to this line

GeosCore/fullchem_mod.F90
```diff
@@ -225,6 +282,7 @@ SUBROUTINE Do_FullChem( Input_Opt,  State_Chm, State_Diag,                 &
     REAL(dp)               :: ScaleCESMLossRate
 #endif
 
+
```
lizziel: Remove added line

GeosCore/fullchem_mod.F90
```diff
@@ -236,20 +294,27 @@ SUBROUTINE Do_FullChem( Input_Opt,  State_Chm, State_Diag,                 &
     ! Defines the slot in which the H-value from the KPP integrator is stored
     ! This should be the same as the value of Nhnew in gckpp_Integrator.F90
     ! Define this locally in order to break a compile-time dependency.
-    !    -- Bob Yantosca (05 May 2022)
+    !   -- Bob Yantosca (05 May 2022)
```
lizziel: Remove change to this line

GeosCore/fullchem_mod.F90
```diff
@@ -236,20 +294,27 @@ SUBROUTINE Do_FullChem( Input_Opt,  State_Chm, State_Diag,                 &
...
+    INTEGER(KIND=MPI_ADDRESS_KIND) :: disp
+#endif
+   
 
     !========================================================================
     ! Do_FullChem begins here!
     ! NOTE: FlexChem timer is started in DO_CHEMISTRY (the calling routine)
     !========================================================================
 
+
```
lizziel: Remove added line

GeosCore/fullchem_mod.F90
```diff
@@ -329,6 +394,7 @@ SUBROUTINE Do_FullChem( Input_Opt,  State_Chm, State_Diag,                 &
     ENDIF
 #endif
 
+
```
lizziel: Remove added line

GeosCore/fullchem_mod.F90
```diff
@@ -389,131 +455,1262 @@ SUBROUTINE Do_FullChem( Input_Opt,  State_Chm, State_Diag,                 &
...
+       C         = 0.0_dp                   ! KPP species conc's
+       RCONST    = 0.0_dp                   ! KPP rate constants
+       PHOTOL    = 0.0_dp                   ! Photolysis array for KPP
+       K_CLD     = 0.0_dp                   ! Sulfur in-cloud rxn het rates
+       K_MT      = 0.0_dp                   ! Sulfur sea salt rxn het rates
+       CFACTOR   = 1.0_dp                   ! KPP conversion factor
+       SRO3      = 0.0_dp                   ! Enhanced sulfate production of
+       SRHOBr    = 0.0_dp                   !  O3, HOBr, HCl in size-resolved
+       SRHOCl    = 0.0_dp                   !  cloud droplets
+#ifdef MODEL_CLASSIC
```
lizziel, line 653: Having this here in the MODEL_GCHPCTM block does not make sense.

GeosCore/fullchem_mod.F90
```diff
@@ -389,131 +455,1262 @@ SUBROUTINE Do_FullChem( Input_Opt,  State_Chm, State_Diag,                 &
...
+                      State_Diag%SatDiagnJvalO3O3P(I,J,L) + PHOTOL(N)
+                   ENDIF
+
+                ENDIF
+             ENDDO
+          ENDIF
+
+       ENDIF
+
+#if defined( MODEL_CESM )
```
lizziel, line 831: Why is this in the MODEL_GCHPCTM block?

GeosCore/fullchem_mod.F90
```diff
@@ -389,131 +455,1262 @@ SUBROUTINE Do_FullChem( Input_Opt,  State_Chm, State_Diag,                 &
...
-          CALL fullchem_AR_KeepHalogensActive( Input_Opt%amIRoot )
-       ENDIF
+    IF ( Failed2x ) THEN
+       ErrMsg = 'KPP failed to converge after 2 iterations!'
+       CALL GC_Error( ErrMsg, RC, ThisLoc )
+       RETURN
     ENDIF
-#endif
+
+#else
```
lizziel: The rest of this seems to be the original code, and the above is a duplicate of it with some modifications. We need to better put these two implementations together to avoid so much duplicate code.

GeosCore/fullchem_mod.F90
```diff
@@ -1549,16 +2746,7 @@ SUBROUTINE Do_FullChem( Input_Opt,  State_Chm, State_Diag,                 &
     ENDDO
     ENDDO
     !$OMP END PARALLEL DO
-
-    !=======================================================================
-    ! Return gracefully if integration failed 2x anywhere
-    ! (as we cannot break out of a parallel DO loop!)
```
lizziel, line 1555: Why are the failure exits removed?

GeosCore/fullchem_mod.F90
```diff
@@ -3040,6 +4242,316 @@ SUBROUTINE Init_FullChem( Input_Opt, State_Chm, State_Diag, RC )
        RETURN
     ENDIF
 
+#ifdef MODEL_GCHP
```
lizziel: I would prefer this implementation to be a compile-time option so that we can turn it off for testing.

lizziel (changes requested):

GeosCore/fullchem_mod.F90
```diff
@@ -17,6 +17,11 @@ MODULE FullChem_Mod
 !
   USE Precision_Mod
 
+#ifdef MODEL_GCHP
+  USE MPI
+  USE, INTRINSIC :: ISO_C_BINDING
+#endif
```
yantosca, line 23: As @lizzlel stated we should change this to a separate #ifdef so that we can turn it off for testing.  Maybe:
```F90
#ifdef MODEL_GCHP
#ifdef MPI_LOAD_BALANCE
... etc ...

#endif
#endif
```
@lizziel, any thoughts on the name of the C-preprocessor switch?

GeosCore/fullchem_mod.F90
```diff
@@ -83,6 +89,51 @@ MODULE FullChem_Mod
   REAL(f4), ALLOCATABLE :: JvSumDay  (:,:,:,:)
   REAL(f4), ALLOCATABLE :: JvSumMon  (:,:,:,:)
 
+#ifdef MODEL_GCHP
```
yantosca: Same comment as above re: `#ifdef MPI_LOAD_BALANCE`

GeosCore/fullchem_mod.F90
```diff
@@ -236,20 +294,27 @@ SUBROUTINE Do_FullChem( Input_Opt,  State_Chm, State_Diag,                 &
...
+    !   -- Bob Yantosca (05 May 2022)
     INTEGER,     PARAMETER :: Nhnew = 3
 
     ! Add Nhexit, the last timestep length -- Obin Sturm (30 April 2024)
     INTEGER,     PARAMETER :: Nhexit = 2
 
     ! Suppress printing out KPP error messages after this many errors occur
     INTEGER,     PARAMETER :: INTEGRATE_FAIL_TOGGLE = 20
+   
+#ifdef MODEL_GCHP
```
yantosca: Same comment as above re: `#ifdef MPI_LOAD_BALANCE`

GeosCore/fullchem_mod.F90
```diff
@@ -389,131 +455,1262 @@ SUBROUTINE Do_FullChem( Input_Opt,  State_Chm, State_Diag,                 &
...
+    ! (bmy, 3/28/16)
+    !
+    ! ICNTRL and RCNTRL now need to be updated within the TimeLoop for
+    ! the Rosenbrock_AutoReduce solver (hplin, 4/13/22)
+    !========================================================================
+    DT   = GET_TS_CHEM()          ! Chemistry timestep [s]
+    ATOL = State_Chm%KPP_AbsTol   ! Absolute tolerance
+    RTOL = State_Chm%KPP_RelTol   ! Relative tolerance
+
+#ifdef MODEL_GCHP
```
yantosca: Same comment above re: `#ifdef MPI_LOAD_BALANCE`

GeosCore/fullchem_mod.F90
yantosca: We need the main fullchem_mod loop for all instances of GEOS-Chem

yantosca (changes requested):

GeosCore/fullchem_mod.F90
Jordan-Sun: This shouldn't be removed, and is added back.

Daisy0419:
Hi Bob and Lizziel,

I hope you’re doing well. I’d like to give you a quick update and follow up on our current tasks.

In the latest commit, the modified code for load balancing has been wrapped in the MPI_LOAD_BALANCE macro. It will only be activated when compiled with the MPI_LOAD_BALANCE flag. By default, the behavior remains unchanged and results remain consistent across models (tested on both GCHP with/without load balancing and Classic).

Given the current state of the implementation, could you please advise on what would be most helpful for us to further optimize?

Thanks!

GeosCore/fullchem_mod.F90
```diff
@@ -35,6 +40,7 @@ MODULE FullChem_Mod
 !
 ! !PRIVATE TYPES:
 !
+! #define BALANCE_DEBUG
```
lizziel:
```suggestion
```

GeosCore/fullchem_mod.F90
```diff
@@ -204,6 +253,12 @@ SUBROUTINE Do_FullChem( Input_Opt,  State_Chm, State_Diag,                 &
     REAL(dp)               :: C_before_integrate(NSPEC)
     REAL(dp)               :: local_RCONST(NREACT)
 
+    ! Local copy of all necessary KPP inputs
+    INTEGER                :: IJL_to_Idx(State_Grid%NX, State_Grid%NY, State_Grid%NZ)
+
+    ! All the KPP inputs remapped to a 1-D array
+    INTEGER                :: NCELL, NCELL_local, I_CELL
+
```
lizziel:
```suggestion
#if defined(MODEL_GCHP) && defined(MPI_LOAD_BALANCE)
    ! Local copy of all necessary KPP inputs
    INTEGER                :: IJL_to_Idx(State_Grid%NX, State_Grid%NY, State_Grid%NZ)

    ! All the KPP inputs remapped to a 1-D array
    INTEGER                :: NCELL, NCELL_local, I_CELL
#endif
```

GeosCore/fullchem_mod.F90
```diff
@@ -3040,6 +3734,290 @@ SUBROUTINE Init_FullChem( Input_Opt, State_Chm, State_Diag, RC )
        RETURN
     ENDIF
 
+#if defined(MODEL_GCHP) && defined(MPI_LOAD_BALANCE)
+   !setup shared mamory
+   ! Initialize window handles
```
lizziel:
```suggestion
    ! Setup shared memory
    ! Initialize window handles
```

GeosCore/fullchem_mod.F90
```diff
@@ -2729,19 +3412,22 @@ SUBROUTINE Init_FullChem( Input_Opt, State_Chm, State_Diag, RC )
...
-    USE Gckpp_Parameters,         ONLY : nFam, nReact
+    USE Gckpp_Parameters,         ONLY : nFam, nReact, nSpec
     USE Gckpp_Global,             ONLY : Henry_K0, Henry_CR, MW, SR_MW
     USE Input_Opt_Mod,            ONLY : OptInput
     USE KppSa_Interface_Mod,      ONLY : KppSa_Config
     USE State_Chm_Mod,            ONLY : ChmState
     USE State_Chm_Mod,            ONLY : Ind_
     USE State_Diag_Mod,           ONLY : DgnState
+    USE State_Grid_Mod,           ONLY : GrdState
+    USE Timers_Mod
```
lizziel: What is used from Timers_Mod?

GeosCore/fullchem_mod.F90
```diff
@@ -17,6 +17,11 @@ MODULE FullChem_Mod
 !
   USE Precision_Mod
 
+#if defined(MODEL_GCHP) && defined(MPI_LOAD_BALANCE)
```
lizziel: Why is MODEL_GCHP a pre-requisite for the load balancing? I would think any MPI model could use it.

lizziel (changes requested):
Hi @Daisy0419 and @Jordan-Sun, could you look over the suggested changes and commit them (you can do that from the github page). I also added some comments with questions. Once complete I'll do another review.

GeosCore/fullchem_mod.F90
Daisy0419: Since the load-balancing implementation is in GEOScore but relies on MPI-3 shared memory and assumes GCHP’s distributed-memory decomposition, it is only valid when building with MODEL_GCHP. Other configurations (e.g., Classic) do not have the same distributed-memory decomposition (Classic is OpenMP/shared-memory–oriented), so enabling it there would be unsupported.

GeosCore/fullchem_mod.F90
Daisy0419: I remembered it's just a module we used for test/debug that we should remove it @Jordan-Sun can you double confirm?

Jordan-Sun: I don't think Timers_Mod is used anywhere in the current code. It might have been previously added to measure the initialization overhead but not removed.

GeosCore/fullchem_mod.F90
lizziel: I was actually thinking about chemistry in GEOS and CESM which do use distributed-memory decomposition.

lizziel: Also WRF-GC

GeosCore/fullchem_mod.F90
lizziel: Hi @Jordan-Sun, could you push an update removing this line?

Daisy0419: The line "USE Timers_Mod" has been removed.

GeosCore/fullchem_mod.F90
Daisy0419: "defined(MODEL_GCHP)" has been removed for load-balance code.

lizziel:
I updated the target to 14.7.1.

lizziel (approved):

yantosca (approved):

lizziel:
Hi @Daisy0419, I rebased your PR on the latest version of GEOS-Chem so that we can merge it in. Let me know if it is okay that I push it to your fork.

Jordan-Sun:
Hi @lizziel, feel free to go ahead.

GeosCore/fullchem_mod.F90
```diff
@@ -3040,6 +3733,290 @@ SUBROUTINE Init_FullChem( Input_Opt, State_Chm, State_Diag, RC )
...
+   CALL GC_CheckVar( 'fullchem_mod.F90:Idx_to_IJL', 0, RC )
+   IF ( RC /= GC_SUCCESS ) Then
+      CALL GC_Error( 'Failed to allocate Idx_to_IJL', RC, ThisLoc )
+      RETURN
+   END IF
+
+
+   !gather local sizes
+    IF (shm_rank == 0) ALLOCATE(sizes(shm_size))
+    CALL MPI_Gather(NCELL_local, 1, MPI_INTEGER, sizes, 1, MPI_INTEGER,        &
```
lizziel: Hi @Daisy0419 and @Jordan-Sun, is the call to the MPI_Gather here supposed to be only if shm_rank is 0? I ran into an errror where 'sizes' was not allocated for the MPI_Gather call.

GeosCore/fullchem_mod.F90
```diff
@@ -1286,6 +1871,35 @@ SUBROUTINE Do_FullChem( Input_Opt,  State_Chm, State_Diag,                 &
...
+    ENDIF
+
+    DO L = 1, State_Grid%NZ
+    DO J = 1, State_Grid%NY
+    DO I = 1, State_Grid%NX
+       ! Figure out which cell the data should be allocated to
+       N       = IJL_to_Idx(I,J,L)
+      !  WRITE (6,'(A,I4,A,3(I6,1X),A,I8)') 'shm_rank=', shm_rank, '  I,J,L=', I, J, L, '  NCELL_local=', N
+
+       If (N.le.0) Cycle
```
lizziel: Is there ever a case where negative N is okay? I would think all cases of (I,J,L) should map to a cell.

GeosCore/fullchem_mod.F90
```diff
@@ -1098,47 +1193,586 @@ SUBROUTINE Do_FullChem( Input_Opt,  State_Chm, State_Diag,                 &
...
-          ! Update autoreduce solver statistics
-          ! (only if the autoreduction is turned on)
-          IF ( Input_Opt%Use_AutoReduce ) THEN
-             CALL fullchem_AR_UpdateKppDiags( I, J, L, RSTATE, State_Diag )
+       ! Print grid box indices to screen if integrate failed
+       IF ( IERR < 0 ) THEN
+
+          ! Turn off error output after a certain limit is reached
+          IF ( .not. doSuppress ) THEN
+             WRITE( 6, * ) '### INTEGRATE RETURNED ERROR AT: ', I, J, L
```
lizziel: I think this I,J,L should actually be I_CELL. I can change that.

GeosCore/fullchem_mod.F90
```diff
@@ -1235,10 +1821,10 @@ SUBROUTINE Do_FullChem( Input_Opt,  State_Chm, State_Diag,                 &
                 C = C_before_integrate
              ENDIF
 
-             ! Keep track of error boxes
-             IF ( State_Diag%Archive_KppError ) THEN
-                State_Diag%KppError(I,J,L) = State_Diag%KppError(I,J,L) + 1.0
-             ENDIF
+             !! Keep track of error boxes
```
lizziel, line 1875: It looks like this diagnostic was never ported to the new implementation. I'll add a TODO comment if that is the case.

GeosCore/fullchem_mod.F90
```diff
@@ -1260,7 +1846,7 @@ SUBROUTINE Do_FullChem( Input_Opt,  State_Chm, State_Diag,                 &
              ! Print rate constants at failure grid box
              PRINT*, REPEAT( '#', 79 )
              PRINT*, '### KPP DEBUG OUTPUT!'
-             PRINT*, '### Reaction rates at problem box ', I, J, L
+             PRINT*, '### Reaction rates'! at problem box ', I, J, L
```
lizziel: This also should be I_CELL since no I,J,L defined in this loop

GeosCore/fullchem_mod.F90
lizziel: This call will not work since several of the arguments are not ported to the new implementation. I think the quick way forward is have KPP standalone only for runs without the load balancing. I will comment out with a TODO.

lizziel (commented):
Hi @Jordan-Sun and @Daisy0419, please take a look at my latest comments and let me know if you agree.

GeosCore/fullchem_mod.F90
Daisy0419: MPI_Gather is a collective on shm_comm, so it should be called by all ranks and 'sizes' is ignored by non-zero ranks

GeosCore/fullchem_mod.F90
Daisy0419: I don't think N would be negative.

lizziel: Okay, I will remove this line. Either we leave it and include an error trap, or remove it if it will simply never be true.

GeosCore/fullchem_mod.F90
lizziel: The problem is when using GCC 12 compiler with debug flags on there is a run-time error that sizes is not allocated. I fixed this by allocating to sizes(1) if shm_rank not equal to 0. This is the same thing you did further down in the code for prefix.

lizziel:
This PR is now ready to go in. Integration tests pass (debug flags on issue fixed) and are zero diff with respect to the no-diff-to-benchmark branch.
@yantosca

yantosca:
All GEOS-Chem Classic integration tests passed:
```console

CodeDir   : 2b7e906 GEOS-Chem update: Merge PR #3015 (MPI load balancing)
GEOS-Chem : 98076d36d Merge PR #3015 (Load-Balancing Solutions for Chemistry Module)
HEMCO     : e23c43b HEMCO 3.12.0 release
Cloud-J   : 1dff6fe Merge pull request #31 from geoschem/release/8.0.2
HETP      : 2a99b24 Merge pull request #2 from geoschem/bugfix/initialize_local_variables

Using 24 OpenMP threads
Number of execution tests: 38

Submitted as SLURM job: 65227253

```
All GEOS-Chem Classic tests were zero-diff w/r/t `gcc.2.c9f84cd`, except

- gc_4x5_merra2_fullchem_APM
  - Parallel error (known issue)

All GCHP integration tests passed:
```console
GCHP: Execution Test Results

CodeDir       : c39f403 Merge PR #532 Compiler option for GCHP load balance in chemistry)
MAPL          : d1d528e5 Merge pull request #39 from geoschem/feature/mapl_2.55.0
GMAO_Shared   : 4ddb3ec Merge pull request #2 from geoschem/feature/mapl-upgrade
ESMA_cmake    : 01bbf2f Merge pull request #9 from geoschem/feature/turn_off_f2py_builds_in_gchp
FMS           : 259759d1 Merge pull request #3 from geoschem/feature/update_gmao_libs
FVdycoreCubed : 348620f Merge PR #11 (Scale tracers by pressure ratio in first timestep)
geos-chem     : 98076d36d Merge PR #3015 (Load-Balancing Solutions for Chemistry Module)
HEMCO         : e23c43b HEMCO 3.12.0 release
Cloud-J       : 1dff6fe Merge pull request #31 from geoschem/release/8.0.2
HETP          : 2a99b24 Merge pull request #2 from geoschem/bugfix/initialize_local_variables
GFE           : bac529f Merge pull request #40 from Goddard-Fortran-Ecosystem/feature/mathomp4/update-gftl-1.15.2

Number of execution tests: 12

Submitted as SLURM job: 65255444

```
All GCHP tests were zero-diff w/r/t gchp.2.97750f2.

PR #3016, opened 2025-09-05
Update mass flux and Courant number scalings for changes to MAPL

lizziel:

Name: Lizzie Lundgren

### Describe the update

This PR updates the GCHP mass flux and Courant number import scaling calculation done in `setCommonRunSettings.sh` for use in `ExtData.rc`. It only impacts GCHP runs that import mass flux rather than use winds in advection.

If using mass flux and Courant number imports from the meteorology archive then these imports must be scaled using factors dependent on run timestep, run grid resolution, timestep used to produce the met archive, and grid resolution used to produce the met archive. These scale factors were introduced in 14.6.3 but must be changed again in 14.7.0 given changes in MAPL horizontal flux regridding in that version. The new scaling is the same as the old but divided by an additional quantity of (met grid resolution / run grid resolution). This results in cancelling out the grid-dependency for mass flux and squaring the grid dependency of courant numbers.

I validated the changes by comparing to mass flux and Courant number diagnostics of GCHP runs that use 3-hr wind imports versus GCHP runs that use 1-hr mass fluxes. Both runs export mass fluxes and Courant numbers of advection and we expect those exports to match, meaning we assume the values computed from winds and pressures are correct. The equation I use for scaling factor works very well for mass fluxes and well enough for Courant numbers. I tested at both c30 and c90 to validate that accuracy does not change with run grid resolution.

### Expected changes

This is a no diff update for GC-Classic and for GCHP benchmarks. Very small differences may be seen in GCHP runs that use mass flux imports.

### Reference(s)

None

### Related Github Issue

This update prevents the problem in this old issue from reappearing in 14.7.0: https://github.com/geoschem/GCHP/issues/503


lizziel:
Validation plots

**c30 Courant number, scale factor 0.037037**

CX c30

**c30 mass flux, scale factor 1.3333333**

MFX c30

**c90 Courant number, scale factor 0.3333333**

CX c90

**c90 mass flux, scale factor 1.3333333**

MFX c90

lizziel:
Fyi @yuanjianz

yantosca (approved):
Looks good!  But I will wait until @yuanjianz has a chance to comment on the PR before I bring it into 14.7.0.

yuanjianz:
@lizziel @yantosca 
This looks good to me. Also pointing to https://github.com/geoschem/GCHP/issues/503#issuecomment-3133754106 in case interested ones see this thread. In summary, the empirical scaling for MFX/Y is more mathematically sounding while for Courant number, some uncertainties might appear at the edge of each cubed-sphere face.

PR #3050, opened 2025-09-26
Fix precision of GCHP import of area from advection

lizziel:

Name: Lizzie Lundgren

### Describe the update

This update fixes a bug in GCHP where area was stored in the State_Grid object as REAL8 but was imported from FV3 advection as REAL4. The area computed in FV3 is REAL8 so there is no reason to truncate to REAL4 for the import.

### Expected changes

This update will cause small changes in GCHP simulations due to the precision change in area. Area is used in unit conversions within the model. This update has no impact on GEOS-Chem Classic.

### Reference(s)

None

### Related Github Issue

Please link to the corresponding Github issue(s) here. If fixing a bug, there should be an issue describing it with steps to reproduce.


yantosca (approved):
Thanks @lizziel!  Good to merge.  I can get this into the 14.7.0 branch.

PR #3066, opened 2025-10-07
Toggle "read_restart_as_real8" from false to true for fullchem benchmark simulations

yantosca:

Name: Bob Yantosca
Institution: Harvard + GCST

### Describe the update

This PR adds a call to bash function `replace_colon_sep_val` in `run/shared/setupConfigFiles.sh` to change:
```yaml
  read_restart_as_real8: false
```
to
```yaml
  read_restart_as_real8: true
```
in `geoschem_config.yml` files for fullchem benchmark simulations. 

Also, we set `read_restart_as_real8: true` by default in the `run/shared/geoschem_config.yml.templates/geoschem_config.yml.TransportTracers` template file, in order to ensure mass conservation in 1-year TransportTracers benchmarks.

### Expected changes
This will cause numerical noise differences in fullchem benchmark simulations that are split up into several run stages, as there will no longer be any `REAL*8` -> `REAL*4` precision truncation.

### DIFF
--- a/run/shared/setupConfigFiles.sh
+++ b/run/shared/setupConfigFiles.sh
@@ -136,6 +136,11 @@ function set_common_settings() {
 
 	# Remove the first comment character on diagnostics
         sed_ie "s|#'|'|"                               HISTORY.rc
+
+	# Make sure that benchmark simulations read the restart file via
+	# GEOS-Chem and not HEMCO.  This will ensure mass conservation
+	# when the run is broken up into several stages.
+	replace_colon_sep_val "read_restart_as_real8" "true" geoschem_config.yml
     fi
 
     #------------------------------------------------------------------------
--- a/run/shared/setupConfigFiles.sh
+++ b/run/shared/setupConfigFiles.sh
@@ -120,9 +120,6 @@ function set_common_settings() {
 	    # Change time cycle flag to allow missing species
 	    sed_ie 's|EFYO|CYS|' HEMCO_Config.rc
 
-	    # Remove the first comment character on diagnostics
-            sed_ie "s|#'|'|"     HISTORY.rc
-
 	    # Make sure that GC-Classic benchmark simulations read the
 	    # restart file via GEOS-Chem and not HEMCO.  This will ensure
 	    # mass conservation when the run is broken up into several
@@ -149,7 +146,10 @@ function set_common_settings() {
 	# Turn @ into # characters for the benchmark simulation,
 	# which should cause MAPL to skip reading these lines.
 	# This is a workaround for a "input file to long" MAPL error.
-	sed_ie 's|@|#|'                                HISTORY.rc
+	sed_ie 's|@|#|'                               HISTORY.rc
+
+	# Remove the first comment character on diagnostics
+        sed_ie "s|#'|'|"                              HISTORY.rc
     fi
 
     #------------------------------------------------------------------------
--- a/run/shared/setupConfigFiles.sh
+++ b/run/shared/setupConfigFiles.sh
@@ -112,11 +112,28 @@ function set_common_settings() {
     #------------------------------------------------------------------------
     if [[ "x${sim_extra_option}" == "xbenchmark" ]]; then
 
-	# Change time cycle flag to allow missing species (GCClassic only)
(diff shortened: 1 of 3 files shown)


run/shared/setupConfigFiles.sh
```diff
@@ -136,6 +136,11 @@ function set_common_settings() {
 
 	# Remove the first comment character on diagnostics
         sed_ie "s|#'|'|"                               HISTORY.rc
+
```
lizziel: This can be moved to the earlier block for GC-Classic only.

run/shared/setupConfigFiles.sh
```diff
@@ -136,6 +136,11 @@ function set_common_settings() {
...
+	# Make sure that benchmark simulations read the restart file via
+	# GEOS-Chem and not HEMCO.  This will ensure mass conservation
+	# when the run is broken up into several stages.
+	replace_colon_sep_val "read_restart_as_real8" "true" geoschem_config.yml
```
lizziel: Could this also be done for transport tracers by default? I don't think anyone does nested transport tracer simulations.

run/shared/setupConfigFiles.sh
```diff
@@ -136,6 +136,11 @@ function set_common_settings() {
 
 	# Remove the first comment character on diagnostics
         sed_ie "s|#'|'|"                               HISTORY.rc
+
+	# Make sure that benchmark simulations read the restart file via
+	# GEOS-Chem and not HEMCO.  This will ensure mass conservation
+	# when the run is broken up into several stages.
```
lizziel: Expand the commend to say "since restart file data is stored as REAL8 but HEMCO reads file data as REAL4"

run/shared/setupConfigFiles.sh
yantosca: @lizziel: In that case, would it be simpler to set `read_restart_as_real8: true` in the template file for TransportTracers?

run/shared/setupConfigFiles.sh
yantosca: Now fixed in https://github.com/geoschem/geos-chem/pull/3066/commits/2b73ee1e29aa95a2b0ae99e460bbc110b0d0a145

run/shared/setupConfigFiles.sh
yantosca: Now fixed in https://github.com/geoschem/geos-chem/pull/3066/commits/2b73ee1e29aa95a2b0ae99e460bbc110b0d0a145.  

I've also set `read_restart_as_real8: true` in `geoschem_config.yml.TransportTracers` by default.

run/shared/setupConfigFiles.sh
yantosca: Now done in https://github.com/geoschem/geos-chem/pull/3066/commits/2b73ee1e29aa95a2b0ae99e460bbc110b0d0a145

run/shared/setupConfigFiles.sh
```diff
@@ -112,11 +112,28 @@ function set_common_settings() {
...
-	# Change time cycle flag to allow missing species (GCClassic only)
+	#--------------------------------
+	# Updates for GC-Classic only
+	#--------------------------------
 	if [[ "x${model}" == "xGCClassic" ]]; then
+
+	    # Change time cycle flag to allow missing species
 	    sed_ie 's|EFYO|CYS|' HEMCO_Config.rc
+
+	    # Remove the first comment character on diagnostics
```
lizziel: Don't we also need this for GCHP?

yantosca: GCHP uses ExtData and not HEMCO though to read the input

yantosca: @lizziel: I moved it back to the original location in https://github.com/geoschem/geos-chem/pull/3066/commits/31f21c5bd5410bdbfa242b4fc628f98c9f3405bd. 

Sorry, I think I misunderstood the comment before so I moved it to the GC-Classic only if block.  Now fixed though.

lizziel: Thanks!

PR #3074, opened 2025-10-15
Fix typos causing carbon simulations with the Carbon collection enabled to crash

msulprizio:
### Name and Institution (Required)

Name: Melissa Sulprizio
Institution: Harvard

### Describe the update

When enabling the Carbon collection in HISTORY.rc, GEOS-Chem crashes with an out-of-bounds error in KPP/carbon/carbon_Funcs.F90. This can be tracked to the uses of incorrect ind_ variable names -- ind_COfromCH4 should be ind_PCOfromCH4 and ind_COfromNMVOC should be ind_PCOfromNMVOC to reflect the species name change introduced in pull request #2800.

We now also turn on the Carbon collection in HISTORY.rc by default to catch potential bugs like this during integration tests.

### Expected changes

This is a zero-difference update with respect to the full-chemistry benchmark simulation.

PR #3095, opened 2025-10-28
Do not reference OHdiurnalFac in carbon_gases_mod.F90 for carbon simulations where OH is not used

yantosca:

Name: Bob Yantosca
Institution: Harvard + GCST

### Describe the update
This PR fixes an out-of-bounds error that can occur in the CO2-only and OCS-only carbon simulations when the State_Diag%OHconcAfterChem diagnostic is selected to be archived, but when the OHdiurnalFac array is not allocated.  To fix the error we enclose the IF block where the diagnostic is updated with in the `IF ( id_CH4_adv > 0 .or. id_CO_adv > 0 )` block.

### Expected changes
This fix will allow the CO2-only and OCS-only carbon simulations to proceed to completion.

### Related Github Issue

- See #3074


msulprizio (approved):
Thanks for the fix @yantosca! Looks good to merge.

PR #3272, opened 2026-04-13
Remove optional valid_range in ExtData2G yaml file for transport trac…

lizziel:

Name: Lizzie Lundgren

### Describe the update

This PR removes the `valid_range` option in the new ExtData2G yaml file for the Transport Tracers simulation introduced in this version as beta in GCHP. The `valid_range` option works fine on the Harvard cluster because the full span of all available emissions and meteorology data is present locally. However, the model crashes on a different cluster with only the data needed for a short run downloaded. This is because the `valid_range` for each data collection in the file is set for the time range of all data available remotely in the GEOS-Chem input data repository rather than just the data needed for the run. ExtData2G crashes when it cannot find all of the data even if it is not actually needed.

From the MAPL ExtData2G user guide:

> valid_range - character string of form "IOS time 1/ISO time 2" specify the valid range of times for the dataset. You are telling the application that you should be able to find a valid file on disk by applying any time between that range to the template (within the constraints of the reference time and reference frequency of course). This does not mean you have to use all the data, this is simply telling you what is available. This is only needed if you need to perform some option that extrapolates outside of the range of the data. Right now this can be detected in limited cases (no tokens in the template), detection in general is yet to be implemented. Obviously could be expensive as there is really no way but brute forcing this without help from the user.

No update to the Changelog is needed because the yaml file has not yet been released.

### Expected changes

None. This is a no diff update.

### Reference(s)

https://github.com/GEOS-ESM/MAPL/wiki/ExtData-Next-Generation---User-Guide

### Related Github Issue

https://github.com/geoschem/GCHP/pull/537

### DIFF
--- a/run/GCHP/ExtData2G.yaml.templates/extdata.yaml.TransportTracers
+++ b/run/GCHP/ExtData2G.yaml.templates/extdata.yaml.TransportTracers
@@ -5,22 +5,16 @@ Collections:
   # Meteorology collections
   #-------------------------------------------------------------------
   MERRA2.A3dyn.05x0625.nc4:
-    valid_range: "2009-01-01/2024-12-31T00:00"
     template: ./MetDir/%y4/%m2/MERRA2.%y4%m2%d2.A3dyn.05x0625.nc4
   MERRA2.I3.05x0625.nc4:
-    valid_range: "2009-01-01/2024-12-31T00:00"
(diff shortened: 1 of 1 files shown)


yantosca (approved):
Thanks @lizziel!  Looks good to merge!

mcdon1718-h (approved):
looks good!

lizziel:
This PR accidentally introduced a problem. We will fix this in 14.8.0. See https://github.com/geoschem/geos-chem/issues/3279.

PR #3364, opened 2026-07-27
Util script checkYamlSettings.sh to check 2G-compatible yamls for file setting errors before run time

mcdon1718-h:

Name: Helena McDonald

### Describe the update

Adding checkYamlSettings.sh to GCHP utils folder. File searches local machine for input data files based on run start and end dates and the file read frequency specified by the yaml. Also flags if run duration is outside valid_range for any collection. Useful for identifying missing files or incorrect data read specifications before runs are submitted and fail. Helpful for transitioning users from ExtData 1G to 2G config files.

### Expected changes

None.

### DIFF
--- a/run/GCHP/utils/checkYamlSettings.sh
+++ b/run/GCHP/utils/checkYamlSettings.sh
@@ -9,8 +9,12 @@ set -euo pipefail
 # SUMMARY: [freq] collection_path: X / Y files missing (start_date to end)
 #
 # [freq] is the frequency files are expected in the provided folder, parsed from 
-# the smallest date token (eg %d2) in the template file path. 
-# note this is NOT the same as the sampling frequency as defined in yaml samplings. 
+# the smallest date token (%y4/%m2/%d2) in the template file path. 
+# Note: this is NOT the same as the sampling frequency as defined in yaml samplings. 
+#
+# Note: %h2 tokens are not parsed as no default collections use hourly file frequency; 
+# hourly input data is bundled into daily-or-greater files. Testing hourly files
+# will not work with this script.
 # 
 # collection_path is the path where files are expected to be found, derived from 
 # the collection template in the extdata.yaml.
--- a/run/GCHP/utils/checkYamlSettings.sh
+++ b/run/GCHP/utils/checkYamlSettings.sh
@@ -26,6 +26,8 @@ set -euo pipefail
 # This setting determines whether to check for files on disk within a provided valid range
 # for a given collection. This can be slow, especially for daily collections like MERRA2, 
 # so it is disabled by default. 
+# This is useful because valid_range currently fails if any files in valid range are missing,
+# regardless if they are within simulation date range or not. 
 CHECK_VALID_RANGE=false
 
 # Lightweight check to determine if the provided simulation date is within the valid_range 
@@ -43,7 +45,7 @@ SIM_START_DATE=20190701
 SIM_END_DATE=20190801
 
 # Path to your YAML file relative to this script
-YAML=extdata.yaml          
+YAML=extdata.yaml        
 
 ##########################################################
 # Derived config
@@ -51,6 +53,19 @@ YAML=extdata.yaml
 
 SIM_START_ISO="${SIM_START_DATE:0:4}-${SIM_START_DATE:4:2}-${SIM_START_DATE:6:2}"
 SIM_END_ISO="${SIM_END_DATE:0:4}-${SIM_END_DATE:4:2}-${SIM_END_DATE:6:2}"
+
+##########################################################
+# Early failure checks
+##########################################################
+if [[ ! -f ${YAML} ]]; then
+   echo "Could not find configuration file: ${YAML}!  Exiting ..."
+   exit 1
+fi
+
+if (( $(awk '/^[[:space:]][[:space:]][^[:space:]]+:[[:space:]]*(#.*)?$/' "${YAML}" | wc -l) == 0 )); then
+   echo "ERROR: no collections parsed from ${YAML} — check indentation/format" >&2
+   exit 1
+fi
 
 #------------------------
 # Config info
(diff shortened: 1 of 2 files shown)


run/GCHP/utils/checkYamlSettings.sh
```diff
@@ -0,0 +1,442 @@
...
+            #     echo "SUMMARY [constant] $name: constant file missing ($path)"
+            # fi
+            ;;
+
+        monthly)
+            local d
+            d=$(date -d "$(date -d "$d_start" +%Y-%m-01)" +%F)
+            if [[ "$d" < "$d_start" ]]; then
+                d=$(date -d "$d +1 month" +%F)
+            fi
```
yantosca: @mcdon1718-h: Have you checked simulation dates that don't start at the beginning of the month?  Claude AI seems to think that this may be problematic.

**AI disclosure** 
Monthly-frequency check skips the simulation's start month whenever it doesn't begin on the 1st
For `d_start=2019-07-15`, this rounds to `2019-07-01`, sees it's earlier than `d_start`, and
jumps to `2019-08-01` — so the July monthly file is never checked at all, even though the
simulation needs it. 

Fix: drop the `if` block entirely — `d` should just stay at the first-of-month value so that
month is always included (mirrors how the `annual` case already handles partial years correctly).

run/GCHP/utils/checkYamlSettings.sh
```diff
@@ -0,0 +1,442 @@
...
+}
+
+/^[[:space:]][[:space:]][[:space:]][[:space:]]valid_range:[[:space:]]*/ {
+    line = $0
+    sub(/^[[:space:]][[:space:]][[:space:]][[:space:]]valid_range:[[:space:]]*/, "", line)
+    gsub(/^"[[:space:]]*|"[[:space:]]*$/, "", line)
+    gsub(/^'[[:space:]]*|'[[:space:]]*$/, "", line)
+
+    n = split(line, a, "/")
+    if (n >= 1) start_date = a[1]
```
yantosca: **AI disclosure**
Recommend changing this to:
```bash
if (n >= 1) {
    start_date = a[1]
    sub(/T.*/, "", start_date)
}
```
This will prevent false "sim precedes valid range" warnings.

run/GCHP/utils/checkYamlSettings.sh
```diff
@@ -0,0 +1,442 @@
...
+    line = $0
+    sub(/^[[:space:]][[:space:]][[:space:]][[:space:]]valid_range:[[:space:]]*/, "", line)
+
+    # strip surrounding quotes if present
+    gsub(/^"[[:space:]]*|"[[:space:]]*$/, "", line)
+    gsub(/^'[[:space:]]*|'[[:space:]]*$/, "", line)
+
+    # split "start/end"
+    n = split(line, a, "/")
+    if (n >= 1) start_date = a[1]
```
yantosca: **AI disclosure**
Recommend changing this to:
```bash
if (n >= 1) {
    start_date = a[1]
    sub(/T.*/, "", start_date)
}
```
This will prevent false "sim precedes valid range" warnings.

run/GCHP/utils/checkYamlSettings.sh
```diff
@@ -0,0 +1,442 @@
...
+            ;;
+
+        daily|hourly|*)
+            local d="$d_start"
+            while [[ "$d" < "$d_end" || "$d" == "$d_end" ]]; do
+                local y4 m2 d2 path
+                y4=$(date -d "$d" +%Y)
+                m2=$(date -d "$d" +%m)
+                d2=$(date -d "$d" +%d)
+                path=${template//%y4/$y4}
```
yantosca, line 328: **AI disclosure**
 `%h2` (hour) template tokens are never substituted.  If any collection in the checked yaml uses `%h2`, the literal string `%h2...` is left in the constructed path, so `-f` will essentially never match and the file gets reported as permanently missing (false positive), or in rare cases could coincidentally match an unrelated literally-named file. `infer_freq_from_template` also has no notion of hourly resolution, so there's no code path that even attempts to loop over hours.

Fix: detect `%h2` in `infer_freq_from_template` and either substitute real hour values in the daily loop, or explicitly print "hourly collections not supported, skipping" rather than silently reporting misleading missing-file counts.

run/GCHP/utils/checkYamlSettings.sh
```diff
@@ -0,0 +1,442 @@
...
+# When true, print file availability summary for all collections, 
+# even those with 0 missing files 
+DEBUG_ALL=false           
+
+# Format: YYYYMMDD
+SIM_START_DATE=20190701   
+SIM_END_DATE=20190801
+
+# Path to your YAML file relative to this script
+YAML=extdata.yaml
```
yantosca: Consider putting a check to make sure the file exists, to protect against user-introduced typos.
```bash
if [[ ! -f ${YAML} ]]; then
   echo "Could not find configuration file: ${YAML}!  Exiting ..."
}
```

run/GCHP/utils/checkYamlSettings.sh
```diff
@@ -0,0 +1,442 @@
...
+
+########################################
+# AWK: parse collections
+# Outputs: collection_name<TAB>start<TAB>end<TAB>template
+########################################
+
+AWK_PARSE_COLLECTIONS=$(cat << 'EOF'
+# 2-space-indented collection name, e.g.
+#   "  MERRA2.20150101.CN.05x0625.nc4:"
+/^[[:space:]][[:space:]][^[:space:]]+:[[:space:]]*$/ {
```
yantosca: **AI disclosure**
 Collection-name/key regexes assume rigid 2-space/4-space indentation with no trailing comments `^[[:space:]][[:space:]][^[:space:]]+:[[:space:]]*$` silently fails to match (and thus silently skips that collection with no warning) if a yaml author uses different indentation or appends an inline `# comment` after the collection name. Matches the yaml files currently in the repo, but brittle.

Fix: Consider a sanity check after the awk pass (e.g. "WARNING: no collections parsed from $YAML — check indentation") so a parsing failure doesn't look like a clean, zero-issue run.

yantosca (changes requested):
Thanks for this @mcdon1718-h.  I parsed this through Claude Code, which came up with some fixes and suggestions (see comments).  Mostly these are to prevent edge cases in the parsing.

Also we would need a CHANGELOG.md update.

lizziel:
I will review this once @yantosca's comments are addressed. Also update the name of the PR to prevent it from rolling over to the description.

run/GCHP/utils/checkYamlSettings.sh
mcdon1718-h: I didn't think it was worth building in specifically an hourly file checker loop because all of our hourly-sampled data is stored in daily-or-greater files. The only input file that has an %h2 string is the example DynHeating for RRTMG, which is disabled by default and requires customizing from the user before it can be enabled. As such it's sort of out of the purview of this script, which is mostly for configuring the default collections. I can add in hourly parsing for completeness if we think it's necessary.

yantosca: @mcdon1718-h that's a good point.  In this case I think it'd be sufficient just to add a comment stating that we aren't parsing hours but that our data is typically bundled into daily files.

yantosca (approved):
Thanks @mcdon1718-h, this is good to merge!

lizziel:
Thanks @mcdon1718-h! I will give test it out, and also test on AWS and discover to see if I can break it. :)

lizziel:
GMAO has written a python script to do this. @mcdon1718-h, could you take a look to see if it does what yours does? https://github.com/GEOS-ESM/MAPL/pull/5370

PR #3369, opened 2026-08-06
Reduce redundant work in unit conversions, RDAER, wetdep, and Arrhenius rate functions

yuyao-cyber:
## Name and Institution

Name: Yuyao (Bill) Zhuge

## Describe the update

Four small performance cleanups in hot paths, found during a profiling study of 14.7.1
(GC-Classic 4x5 MERRA-2 fullchem, 7-day benchmarks with `use_gcclassic_timers: true`,
24 OpenMP threads, 2 repetitions per configuration on a dedicated node):

1. **`GeosUtil/unitconv_mod.F90`** (`ConvertSpc_KgKgDry_to_Kgm2`, `ConvertSpc_Kgm2_to_KgKgDry`):
   the factor `g0_100 * State_Met%DELP_DRY` is identical for all species but was recomputed
   inside the per-species loop (~nAdvect times per call; these routines run 4x per dynamic
   timestep from mixing and wetdep). It is now computed once into a local `factor` array and
   reused. Bit-for-bit identical.

2. **`GeosCore/aerosol_mod.F90`** (`RDAER`): the relative humidity (including a
   saturation-vapor-pressure `EXP`) and RH-bin index depend only on the grid cell, but were
   recomputed inside the aerosol-type (NAER) and wavelength (NWVS) loops — up to NAER x NWVS
   times per cell per chemistry timestep. A new pre-pass computes `RELHUM3D` and `IRHARR`
   once per cell; the inner loops read them. The `InChemGrid` skip also moved above the
   optics LUT interpolation so non-chemistry boxes skip that work. Bit-for-bit identical.

3. **`GeosCore/wetscav_mod.F90`** (`WetDep`): `DSpc` was declared as a full-grid automatic
   array `(nWetDep, NZ, NX, NY)` despite the "for a given (I,J) column" comment, so it was
   re-allocated on every wetdep timestep (contributing to the `ulimit -s unlimited`
   requirement). It is now `ALLOCATABLE, SAVE`, allocated on first call. Values were already
   re-initialized per column inside the loop.

4. **`KPP/fullchem/rateLawUtilFuncs.F90`** (`GCARR_ac`, `GCARR_abc`): replaced
   `EXP( c0 / TEMP )` with `EXP( c0 * INV_TEMP )`; `INV_TEMP` is already precomputed per
   grid cell (THREADPRIVATE in gckpp_Global). `GCARR_ac` has ~395 call sites in
   `Update_RCONST`, so this removes ~400 FP divisions per cell per chemistry timestep.
   **NOTE: results change in the last ULP — this item is NOT bit-for-bit vs. prior versions**
   (the other three items are). Happy to split it into a separate PR if preferred.

## Expected changes to outputs

Items 1-3: none (bit-for-bit; verified in 7-day benchmarks — transported tracers
BCPI/OCPI/SALA identical to baseline at machine precision).

Item 4: last-ULP perturbation to rate constants; after 7 days this produces chaos-amplified
differences with mean |rel diff| ~3e-5 for OH and smaller for O3/CO. Science-neutral but
will show in diff tests.

## Reference(s)

Performance study on WashU compute1 (LSF, Xeon Gold 6154). Benchmark protocol:
4 configurations x 2 repetitions, GC-Classic built-in timers. Timing effect of this PR alone
is modest (wetdep timer -4%, unit conversion cost roughly halved for the affected pair;
total ~-1%); the primary motivation is removing objectively redundant work in hot loops
without changing science.

## Related Github Issue(s)

(none)


KPP/fullchem/rateLawUtilFuncs.F90
```diff
@@ -57,7 +57,9 @@ FUNCTION GCARR_ac( a0, c0 ) RESULT( k )
     REAL(dp), INTENT(IN) :: a0, c0
     REAL(dp)             :: k
     !
-    k = a0 * EXP( c0 / TEMP )
+    ! Use precomputed INV_TEMP to replace the costly division by TEMP
+    ! with a multiply (result differs at most in the last ULP)
```
lizziel, line 61: I think we should remove the justification for the change from the comments.

KPP/fullchem/rateLawUtilFuncs.F90
```diff
@@ -67,7 +69,9 @@ FUNCTION GCARR_abc( a0, b0, c0 ) RESULT( k )
     REAL(dp), INTENT(IN) :: a0, b0, c0
     REAL(dp)             :: k
     !
-    k = a0 * EXP( c0 / TEMP ) * K300_OVER_TEMP**b0
+    ! Use precomputed INV_TEMP to replace the costly division by TEMP
+    ! with a multiply (result differs at most in the last ULP)
```
lizziel, line 73: Same comment as above. Remove the justification comment. The reasoning for the change will be in the git history.

yantosca:
Thanks @yuyao-cyber.  I probably won't be able to get to this until 14.8.1 though.

CHANGELOG.md
```diff
@@ -20,6 +20,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Added calls to KPP `Initialize` routine from `INIT_FULLCHEM` and `INIT_MERCURY` to specify an absolute tolerance threshold for passive species
 
 ### Changed
+- Hoisted the species-invariant conversion factor (`g0_100 * State_Met%DELP_DRY`) out of the per-species loops in `GeosUtil/unitconv_mod.F90` routines `ConvertSpc_KgKgDry_to_Kgm2` and `ConvertSpc_Kgm2_to_KgKgDry` (bit-for-bit identical, ~2x faster conversions)
+- Precompute relative humidity and RH-bin index once per grid cell in `GeosCore/aerosol_mod.F90` routine `RDAER` instead of once per aerosol type per wavelength; skip non-chemistry grid boxes before the optics LUT interpolation (bit-for-bit identical)
+- Changed the `DSpc` accumulator in `GeosCore/wetscav_mod.F90` routine `WetDep` from a full-grid automatic array (re-allocated every timestep) to an allocate-once `ALLOCATABLE, SAVE` array
+- Use precomputed `INV_TEMP` (multiply) instead of dividing by `TEMP` in `GCARR_ac` and `GCARR_abc` in `KPP/fullchem/rateLawUtilFuncs.F90`. NOTE: this changes results in the last ULP and is therefore not bit-for-bit vs. prior versions
```
lizziel, line 26: Could you cut the length of these entries in half? I also wonder if they were generated by AI? They have an over-complicated feel. Best rewrite to be plainer, e.g. "Changed unit conversion routines to move computation outside of loops where possible". The exact details will be in the git history.

GeosCore/aerosol_mod.F90
```diff
@@ -1505,6 +1506,62 @@ SUBROUTINE RDAER( Input_Opt, State_Chm, State_Diag, State_Grid, State_Met, &
        ENDIF
     ENDIF
 
+    !=======================================================================
+    ! Precompute relative humidity and the RH-bin index once per grid cell.
+    ! These depend only on (I,J,L) but were formerly recomputed inside the
+    ! wavelength (NWVS) and aerosol-type (NAER) loops below, repeating the
+    ! saturation-vapor-pressure EXP up to NWVS*NAER times per cell.
+    !=======================================================================
```
lizziel, line 1514: Remove the explanation of why the change was made. Just keep the first line of the commit. As mentioned in later review comments, we will have too many comments in the code if there are justifications for structural changes.

GeosCore/wetscav_mod.F90
```diff
@@ -3284,10 +3284,14 @@ SUBROUTINE WETDEP( Input_Opt, State_Chm, State_Diag, State_Grid, &
...
-    ! soluble species for a given (I,J) column
-    REAL(fp)               :: DSpc(State_Chm%nWetDep, &
-                                   State_Grid%NZ,State_Grid%NX,State_Grid%NY)
+    ! DSpc is the accumulator array of rained-out soluble species,
+    ! indexed (wetdep species, level, I, J).  It is allocated once on
+    ! the first call and kept for the whole run (it was formerly a
+    ! full-grid automatic array, re-allocated on the stack or heap on
+    ! every wetdep timestep, which required "ulimit -s unlimited" and
+    ! cost allocation + page-fault overhead each timestep).  Values
+    ! are re-initialized per column inside the parallel loop below.
```
lizziel, line 3293: Same comment as before. Remove the justification for the change. This should be part of the pull request description, and could also be put in the commit comments.

lizziel (changes requested):
Thanks @yuyao-cyber! I left a few comments.

yantosca:
@yuyao-cyber I would recommend splitting item 4 into a separate PR since that will change benchmark results.

PR #3385, opened 2026-09-09
Fixes run directory git tracking bug introduced in 14.8.0

mcdon1718-h:

Name: Helena McDonald

### Describe the update

GCHP directories missing a .yaml file (fullchem and carbon) have almost all files not under version control when git add fails to find any *.yaml. This fix changes the approach to git add so that an individual missing file type does not prevent version control for all the others. 

No update to the CHANGELOG since this bug was introduced in 14.8.0 when *.yaml was added. 

### Expected changes

No diff to chemistry.

### DIFF
--- a/run/GCHP/createRunDir.sh
+++ b/run/GCHP/createRunDir.sh
@@ -894,7 +894,7 @@ while [ "$valid_response" -eq 0 ]; do
 	printf "\n"
 	git init
         for f in *.rc *.sh *.yml *.yaml *.py input.nml; do
-	    if [[ -f "${f}" ]]; then
+            if [[ -f "${f}" ]]; then
                 git add "${f}"
             fi
         done
--- a/run/GCHP/createRunDir.sh
+++ b/run/GCHP/createRunDir.sh
@@ -893,14 +893,11 @@ while [ "$valid_response" -eq 0 ]; do
 	printf "\n\nChanges to the following run directory files are tracked by git:\n\n" >> ${version_log}
 	printf "\n"
 	git init
-	for f in *.rc *.sh *.yml *.yaml input.nml; do
-    		if [[ -f "${f}" ]]; then
-        		git add "${f}"
-    		fi
+	for f in *.rc *.sh *.yml *.yaml *.py input.nml; do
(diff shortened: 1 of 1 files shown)


yantosca (changes requested):
Thanks @mcdon1718-h.  We just need a CHANGELOG.md update and then we're good to merge.

yantosca (approved):
Actually forget about the changelog, since this fixes a bug from 14.8.0.  Good to merge.

run/GCHP/createRunDir.sh
```diff
@@ -893,7 +893,11 @@ while [ "$valid_response" -eq 0 ]; do
 	printf "\n\nChanges to the following run directory files are tracked by git:\n\n" >> ${version_log}
 	printf "\n"
 	git init
-	git add *.rc *.sh *.yml *.yaml input.nml
+	for f in *.rc *.sh *.yml *.yaml input.nml; do
+    		if [[ -f "${f}" ]]; then
+        		git add "${f}"
+    		fi
+	done
```
lizziel: Interesting, did the flag to ignore missing files not work?
Also, could you change indenting to match the nest indenting used in the rest of the file? I think we are doing 4 spaces.

lizziel: I also just noticed the next few lines add *.py only if using fullchem and carbon. With the updates I think we can put *.py in the for loop list instead.

lizziel:
@yantosca, please don't merge until I reviewed and accepted.

yantosca:
@lizziel not going to merge yet until you say the word. :-)

yantosca:
@lizziel @mcdon1718-h I've updated the branch to release/14.8.0, as this should go into the 14.8.0 release.

run/GCHP/createRunDir.sh
mcdon1718-h: --ignore-missing only works with --dry-run, at which point git isn't adding anything, it's just printing which things would be added when actually run. Not helpful for us, as we already know what's causing the issue.  

Fixes to come for the *.py and the spacing.

lizziel (approved):

